PR #458, opened 2013-11-23
Update 2.1.0-preview2

samgranieri:
A minor typo prevents ruby 2.1.0 preview2 from being installed.


doppioslash:
The same error message is happening for me on Mac Os X 10.10.4, latest rbenv and ruby-build from brew, it happens to all ruby versions I've tried to install:
- 2.2.2
- 2.2.1
- 2.1.2
- 2.1.0
unable to install any ruby basically.

PR #478, opened 2013-12-24
Set -Wno-error=shorten-64-to-32 only for Ruby >= 1.9.3 and use warnflags for it

mrkn:
Using warnflags is not overwrite CFLAGS. It allows us to specify optflags,
debugflags, and warnflags.


mislav:
Closing because the PR is stale and it's unclear to me why we need the change. I'm not aware of similar reports since Dec 2013.

PR #507, opened 2014-02-10
Require Java 7 for jruby-9000-dev.

chrisseaton:
See #505.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -576,6 +576,23 @@ fix_rbx_irb() {
     true
 }
 
+require_java7() {
+  local version=`java -version 2>&1`
+  if [ $? -eq 0 ]
+  then
+    if [[ $version == *1.[78]* ]]
+    then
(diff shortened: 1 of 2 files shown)


rtlong:
Before I make a PR, I just wanted to clarify: JRuby 9000 should support Java 8 as well, yeah? And from the code here and in  047953e125c9062bc050c26b80d4e742497ece14, looks like that was the intention. 

On Java 1.8.0_66, I'm seeing `java -version` output only on stderr, so this check is failing.

rtlong:
Disregard that. :grimacing: I was looking at older code and jumped to a conclusion. My issue was that I'm running this on Alpine and the default busybox grep implementation doesn't behave enough like GNU grep that this doesn't match:

https://github.com/rbenv/ruby-build/blob/v20151028/bin/ruby-build#L707

Sorry reviving an old issue with my nonsense.

PR #599, opened 2014-07-14
Specify inet protocol

okkez:
We can specify inet protocol via command line option.

I don't want to change system settings persistently.
Such as editing /etc/sysctl.conf.


hsbt:
@okkez Sorry, late response. I merged this.

mislav:
What is this feature useful for? I'm confused. I never saw this.

PR #623, opened 2014-09-03
RUBY_BUILD_MIRROR_URL defaults to https://...

paxan:
Because encryption and authenticity are now cool :smiley_cat:

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1026,7 +1026,7 @@ else
 fi
 
 if [ -z "$RUBY_BUILD_MIRROR_URL" ]; then
-  RUBY_BUILD_MIRROR_URL="http://dqw8nmjcqpjn7.cloudfront.net"
+  RUBY_BUILD_MIRROR_URL="https://dqw8nmjcqpjn7.cloudfront.net"
 else
   RUBY_BUILD_MIRROR_URL="${RUBY_BUILD_MIRROR_URL%/}"
 fi


konklone:
For anyone watching - I've filed a bug with Ruby to add HTTPS to the `cache.ruby-lang.org` server:

https://bugs.ruby-lang.org/issues/10672

konklone:
@mislav It doesn't look like the Ruby team is very interested in doing this right now. Would you please consider moving the link over to `https://` here in rbenv regardless?

My main practical issue with using `http://` for downloading Ruby, even with checksum verification, is the privacy leakage. The packages I download from my laptop or server should provide as little information as possible to any actor doing large-scale traffic correlation, or targeted analysis on the software versions installed on my machine. 

This privacy leak is completely addressed by using HTTPS between my computer and the CloudFront edge. I'd still love for the connection between CloudFront and Ruby's servers to be HTTPS, but the practical ramifications are much lower.

mislav:
> My main practical issue with using `http://` for downloading Ruby, even with checksum verification, is the privacy leakage.

You're allowed to set your own `RUBY_BUILD_MIRROR_URL` to a value that you want, even to a HTTPS version of our current Cloudfront mirror. In a large organization that's concerned about the sort of things you're describing, I'm pretty sure you would want to maintain your own internal mirror. [See `script/mirror` for an example](https://github.com/sstephenson/ruby-build/blob/73988db2ae1cb0339f7e8fde959baf243f19f05e/script/mirror) of a script that could get you going.

You see, my resistance to make this change isn't so much technical as it's on principle. You're asking us: "The official Ruby download locations aren't secure, can you please add a layer of privacy via ruby-build?" and my answer is: I'm not sure whether this is a feature is something we want to take responsibility for. Ruby-build is a simple wrapper around `wget && ./configure && make install`. It is not a project to solve a wider range of ops or security problems.

You see, even a one-character change can be us commiting to a certain promise. In this case, our promise would be "we make the process of installing Ruby versions more secure than if you do it from the official source". I feel uneasy about making and maintaining that promise in the long run.

konklone:
> You're allowed to set your own RUBY_BUILD_MIRROR_URL to a value that you want, even to a HTTPS version of our current Cloudfront mirror. In a large organization that's concerned about the sort of things you're describing, I'm pretty sure you would want to maintain your own internal mirror.

I'm actually more concerned about smaller orgs and individuals, especially those who don't realize you can use `RUBY_BUILD_MIRROR_URL`. I've been an enthusiastic user of `rbenv` for a year or two now, and I had no idea you could do that.

> "The official Ruby download locations aren't secure, can you please add a layer of privacy via ruby-build?" and my answer is: I'm not sure whether this is a feature is something we want to take responsibility for. 

I understand that you can't guarantee end-to-end encrypted transport for the tarball from Ruby to one's computer. But what you can guarantee is encrypted transport between CloudFront and the user's computer. And that, even without any changes by Ruby's servers, closes off information about individual Ruby users from traveling across the network. The traffic between CloudFront and Ruby is miniscule and, since you're caching heavily, not trackable to individual downloads. So you can accurately take responsibility for that.

I think the promise here isn't about security, because the checksums take care of that well. You'd be promising privacy, and can meet that promise for all rbenv users, whether they know about the advanced options or not, on your own.

reedloden:
@konklone was successful in getting Ruby to use https:// for cache.ruby-lang.org, and all the links on https://www.ruby-lang.org will use https:// for downloading once they do a site push. I've filed #786 to update various places in ruby-build to use https:// that now explicitly support them.

paxan:
Thank you @konklone for your relentless push to encrypt HTTP traffic even in seemingly uncalled-for cases like this. 👍🏼

Issue #557, opened 2014-04-22
Please list ruby core optional dependencies in the documentation


a1exsh:
Installed ruby-1.9.3 using rbenv+ruby-build on a clean system today.

Problems:
1. The `gem` command is broken due to missing `zlib` (e.g. `gem update --system`)
2. The `irb` prompt is less functional than used to be due to lack of readline support.

Fixed by installing dev packages for zlib and libreadline, then re-installing the ruby version (had to do that two times actually: first for zlib, then discovered that `irb` problem--not nice.)

It would be really nice if you mention these as caveats in the documentation (by the way, I believe libssl is another optional dependency of ruby core, but I can't find the definitive list ATM, guess checking ./configure options in ruby source tree would be it.)

mislav:
Closing as outdated

Issue #509, opened 2014-02-15
Ruby 2.1.0 libyaml not functional without setting LD_LIBARY_PATH on FreeBSD


p-originate:
<pre>
stealth% rbenv install 2.1.0
Downloading yaml-0.1.5.tar.gz...
-> http://dqw8nmjcqpjn7.cloudfront.net/24f6093c1e840ca5df2eb09291a1dbf1
Installing yaml-0.1.5...
Installed yaml-0.1.5 to /home/me/.rbenv/versions/2.1.0

Downloading ruby-2.1.0.tar.gz...
-> http://dqw8nmjcqpjn7.cloudfront.net/9e6386d53f5200a3e7069107405b93f7
Installing ruby-2.1.0...
Installed ruby-2.1.0 to /home/me/.rbenv/versions/2.1.0
</pre>

Ok. Now let's try:

<pre>
stealth% gem install rails
/home/me/.rbenv/versions/2.1.0/lib/ruby/2.1.0/yaml.rb:4:in `<top (required)>':
It seems your ruby installation is missing psych (for YAML output).
To eliminate this warning, please install libyaml and reinstall your ruby.
/home/me/.rbenv/versions/2.1.0/lib/ruby/2.1.0/rubygems/core_ext/kernel_require.rb:55:in `require': Shared object "libyaml-0.so.2" not found, required by "psych.so" - /home/me/.rbenv/versions/2.1.0/lib/ruby/2.1.0/x86_64-freebsd9.1/psych.so (LoadError)
    from /home/me/.rbenv/versions/2.1.0/lib/ruby/2.1.0/rubygems/core_ext/kernel_require.rb:55:in `require'
    from /home/me/.rbenv/versions/2.1.0/lib/ruby/2.1.0/psych.rb:1:in `<top (required)>'
    from /home/me/.rbenv/versions/2.1.0/lib/ruby/2.1.0/rubygems/core_ext/kernel_require.rb:55:in `require'
    from /home/me/.rbenv/versions/2.1.0/lib/ruby/2.1.0/rubygems/core_ext/kernel_require.rb:55:in `require'
    from /home/me/.rbenv/versions/2.1.0/lib/ruby/2.1.0/yaml.rb:5:in `<top (required)>'
    from /home/me/.rbenv/versions/2.1.0/lib/ruby/2.1.0/rubygems/core_ext/kernel_require.rb:55:in `require'
[...]

PR #629, opened 2014-09-08
Check that openssl, yaml, readline, zlib are all loadable

mislav:
Before, we detected whether `require "openssl"` works after compiling Ruby 2.0.0+ and failed the build if the openssl extension never compiled. This change extends this practice to zlib, yaml and readline as well since we consider these to be essential packages for most Ruby development. Rubygems won't work, for instance, without zlib.

This also suggests how you could install necessary dependencies if a package manager `apt-get` or `yum` is detected on the system.

Example output:

    The Ruby zlib extension was not compiled.
    The Ruby openssl extension was not compiled.
    Try running `apt-get install -y libssl-dev zlib1g-dev` to fetch missing dependencies

Pros:

- Detects stuff like #509 or #557 and aborts the install
- Tells you how you can use your package manager to fetch the missing dependency

Cons:

- Maybe some people don't need stuff like readline or zlib and are fine if these aren't compiled?
- With this, `verify_openssl` function becomes a misnomer since it verifies more stuff than just openssl.

How does everyone feel about this?

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -834,15 +834,38 @@ build_package_mac_openssl() {
 
 # Post-install check that the openssl extension was built.
 build_package_verify_openssl() {
-  "$RUBY_BIN" -e 'begin
-    require "openssl"
-  rescue LoadError
-    $stderr.puts "The Ruby openssl extension was not compiled. Missing the OpenSSL lib?"
-    $stderr.puts "Configure options used:"
-    require "rbconfig"; require "shellwords"
-    RbConfig::CONFIG.fetch("configure_args").shellsplit.each { |arg| $stderr.puts "  #{arg}" }
-    exit 1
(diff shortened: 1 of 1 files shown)


p-run:
Why not make the message a warning at the very end of installation process?

Narayanan170:
Please merge this as soon as possible, I'm using ruby-build on a lot of servers now, every time I install a new one I end up on the issues page of this project.

mislav:
@sstephenson How do you feel about this? The changeset is a bit large.

sstephenson:
@mislav +1 on the change, in general. I'd suggest renaming the function (as you pointed out) and splitting the Ruby script out into its own file.

sferik:
I’m going to merge this change. Renaming and refactoring can be done in the future, as this function is not part of the public API.

PR #633, opened 2014-09-09
Show specific upgrade instructions based on environment

mislav:
Because ruby-build is both a Homebrew formula and an rbenv plugin, some people end up with both installed. In these cases, executing `rbenv install` vs. `ruby-build` might not use the same install. Users end up being unsure where `rbenv install` is coming from and upgrading it from the wrong location.
- https://github.com/sstephenson/ruby-build/issues/588#issuecomment-48026552
- https://github.com/sstephenson/ruby-build/issues/569#issuecomment-42778989
- https://github.com/sstephenson/ruby-build/pull/480
- https://github.com/sstephenson/ruby-build/issues/366#issue-15051163
- https://github.com/sstephenson/ruby-build/issues/362#issuecomment-19513846
- https://github.com/sstephenson/ruby-build/issues/254#issuecomment-16306018

This detect whether ruby-build is contained in Homebrew or git and shows upgrade instructions accordingly. Example:

```
See all available versions with `rbenv install --list'.

If the version you need is missing, try upgrading ruby-build:

  brew update && brew upgrade ruby-build
```

How does everyone feel about this?

### DIFF
--- a/bin/rbenv-install
+++ b/bin/rbenv-install
@@ -195,17 +195,25 @@ ruby-build $KEEP $VERBOSE $HAS_PATCH "$DEFINITION" "$PREFIX" || STATUS="$?"
 # Display a more helpful message if the definition wasn't found.
 if [ "$STATUS" == "2" ]; then
   { candidates="$(definitions "$DEFINITION")"
+    here="$(dirname "${0%/*}")"
     if [ -n "$candidates" ]; then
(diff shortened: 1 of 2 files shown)


muescha:
In the readme i would also add an info to the upgrade section:

Which method you need is printed via `rbenv install --list`

PR #634, opened 2014-09-10
Skip `ldflags_dirs` step for Rubies where it's not necessary

mislav:
`ldflags_dirs` was added for Ruby 2.1.0-dev in 50bf60f9c2510fbee7f111046e97fe8bf9e66472.

Since then, in releases of Ruby 2.1.x the LDFLAGS checker seems to have gotten more permissive and this step isn't necessary anymore. However, Ruby 2.1.2 and 2.2.0-dev seem to need it still.

This removes `ldflags_dirs` from at least those build definitions where it's not necessary in order to clean them up and avoid cargo-culting.

@hsbt: Why do you think LDFLAGS checker has gotten less strict since 2.1.0 release but has become more strict again in 2.1.2 and onwards?


hsbt:
@mislav Sorry, too late response. I confirmed this.

PR #658, opened 2014-10-29
Add small note to use `rbenv global` after `rbenv install`

tombh:
Installing Ruby from scratch and I just spent way too long figuring out that you have to `rbenv global [version]` after install. It was only because I remembered from a few months ago when I last installed from scratch that I didn't pull my hair out and give up :/


p-run:
+1 to tell users to set a version globally.

Issue #706, opened 2015-01-19
Let ruby-build try to build with the best compiler available to the system


grepwood:
https://p8952.info/ruby/2014/12/12/benchmarking-ruby-with-gcc-and-clang.html
According to this benchmark, building with Clang is preferrable on a number of server-oriented Linux systems.
CentOS 6 has GCC 4.4 and Clang 3.4, the latter wins.
CentOS 7 has GCC 4.8 and Clang 3.5, the former wins.
Ubuntu 14.04 is in the same situation as CentOS 7.

grepwood:
Pull request #707 fixes this issue for CentOS. Ubuntu and Debian will be added later.

mislav:
Closing for the same reason as https://github.com/sstephenson/ruby-build/pull/707#issuecomment-70571673

Thanks for the suggestion, though! :dizzy:

PR #707, opened 2015-01-19
Fixes issue #706

grepwood:
Performs compiler profiling based on https://p8952.info/ruby/2014/12/12/benchmarking-ruby-with-gcc-and-clang.html

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -14,6 +14,51 @@ RUBY_BUILD_VERSION="20150116"
 set -E
 exec 3<&2 # preserve original stderr at fd 3
 
+# compiler ranks, higher is better
+# gcc 4.4	1
+# clang 3.4	2
+# gcc 4.7	3
+# gcc 4.8	4
+# clang 3.2	5
+# clang 3.3	6
+# clang 3.5	7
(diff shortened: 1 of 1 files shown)


mislav:
No, sorry; thanks for looking into this, but I don't think we want so much extra code in core.

Do you want to make an rbenv plugin instead? The plugin could publish an `install` hook and set the necessary environment variables based on this criteria if no `CC` was explicitly set.

grepwood:
That'd be alright I think. So the plugin would have to perform the function I implemented and then call `install` from a child process that has `CC` set? Just want to clarify.

mislav:
See [rbenv docs on authoring plugins](https://github.com/sstephenson/rbenv/wiki/Authoring-plugins) and the source of [`rbenv-install`](https://github.com/sstephenson/ruby-build/blob/73988db2ae1cb0339f7e8fde959baf243f19f05e/bin/rbenv-install).

Basically the plugin should have something like `etc/rbenv.d/install/smart_set_compiler.bash` with the contents of:

```sh
smart_set_compiler() {
  # do something to set CC based on VERSION_NAME that's being installed
}

before_install smart_set_compiler
```

PR #715, opened 2015-01-28
Cleaner error if you have a .rbenv file, not a directory.

paulgeringer:
If you do something incredibly stupid and misread the error generated
from a clean install of rbenv (see *), you might do something like I did
and `touch ~/.rbenv` thinking that will silence the warning. The error
you get when you install ruby-build and subsequently try to install a
ruby is a bit unclear.

There is no real reason this needs to be merged, though it
might be somewhat helpful in case someone else does the same goof.

*: /usr/local/Cellar/rbenv/0.4.0/libexec/rbenv---version: line 17: cd:
/Users/jenkins/.rbenv: No such file or directory

### DIFF
--- a/bin/rbenv-install
+++ b/bin/rbenv-install
@@ -26,6 +26,11 @@ set -e
 
 if [ -z "$RBENV_ROOT" ]; then
   RBENV_ROOT="${HOME}/.rbenv"
+  if [ ! -d $RBENV_ROOT ]; then
+    echo "$RBENV_ROOT is not a directory!"
+    echo "'rm $RBENV_ROOT' to fix this issue."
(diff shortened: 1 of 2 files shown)


mislav:
What error did you get in the first place and how did you reach it?

p-run:
`-d` will also fail on a broken symlink, might want to add this to the error message.

mislav:
I don't think it should be responsibility of this project (that's partly a rbenv plugin) to detect and warn about botched rbenv installs. I also don't think many people have this problem.

This would be better handled by having an installer/doctor script for rbenv. I'm working on something there.

PR #722, opened 2015-02-25
Use HTTPS for CloudFront URLs

lreeves:
Currently ruby-build uses straight HTTP for downloads from CloudFront. Since CloudFront has SSL enabled for any endpoints this is an easy change that should help secure downloads from CloudFront.

Like most CDNs CloudFront does charge slightly more for HTTPS over HTTP but that's probably worth it - if not then feel free to ignore this PR, no harm no foul :-)

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1117,7 +1117,7 @@ else
 fi
 
 if [ -z "$RUBY_BUILD_MIRROR_URL" ]; then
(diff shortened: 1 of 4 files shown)


sferik:
See https://github.com/sstephenson/ruby-build/pull/623.

lreeves:
Thanks, will close this one!

PR #711, opened 2015-01-21
Fix JRuby 9.0.0.0-pre1 definition

deepj:
Sorry for the typo, my fault. I overlooked it :(

PR #724, opened 2015-02-25
Rename jruby-9.0.0.0-pre1 to jruby-9.0.0.0.pre1

robbiegill:
So the installed ruby version will match what is used in `.ruby-version` files

This is the version used by the jruby team.


mislav:
Could we please have comment from @deepj? #711

I'm all for consistent naming, but we have to get it right finally. If we keep renaming JRuby 9000 versions, we break things for ruby-build and rbenv users who are already using it under the old name.

mislav:
@headius do you have recommendations on which syntax is preferred for referencing JRuby 9000 version?

deepj:
I used `-pre1` because it had been used on the official page announcing releasing new JRuby 9000 preview1 release.

Proof: http://web.archive.org/web/20150121021101/http://jruby.org/

The same syntax was also used in several other cases of JRuby versions like

https://github.com/sstephenson/ruby-build/blob/master/share/ruby-build/jruby-9.0.0.0-dev
https://github.com/sstephenson/ruby-build/blob/master/share/ruby-build/jruby-9.0.0.0%2Bgraal-dev
https://github.com/sstephenson/ruby-build/blob/master/share/ruby-build/jruby-1.7.0-preview1

So I didn't see anything wrong with using `-pre1`.

mislav:
On Thu, Feb 26, 2015 at 12:00 PM, deepj <notifications@github.com> wrote:

> So I didn't see anything wrong with using -pre1.
>
Thanks; it seems that they were using `-pre1` initially but have since
pretty consistently (at least for the 9000 series) switched to `.pre1`.
Syntax. I'm :+1: for the rename but feel a bit bummed out if this is going
to inconvenience some of our users. But since we're dealing with a
developer preview of an alternate Ruby implementation, I guess this is OK
since we can assume that the users of this prerelease are advanced users
how will manage to handle renaming themselves if they already have used the
previous version syntax.

deepj:
I'm in for renaming. As you said JRuby 9.0.0.0.pre1 is for advanced users so there would be no complain about the change. I don't see any problem in that.

deepj:
:+1:

PR #727, opened 2015-03-03
added definition of Ruby 2.2.1

hsbt:
see https://www.ruby-lang.org/en/news/2015/03/03/ruby-2-2-1-released/


sferik:
I’ll release it.

PR #733, opened 2015-03-19
Upgrade openssl-1.0.1m

deepj:


mislav:
Thanks!

OpenSSL 1.0.1 is halfway through the alphabet already. When they get to the end, I wonder if they'll consider using numbers for versioning again.

deepj:
Nope, they stop at `z` and they start from the beginning on the second position :smile: i.e. `openssl-0.9.8zf.tar.gz` 

https://www.openssl.org/source/

It'd be nice to consider to release new ruby-build when new openssl was released. But that's up to you.

sferik:
Yeah, I’ll make a release. I think this is the only change since the last release, so the release notes should be easy.

mislav:
@sferik: ruby-build 20150319zf

deepj:
:smile_cat:

sferik:
@mislav I am so doing that.

mislav:
Oh wow he actually did it :joy_cat:

PR #735, opened 2015-03-21
No longer installs the jruby-launcher gem unless explicitly requested.

suetanvil:
Since jruby already comes with launch scripts, jruby-launcher isn't
really necessary.  It does, however completely scuttle the
installation if the compile fails.

This change skips the installation of jruby-launcher unless the
install command was invoked with the flag '--launcher-gem'.

This is a fix for issue:

<https://github.com/sstephenson/ruby-build/issues/734>.

### DIFF
--- a/bin/rbenv-install
+++ b/bin/rbenv-install
@@ -16,6 +16,7 @@
 #                      (defaults to $RBENV_ROOT/sources)
 #   -v/--verbose       Verbose mode: print compilation status to stdout
 #   -p/--patch         Apply a patch from stdin before building
+#   --launcher-gem     Install java-launcher gem for jruby
 #
 # For detailed information on installing Ruby versions with
 # ruby-build, including a list of environment variables for adjusting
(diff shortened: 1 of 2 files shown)


sferik:
This pull request causes the tests to fail and no longer merges cleanly. If you still care about this issue, can you please fix the tests and rebase from `master`. Thanks!

mislav:
Sorry for my lack of review on this one. I'm not a fan of adding a new configuration option. How about just installing the launcher gem for JRuby versions that need it, and avoid installing it for newer JRuby? Do we know which JRuby versions needed it in the first place?

mislav:
@headius Do we still need to install jruby-launcher unconditionally for newer JRuby versions? https://github.com/sstephenson/rbenv/issues/94#issuecomment-2172687 7b3dc50ce1ec1e5fc82060d4b5d312e6b4209177

If not, then for which old JRubies should we keep jruby-launcher?

suetanvil:
It looks like this is a non-starter, so I'm closing the pull request.

I do think that this shows a need for some more fine-grained control of the installation process. Perhaps the installation process could be split into several stages with command-line options available to disable one or more stages?  In any case, there ought to be tools available to work around bugs in the installation process.

PR #737, opened 2015-04-13
Add 'cd -' to proposed steps to upgrade ruby-build

huerlisi:
When a target is not found, rbenv-install proposes pulling changes from
upstream. This involves changing the directory to the ruby-build checkout.

This patch adds a '&& cd -' to change the working directory back to where
we've been. Should work on at least Bash and Zsh.

### DIFF
--- a/bin/rbenv-install
+++ b/bin/rbenv-install
@@ -211,7 +211,7 @@ if [ "$STATUS" == "2" ]; then
       echo "  brew update && brew upgrade ruby-build"
     elif [ -d "${here}/.git" ]; then
       printf ":\n\n"
-      echo "  cd ${here} && git pull"
+      echo "  cd ${here} && git pull && cd -"
     else
       printf ".\n"
     fi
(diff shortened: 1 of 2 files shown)


mislav:
Thanks for the suggestion but I have faith in our users that they will know how to return to the previous directory.

huerlisi:
It's not about them (or me actually:-) not knowing how to get back, but about convenience. It's a perfect copy/paste job, that's why I like that you output the exact steps in a shell compatible way.

But I do not think one will ever stay in this directory unless the pull didn't work out. In that case, the 'cd -' will be skipped.

And by the way, there's actually quite some guys that do not know how to get back to the previous directory. Trained some of them, so they know now;-)

So this patch makes it more convenience. I don't think it would hurt? Or why do you think it would be worse than what you print now?

mislav:
Alright but then you have to fix the tests :stuck_out_tongue:

huerlisi:
Fair enough:-) Will do.

huerlisi:
There you go.

Do you want me to squash the spec change into the functional change?

mislav:
Yep

huerlisi:
Voilà :-)

mislav:
Thanks :)

PR #744, opened 2015-04-14
documenting the -s/--skip-existing option

jstriebel:
This will also show the skip-option in the help
(see https://github.com/sstephenson/ruby-build/blob/master/bin/rbenv-install#L83).

Thank you for regarding this PR. Best, Jonathan


mislav:
`ruby-build` executable doesn't support `--skip-existing`.

jstriebel:
Oh sorry, true, I mixed things up.

PR #750, opened 2015-04-24
SSL/TLSize all the things! (convert http:// to https:// where possible)

reedloden:
Even though ruby-build uses hardcoded checksums to help prevent MITM and corruption issues, using SSL/TLS to download the files adds another layer of protection, ensuring that nobody can see that somebody is downloading a certain ruby version.

Quoting Eric Mill from https://bugs.ruby-lang.org/issues/10672

> Even when clients do perform client-side integrity checking, there is always a privacy implication to downloading information. Downloading Ruby without SSL leaks information about the client performing the download through request headers, and informs anyone watching the connection what version of Ruby is likely to be running on the downloading machine. In addition, traffic can be correlated in unpredictable ways: for example, a user agent sent to connect to a download of a Ruby build may appear later to download other information, providing a pattern of client behavior.
> 
> In any case, the web is, in general, moving to favor encrypted connections. SSL is faster, CAs like SSLMate and Let's Encrypt are emerging to make the process simpler, and web browsers are starting to encourage encrypted connections and discourage unencrypted ones.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ Example usage:
 $ rbenv install --patch 1.9.3-p429 < /path/to/ruby.patch
 
 # applying a patch from HTTP
-$ rbenv install --patch 1.9.3-p429 < <(curl -sSL http://git.io/ruby.patch)
+$ rbenv install --patch 1.9.3-p429 < <(curl -sSL https://git.io/ruby.patch)
 
 # applying multiple patches
 $ cat fix1.patch fix2.patch | rbenv install --patch 1.9.3-p429
@@ -179,7 +179,7 @@ mirror and use official URLs instead. You can force ruby-build to bypass the
 mirror by setting the `RUBY_BUILD_SKIP_MIRROR` environment variable.
 
 The official ruby-build download mirror is sponsored by
-[37signals](http://37signals.com/).
+[Basecamp](https://basecamp.com/).
 
 ### Package download caching
 
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1139,7 +1139,7 @@ else
 fi
 
 if [ -z "$RUBY_BUILD_MIRROR_URL" ]; then
-  RUBY_BUILD_MIRROR_URL="http://dqw8nmjcqpjn7.cloudfront.net"
+  RUBY_BUILD_MIRROR_URL="https://dqw8nmjcqpjn7.cloudfront.net"
 else
   RUBY_BUILD_MIRROR_URL="${RUBY_BUILD_MIRROR_URL%/}"
 fi
--- a/share/ruby-build/1.8.6-p383
+++ b/share/ruby-build/1.8.6-p383
@@ -1,3 +1,3 @@
 require_gcc
 install_package "ruby-1.8.6-p383" "http://cache.ruby-lang.org/pub/ruby/1.8/ruby-1.8.6-p383.tar.gz#cea6c67f737727007ec89d1c93fd7d0dba035220f981706091b8642d7a43c03a" auto_tcltk standard
-install_package "rubygems-1.3.7" "http://production.cf.rubygems.org/rubygems/rubygems-1.3.7.tgz#388b90ae6273f655507b10c8ba6bee9ea72e7d49c3e610025531cb8c3ba67c9d" ruby
+install_package "rubygems-1.3.7" "https://d2chzxaqi4y7f8.cloudfront.net/rubygems/rubygems-1.3.7.tgz#388b90ae6273f655507b10c8ba6bee9ea72e7d49c3e610025531cb8c3ba67c9d" ruby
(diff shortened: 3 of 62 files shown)


reedloden:
Yes, I'm aware of #623 and #722. There's no reason why *not* to use https:// today when it's available.

reedloden:
Tests are failing because http://files.rubyforge.vm.bytemark.co.uk doesn't exist anymore... Going to swap to http://mirror.as24220.net.

reedloden:
Apparently, the checksum for rubinius-2.2.10.tar.bz2 has changed? Confirmed that both the http:// and https:// show a different checksum (`3cb1a6ab2eba19b6dc84734666bb17a34332d247641b1a88b4c9324c69347780` vs. `2a3e6b71f27073b8d83b9592b05523af70bc147ddcd0673bffae55b4167c9d81`).

Seems odd to me...

reedloden:
I've gone ahead and updated it in my diff for the purposes of this PR, but somebody should reach out to the rubinius folk to figure out what happened with the changing checksum.

mislav:
Hi, thanks for putting work into this. However, it's very unlikely that we will accept this, mostly for reasons you already read in the threads that you referenced.

1. We're not switching to HTTPS across the board until official Ruby download locations support HTTPS.
2. You swapped a lot of canonical download locations in definition files with `d2chzxaqi4y7f8.cloudfront.net`, which is not even our mirror and we don't trust it. Even if it was ours, our mirror service is supposed to be transparent; i.e. a definition file lists the canonical download location, that resource gets copied to our mirror during CI process, ruby-build users transparently hit our mirror for faster download. But, the definition file must list the official download location and official checksum, because users trust those. If our mirror is down, the installs are never affected.
3. You swap hosts like `ci.jruby.org` and `asset.rubini.us` with `s3.amazonaws.com`, and this is unacceptable. Even though right now those download locations might be powered by S3, it doesn't mean this will continue to be so in the future. The canonical download locations for JRuby and Rubinius are NOT from Amazon's hostnames. They are from JRuby's and Rubinius' own domains, respectively. For them, S3 is just an implementation detail. I refuse to expose that implementation detail here just because they haven't provided security by buying SSL certificates.
4. It sucks that `files.rubyforge.vm.bytemark.co.uk` doesn't exist anymore; thanks for letting us know. However, that was the canonical download location then, and I don't trust the `mirror.as24220.net` hostname. Maybe we should just leave this broken since REE 1.8.6 was EOL so long ago that nobody should be using it anymore.

In short: mirrors and CDNs should *not* be canonical download locations.

Thanks for letting us know that the checksum for `rubinius-2.2.10.tar.bz2` changed. That's indeed worrisome. I've opened https://github.com/rubinius/rubinius/issues/3380, but since I've verified the integrity of the updated archive (they're basically the same), I will bump the checksum.

mislav:
I hope my rationale above explains everything. Closing for the same reasons. Thanks for the suggestion, though!

reedloden:
On Mon, Apr 27, 2015 at 2:19 AM, Mislav Marohnić <notifications@github.com>
wrote:

> Hi, thanks for putting work into this. However, it's very unlikely that we
> will accept this, mostly for reasons you already read in the threads that
> you referenced.
>
>    1. We're not switching to HTTPS across the board until official Ruby
>    download locations support HTTPS.
>
> This just changed (https://bugs.ruby-lang.org/issues/10672#change-53270),
so I'll be resubmitting this PR.

>
>    1. You swapped a lot of canonical download locations in definition
>    files with d2chzxaqi4y7f8.cloudfront.net, which is not even our mirror
>    and we don't trust it. Even if it was ours, our mirror service is supposed
>    to be transparent; i.e. a definition file lists the canonical download
>    location, that resource gets copied to our mirror during CI process,
>    ruby-build users transparently hit our mirror for faster download. But, the
>    definition file must list the official download location and official
>    checksum, because users trust those. If our mirror is down, the installs
>    are never affected.
>
> d2chzxaqi4y7f8.cloudfront.net is production.cf.rubygems.org. They are the
same thing. Note the 'cf' in the FQDN stands for CloudFront.

Could also use https://s3.amazonaws.com/production.s3.rubygems.org/, which
is what rubygems.org redirects to, according to
https://github.com/rubygems/rubygems-infrastructure/blob/master/cookbooks/rubygems-balancer/templates/default/site.conf.erb
.

I just filed https://github.com/rubygems/rubygems-infrastructure/issues/25
to get SSL working for https://production.cf.rubygems.org directly, but I
don't think there is any reason to not use either the appropriate https://
version of the CloudFront URL or the S3 one, as both are using anyway and
are tied to an implementation via the hostname.

>
>    1. You swap hosts like ci.jruby.org and asset.rubini.us with
>    s3.amazonaws.com, and this is unacceptable. Even though right now
>    those download locations might be powered by S3, it doesn't mean this will
>    continue to be so in the future. The canonical download locations for JRuby
>    and Rubinius are NOT from Amazon's hostnames. They are from JRuby's and
>    Rubinius' own domains, respectively. For them, S3 is just an implementation
>    detail. I refuse to expose that implementation detail here just because
>    they haven't provided security by buying SSL certificates.
>
> I disagree, especially since a new version of ruby-build can just be
release if they do change in the future, but I will revert that in my PR
update if it helps move things forward.


>
>    1. It sucks that files.rubyforge.vm.bytemark.co.uk doesn't exist
>    anymore; thanks for letting us know. However, that was the canonical
>    download location then, and I don't trust the mirror.as24220.net
>    hostname. Maybe we should just leave this broken since REE 1.8.6 was EOL so
>    long ago that nobody should be using it anymore.
>
> I agree the mirror is weird, but you yourself have mentioned the reliance
on the checksum, so how is it any different from using an http:// mirror,
which can be MITM'd?

>
>
> Thanks for letting us know that the checksum for rubinius-2.2.10.tar.bz2
> changed. That's indeed worrisome. I've opened rubinius/rubinius#3380
> <https://github.com/rubinius/rubinius/issues/3380>, but since I've
> verified the integrity of the updated archive (they're basically the same),
> I will bump the checksum.
>
Cool.

Updated PR coming shortly.

PR #751, opened 2015-04-28
Add jruby-9.0.0.0.pre2

jirutka:


headius:
:+1:

jirutka:
@headius Is the requirement for Java 7 (i.e. it excludes Java 8) still valid?

jirutka:
ping @headius

deepj:
@sferik or @mislav Guys, is it possible to release new version of `ruby-build` in favour of this pull (don't forget to publish it for `homebrew` as well). It'd be worth to have `JRuby 9.0.0.0.pre2` available for other developers for their environment using `ruby-build` :)

sferik:
@deepj We don’t normally ship new versions of `ruby-build` solely for pre-release versions. If you want to try this out with `homebrew`, you can type:

```
brew upgrade ruby-build --HEAD
```

headius:
JRuby 9k will required Java 7 _or higher_. You can use 8 just fine (or 9, but we haven't done much testing there).

jirutka:
@headius Ah, `require_java7` actually means “Java 7, 8, or 9” ([here](https://github.com/sstephenson/ruby-build/blob/e9d25d0f615d2dc2e6b7b32562d6a4eee9e9889e/bin/ruby-build#L703-L709), so the problem is elsewhere – I have OpenJDK 8 and `ruby-build` complains that I don’t have Java 7. Okay, I’ll examine it more.

PR #752, opened 2015-04-29
New Update Section in README.md

ahoglund:
Was updating my ruby-build version today and thought the README could be a bit clearer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,12 +45,12 @@ variable.
 To update ruby-build after it has been installed, run `git pull` in your cloned
 copy of the repository, then re-run the install script.
 
-### For plug-in
+### For rbenv plugin
     cd /path/to/.rbenv/plugins/ruby-build
     git pull
     ./install.sh
 
-### For standalone
+### For standalone program
     cd /path/to/ruby-build  (usually /usr/local/ruby-build)
     git pull
     ./install.sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,9 +40,21 @@ permission to `/usr/local`, you will need to run `sudo ./install.sh` instead.
...
 
+## Updating
+
 To update ruby-build after it has been installed, run `git pull` in your cloned
 copy of the repository, then re-run the install script.
 
+### For rbenv plugin
+    cd /path/to/.rbenv/plugins/ruby-build
+    git pull
+    ./install.sh
```
hsbt: this is redundant instruction.

mislav: Yes, if it's an rbenv plugin then it doesn't need running `./install.sh`.

mislav:
Thanks for the suggestion, but I think we will stick with [the update information in the wiki](https://github.com/sstephenson/ruby-build/wiki#updating-ruby-build) for now. Updating ruby-build can be tricky for users who don't remember how they installed it (or have installed it twice, for that matter) so I'll try to come up with some tool that would automate this.

README.md
ahoglund: thanks! didn't know that.

PR #753, opened 2015-04-29
Properly detect OpenJDK versions in bin/ruby-build

bbrowning:
OpenJDK reports its version slightly different than Oracle Java, and
this was causing the logic in bin/ruby-build to not detect Java 7 or
higher when OpenJDK was used.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -701,7 +701,7 @@ fix_rbx_irb() {
 }
 
 require_java7() {
-  local version="$(java -version 2>&1 | grep 'java version' | head -1)"
+  local version="$(java -version 2>&1 | grep '\(java\|openjdk\) version' | head -1)"
   if [[ $version != *1.[789]* ]]; then
     colorize 1 "ERROR" >&3
     echo ": Java 7 required. Please install a 1.7-compatible JRE." >&3
--- a/test/build.bats
+++ b/test/build.bats
@@ -599,6 +599,18 @@ DEF
   assert_success
 }
 
+@test "Java version string on OpenJDK" {
+  cached_tarball "jruby-9000.dev" bin/jruby
+
+  stub java "-version : echo 'openjdk version \"1.8.0_40\"' >&2"
+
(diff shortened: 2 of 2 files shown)


jirutka:
I have exactly the same problem!

bin/ruby-build
```diff
@@ -701,7 +701,7 @@ fix_rbx_irb() {
 }
 
 require_java7() {
-  local version="$(java -version 2>&1 | grep 'java version' | head -1)"
+  local version="$(java -version 2>&1 | grep '\(java\|openjdk\) version' | head -1)"
```
jirutka: This doesn’t work on OS X (BSD grep). Grep command should be replaced with: `grep -E '(java|openjdk) version'`. This works with both BSD and GNU grep.

jirutka:
@mislav Could you please merge this?

bin/ruby-build
mislav: Because of backslashes, this works for me on BSD grep, and since it passes CI (on Linux) I presume with GNU grep as well. It doesn't need changing right now.

mislav:
Thanks @bbrowning! And thanks for the review @jirutka

bin/ruby-build
jirutka: Ah, you’re right, I’ve apparently messed something when I tested it.

chnoack: This does not work for me (alpine linux, idk 1.7.0_79)

When I do:
`java -version`

I get this result:
```
java version "1.7.0_79"
Java(TM) SE Runtime Environment (build 1.7.0_79-b15)
Java HotSpot(TM) 64-Bit Server VM (build 24.79-b02, mixed mode)
```

But when I do:
`java -version 2>&1 | grep '\(java\|openjdk\) version' | head -1`

I get no result.

mislav: @chnoack Thanks for reporting. It seems that `grep` on your system doesn't support the regular expression used here?

PR #764, opened 2015-05-26
Fix: Detect LLVM 3.5 installed from Homebrew

brandt:
- Updates the command used to detect LLVM 3.5 installed with Homebrew on Mac OS X.
- Updates instructions with current command to install LLVM 3.5 using Homebrew.
- Falls back to existing behavior if version-specific `llvm-config` isn't found.

Around two weeks ago, [Homebrew bumped the version](https://github.com/Homebrew/homebrew/commit/0196a069a01e455b61778d804ebb81f46e7c31ff) of you get with `brew install llvm` from LLVM 3.5.1 to 3.6.0.

So now users will have to tap homebrew-versions and run `brew install llvm35` in order to get the version necessary to install recent versions of Rubinius. The 3.5-specific `llvm-config` becomes `llvm-config-3.5` accordingly.

For users who installed LLVM before Homebrew bumped the version (i.e. their `llvm-config` points to 3.5), we fallback to the existing behavior if the version-specific formula isn't found.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -815,8 +815,8 @@ require_llvm() {
       if [ "$llvm_version" = "3.2" ]; then
         package_option ruby configure --prebuilt-name="llvm-3.2-x86_64-apple-darwin13.tar.bz2"
       else
-        local llvm_prefix="$(brew --prefix llvm 2>/dev/null || true)"
-        local llvm_config="${llvm_prefix}/bin/llvm-config"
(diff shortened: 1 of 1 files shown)


mislav:
Thanks for bringing this to my attention. However, I'm not a fan that this relies on `cut` and version string manipulation. The code feels brittle. I opened an alternative implementation as another pull and would appreciate if you gave it a :eyes: thanks

brandt:
Sure thing. I like your solution better -- closing

PR #765, opened 2015-05-26
Improve detecting LLVM 3.5 from Homebrew

mislav:
This lists through all installed Homebrew packages named "llvm\*" and tests each `llvm-config --version` for "3.5\*" until it finds one.

Alternative to #764

/cc @brandt

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -812,12 +812,13 @@ require_llvm() {
   local llvm_version="$1"
   if [ "$(uname -s)" = "Darwin" ] && [ "$(osx_version)" -ge 1010 ]; then
     if [[ "$RUBY_CONFIGURE_OPTS" != *--llvm-* ]]; then
-      if [ "$llvm_version" = "3.2" ]; then
+      case "$llvm_version" in
+      3.2 )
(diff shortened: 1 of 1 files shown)


sferik:
The next version of Rubinius should support LLVM 3.6. When that’s released will this be reverted?

mislav:
No, we will tag new Rubinius formulas with `require_llvm 3.6`

sferik:
@mislav Makes sense.

:+1:

mislav:
However if new Rubinius will support *both* 3.5 and 3.6, I suppose this code would be amended with that logic.

brandt:
@mislav Looks good. I like your solution better.

- :+1: Properly deals with version numbers that have "svn" appended, e.g. "3.7.0svn". (Happens when LLVM is installed from HEAD)
- :+1: Selects compatible version regardless of order in which packages appear in `brew list`.
- :+1: Doesn't panic when `llvm-config*` is missing from LLVM package.
- :+1: Gracefully deals with other packages that begin with "llvm". (Note: There aren't any in Homebrew currently)
  -  As long as they don't have an `llvm-config`, which seems like a reasonable assumption.

PR #766, opened 2015-06-02
remove sed fallback for help/usage

jasonkarns:
The sed fallback was necessary before rbenv 0.4.0, which has been out for 2.5 years. I think it's time we can remove the fallback and just rely on rbenv-help.

### DIFF
--- a/test/rbenv.bats
+++ b/test/rbenv.bats
@@ -13,7 +13,7 @@ stub_ruby_build() {
 }
 
 extract_usage_from() {
-  local program="../bin/$1"
+  local program="$BATS_TEST_DIRNAME/../bin/$1"
   sed -ne '/^#/!q;s/.//;s/.//;1,4d;p' < "$program"
 }
 
--- a/test/rbenv.bats
+++ b/test/rbenv.bats
@@ -12,11 +12,6 @@ stub_ruby_build() {
   stub ruby-build "--lib : $BATS_TEST_DIRNAME/../bin/ruby-build --lib" "$@"
 }
 
-extract_usage_from() {
-  local program="$BATS_TEST_DIRNAME/../bin/$1"
-  sed -ne '/^#/!q;s/.//;s/.//;1,4d;p' < "$program"
-}
-
 @test "install proper" {
   stub_ruby_build 'echo ruby-build "$@"'
 
@@ -180,7 +175,7 @@ OUT
 }
 
 @test "rbenv-install has usage help preface" {
-  run extract_usage_from rbenv-install
+  run head "$(which rbenv-install)"
   assert_output_contains 'Usage: rbenv install'
 }
 
@@ -209,6 +204,6 @@ OUT
 }
 
 @test "rbenv-uninstall has usage help preface" {
-  run extract_usage_from rbenv-uninstall
+  run head "$(which rbenv-uninstall)"
   assert_output_contains 'Usage: rbenv uninstall'
 }
--- a/test/rbenv.bats
+++ b/test/rbenv.bats
@@ -158,8 +158,6 @@ OUT
 
   run rbenv-install
   assert_failure
-  assert_output_contains 'Usage: rbenv install'
-
   unstub rbenv-help
 }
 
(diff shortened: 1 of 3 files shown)


mislav:
This fails the build, presumably because `rbenv-help` is not available in the test environment. Could you define a fake rbenv-help that basically does the `sed` that you removed?

jasonkarns:
that's what i get for not running the tests...

all fixed now.

jasonkarns:
I tried stubbing rbenv-help (for both install and uninstall) in `setup` but it was throwing errors in other tests from that file. I'm not terribly familiar with bats so I didn't pursue it further. But it would probably be nice to extract the rbenv-help stubs into setup/teardown.

test/rbenv.bats
```diff
@@ -149,39 +149,63 @@ OUT
 
 @test "not enough arguments for rbenv-install" {
   stub_ruby_build
+  stub rbenv-help 'install : echo "Usage: rbenv install"'
```
mislav: Thanks, but with this stub you're just verifying that `rbenv-help` gets called. Because fake rbenv-help stubs its own output, we're never testing that this help text is indeed present within ruby-build.

Instead, try a different stubbing approach. Inside the `setup` block, create a new executable called `${TMP}/bin/rbenv-help` with the contents:

```sh
#!/bin/bash
set -e
cmd="${1?}"
file="$(which "rbenv-$cmd")"
sed -ne '/^#/!q;s/.//;s/.//;1,4d;p' < "$file"
```

This simulates real `rbenv-help` by extracting the help text off the top of a `rbenv-<command>` exectuable file.

jasonkarns: Ah, I assumed the purpose of the test was to ensure rbenv-help gets called, and that rbenv-help's own tests ensure that it does the right thing.

It sounds like this test is actually testing two things: 
1. that rbenv-help is invoked correctly and at the correct times (too few args, too many args, -h flag)
2. that ruby-build contains the correct 'usage' preface comments.

Rather than merge the two assertions into the same tests, I think I'd rather create an explicit test for 2. It would be more intention-revealing than hiding the second assertion in the body of a stub. Since we can still rely on rbenv-help to do its thing correctly, I'll create two new tests that simply assert that `sed -ne '/^#/!q;s/.//;s/.//;1,4d;p' < rbenv-{install | uninstall}` contain the correct messaging. Cool?

jasonkarns: new assertions added in 85a01065d1683a1eb768e93f3957859aa6933f8a

mislav: > Rather than merge the two assertions into the same tests, I think I'd rather create an explicit test for 2.

That works. Then keep the `stub rbenv-help` lines, but they don't have to output anything anymore. They can simply be a no-op, and the `unstub rbenv-help` line will verify that rbenv-help was called with the argument it was decleared with (either "install" or "uninstall"). Then remove the "Usage:" output assertions from all tests that don't explicitly test the help text.

jasonkarns:
there we go

test/rbenv.bats
```diff
@@ -149,39 +154,61 @@ OUT
...
   stub_ruby_build
+  stub rbenv-help 'install : true'
+
   run rbenv-install -h
   assert_success
+  unstub rbenv-help
+}
+
+@test "rbenv-install has usage help preface" {
+  run extract_usage_from rbenv-install
```
mislav: How about instead of having `extract_usage_from` function, simply do

```sh
run head -3 `which rbenv-install`
```

It seems that it could pass the test.

mislav: Of course, quote the string in case the path contains spaces:

```sh
run head -3 "$(which rbenv-install)"
```

jasonkarns:
updated per your suggestion. the two scripts would have needed `head` of at least 5 lines, so I just left it to the default 10

mislav:
OK thanks. I'll merge this in when I do the next round of rbenv + ruby-build updates.

sferik:
It’s been over a month and this looks good to me. I’m going to merge it in now.

PR #773, opened 2015-06-12
Add Rubinius 2.5.6

deepj:


deepj:
Hi, is it possible to release new version after merging this? There are many news to desire being worth merged into `ruby-build` and released.

/cc @mislav or @sferik

mislav:
Thanks. Yeah we can release after this, but I also want to get my LLVM detection fix in that release as well. I have to finalize #765

deepj:
Ok

PR #783, opened 2015-07-09
Add JRuby 9.0.0.0.rc2

deepj:


sferik:
:hankey: Just missed the latest release. :frowning:

deepj:
I make PRs fast as I can :santa:

sferik:
@deepj Thanks for all your help! I normally try to cut releases whenever there is an updated to OpenSSL. If you’re planning to open multiple pull requests in a row, maybe say something like “Planning to open two more PRs after this.” in the description and the “This is the last one for today.”

That would ensure more things make it into each releases and lower the release frequency (and update frequency for our users).

deepj:
@sferik Yes, of course. Unfortunately, I haven't planned anything. I just saw an announcement of releasing JRuby 9k rc2 on the Twitter in short time after making the OpenSSL PR.

PR #791, opened 2015-07-24
Update Rubinius download location to use https://

reedloden:
The Rubinius folk updated their download links to use https://, so let's match that now.


mislav:
Thanks! It seems like they use the same download locations officially from their blog. I hope in the future they add a redirect from their own domain.

PR #792, opened 2015-07-24
Rubygems added a redirect on their main domain, so use it as the canonical download URL

reedloden:
Swap from directly using CDN to https://rubygems.org, which just redirects to CDN (but then allows for easy change of CDN in future).


hsbt:
Thank you!

mislav:
Thanks!

PR #807, opened 2015-09-28
Add JRuby 9.0.2.0+graal-dev

pitr-ch:
the old JRuby 9.0.0.0+graal-dev installs outdated version


pitr-ch:
I've added two versions: `share/ruby-build/jruby-9.0.2.0+graal-dev` and `share/ruby-build/jruby-master+graal-dev`. `master` has an advantage that it won't have to be updated when next minor version of JRuby is out, but I am not sure if it isn't agains some policy to have explicit versions. Therefore please let me know which version would you like more and I'll update the PR. Thanks.

mislav:
Thanks!

About versions: with `-dev`, anything goes! But with stable releases, yes we do pinpoint a specific version that is supposed to never change.

pitr-ch:
Thanks! `jruby-master+graal-dev` will make it much more easier to test jruby+truffle.

PR #808, opened 2015-10-03
install OpenSSL for 1.9.3

orien:
OpenSSL is not compiled into Ruby 1.9.3 when installing on OS X 10.11 El Capitan.

To resolve this issue let's install the OpenSSL package when `has_broken_mac_openssl`.


mislav:
To clarify for others: system openssl is still present on El Capitan, and is still the "broken" 0.9.8 version, but doesn't ship with header files anymore. Therefore Ruby 1.9.3 versions would try to link to system openssl and fail because of missing header files, and the resulting build wouldn't have HTTPS capabilities.

PR #817, opened 2015-10-29
Extract assertion helpers

jasonkarns:
After working in the BATS test suites of rbenv, ruby-build and @tpope's rbenv-aliases, I began to bemoan the fact that the assertion test helpers were simply copy/pasted across these and various other repos.

While the majority of the helpers remained unchanged across the repos, there were a couple with ever-so-subtle distinctions. (Usually around optional args or slightly different error messaging)

I thought it would be beneficial if the common assertions were extracted into a separate repo that could be included into each test suite. So I created [bats-assert](https://github.com/jasonkarns/bats-assert)

The benefit to this, besides not dealing with copy/paste issues like drift and bug fixes, is having a place to document the behavior of the assertions. Most of the assertions are pretty clear, but some of the optional behaviors aren't clear without reading the source. Having a separate repo means the assertion behaviors can be documented.

The downside to exctracting the helpers is that they are pulled in as a git submodule. Even though there are a couple bash package managers popping up, I think it still makes sense to just use a submodule instead of imposing a package manager on all the maintainers of BATS test suites. Of course, if desired, the bats-assert repo can be restructured to support bpkg or basher if desired.


mislav:
Why the switch TMP → BATS_TMPDIR which makes up the most of the diff for this PR?

Not sure whether I see the benefits of extracting the assertion helpers to an external library, especially being a submodule. Effectively, they just get moved from one file to another, as far as this test suite is concerned.

I'm glad that you liked my very ad-hoc assertion helpers that much to extract them to their own project and document them, though 😸

jasonkarns:
The switch from `TMP` was for the `flunk` helper, which replaces the literal `TMP` path with `$TMP` for clarity. The bats-assert module does the same string replacement but for `BATS_TMPDIR` (since `BATS_TMPDIR` is one of the special vars that BATS makes available, and bats-assert shouldn't rely on `TMP`). Initially, I just set `BATS_TMPDIR` equal to `TMP` (c8100c21f20e30957529536d3ba71ec143e139a7) but since they represent the same thing, I figured it made more sense to just use `BATS_TMPDIR` directly. The commit that does the switch can be discarded to minimize the changes in this PR, if desired.


The goal of extracting them is twofold:

1. documentation of the helper methods
2. consistency across repos that use the same/similar helpers
    - bugs/changes can be shared without copy/paste
    - helpers from repo1 are available to repo2 whenever they are needed (e.g., assert_line, etc)

The idea being that rbenv itself (and rbenv-aliases, et. al) also be moved to use bats-assert.

I think the ideal is for bats-assert itself to be owned by @sstephenson or maybe a bats-contrib org or something. I would see similar treatment being done for the stub helpers (as bats-mock or similar). I think there is merit in having a core set of assertions that can be used consistently across projects. When I first extracted bats-assert, I found there are other projects on github that use BATS which have also copy/pasted some of these helpers (or just created eerily similar ones). As is usual with copy/paste, though, they have drifted out of sync with some of the changes to those from the rbenv family.

jasonkarns:
I should also point out that in addition to the rbenv/ruby-build family of test suites using BATS, the nodenv/node-build fork that I help maintain keeps (mostly) up to date with rbenv. Which means our test suites also use these helpers. So I have my hands in a number of the rbenv repos but doubly so with all of the nodenv repos.

jasonkarns:
Known rbenv/nodenv repos with BATS test suites and copy/pasted assertion helpers:

- rbenv
- ruby-build
- rbenv-aliases
- rbenv-bundler-ruby-version
- rbenv-env
- nodenv
- node-build
- nodenv-aliases
- nodenv-npm-rehash
- nodenv-package-rehash (already using bats-assert)

mislav:
Thanks for the research, @jasonkarns!

jasonkarns:
Closing this as I would rather see collective effort behind https://github.com/ztombol/bats-assert instead of https://github.com/jasonkarns/bats-assert

PR #819, opened 2015-10-30
Remove license from readme

jasonkarns:
LICENSE already exists as its own file (https://github.com/sstephenson/ruby-build/blob/master/LICENSE)


mislav:
Sure

PR #820, opened 2015-10-30
A lightweight version of Subversion is already installed on FreeBSD a…

silenius:
…s svnlite (the only difference from normal Subversion use is that the command name is svnlite).


mislav:
Thanks!

PR #821, opened 2015-11-04
TRAVIS_COMMIT_RANGE is correct now

jasonkarns:
According to https://github.com/travis-ci/travis-ci/issues/1719 (closed by https://github.com/travis-ci/travis-core/pull/383), the behavior of $TRAVIS_COMMIT_RANGE should be correct now. We should be able to use it directly.

### DIFF
--- a/script/test
+++ b/script/test
@@ -4,11 +4,6 @@ set -e
 STATUS=0
 bats ${CI:+--tap} test || STATUS="$?"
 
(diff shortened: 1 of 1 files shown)


sferik:
:+1: Tests pass! Thanks for the update.

mislav:
Thanks for the update! Technically the block was a no-op if `TRAVIS_COMMIT_RANGE` was populated, but lets hope Travis has solved this once and for all and that we won't need the workaround anymore.

PR #830, opened 2015-11-15
Allow for space-separated LDFLAGS

andersr:
Prevented errors caused by a space between `-L` and the path by not creating a directory if the arg is `-L`

The base case of the switch statement will be called with either `-L/path/` or `/path/`. 
`mkdir` will then be called with that argument after the `-L` has been removed.

Fixes https://github.com/sstephenson/ruby-build/issues/829

P.s. this is my first pull request to an open source repo :-)

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -968,11 +968,13 @@ build_package_verify_openssl() {
 }
 
 # Ensure that directories listed in LDFLAGS exist
+# $LDFLAGS must be in format '-L /path/' or '-L/path/'
 build_package_ldflags_dirs() {
   local arg
   for arg in $LDFLAGS; do
     case "$arg" in
-    -L* ) mkdir -p "${arg#-L}" ;;
+    -L ) ;;
+    * ) mkdir -p "${arg#-L}" ;;
     esac
   done
 }


bin/ruby-build
```diff
@@ -968,11 +968,13 @@ build_package_verify_openssl() {
...
 
 # Ensure that directories listed in LDFLAGS exist
+# $LDFLAGS must be in format '-L /path/' or '-L/path/'
 build_package_ldflags_dirs() {
   local arg
   for arg in $LDFLAGS; do
     case "$arg" in
-    -L* ) mkdir -p "${arg#-L}" ;;
+    -L ) ;;
+    * ) mkdir -p "${arg#-L}" ;;
```
mislav: Handling the `*` argument (which matches any string) is not optimal here because LDFLAGS might have arguments that are different from `-L` and directory names. I've pushed a commit that fixes the original issue, but ignoring any potential arguments that might not be related to `-L`.

mislav:
Congrats on your first pull request! I've left an inline comment about my concerns about argument flag matching. I had an idea how to handle this better, and I've pushed that and an automated test for it.

Sorry that your PR didn't make it in this time, but I promise to assist you more with your next one. :wink:

PR #839, opened 2015-12-02
safer teardown

jasonkarns:
Ensure that if `TMP` is empty, we don't do bad things, like `rm -rf /*`


hsbt:
it's reasonable change. Thanks!

mislav:
:+1:

PR #844, opened 2015-12-11
Added definition of 2.3.0-preview2

hsbt:
See. http://blade.nagaokaut.ac.jp/cgi-bin/scat.rb/ruby/ruby-core/72045


jeremy:
:metal:

PR #845, opened 2015-12-11
use official CDN for ruby

hsbt:
I have been prepared official CDN named "https://cache.ruby-lang.org" using HTTPS. We should use official CDN instead of third party's CDN. cache.ruby-lang.org provided by fastly has desitributed function for world wide.

@mislav @sferik How do you think this?

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1189,7 +1189,7 @@ else
 fi
 
 if [ -z "$RUBY_BUILD_MIRROR_URL" ]; then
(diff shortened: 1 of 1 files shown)


mislav:
Is ruby-lang.org comfortable hosting not just Ruby downloads but all related things like Rubinius, JRuby and copies of similar downloads?

We would also have to update credentials in Travis for auto-updating the mirror for new definitions.

hsbt:
cache.ruby-lang.org is not support to host Rubinius and JRuby. ok I will make another patchs for download only CRuby via cache.ruby-lang.org.

PR #846, opened 2015-12-11
Don't use cloudfront for cache.ruby-lang.org

nurse:
cache.ruby-lang.org is CDN powered by Fastly.
https://www.ruby-lang.org/en/about/website/

```
% host cache.ruby-lang.org
cache.ruby-lang.org is an alias for j.global-ssl.fastly.net.
j.global-ssl.fastly.net has address 103.245.222.68
```

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -327,7 +327,14 @@ fetch_tarball() {
     package_url="${package_url%%#*}"
 
     if [ -n "$RUBY_BUILD_MIRROR_URL" ]; then
-      mirror_url="${RUBY_BUILD_MIRROR_URL}/$checksum"
+      case "$package_url" in
+      https://cache.ruby-lang.org/*)
+        # This is Fastly.
+        ;;
+      *)
+        mirror_url="${RUBY_BUILD_MIRROR_URL}/$checksum"
+        ;;
+    esac
     fi
   fi
 


mislav:
Some people will want to specify their own mirror, for instance to cache downloads within their local network. They will want cache.ruby-lang.org downloads to be pulled from the custom mirror as well. This patch prevents that.

Also, the auto-mirroring script would have to be updated to exclude mirroring new cache.ruby-lang.org tarballs if we indeed want to blacklist this. But I'm not sure if it's worth the trouble.

@hsbt: Why are we doing all this again?

miyagawa:
I was working on a similar patch https://github.com/miyagawa/ruby-build/commit/01b03908600d97ec7571bf488c1882d60d2586d2 and it addresses @mislav's concern by honoring `RUBY_BUILD_MIRROR_URL` if it's set by the user. My patch bypasses the cloudfront host only if a) there's no `RUBY_BUILD_MIRROR_URL` set and b) the tarball is for cruby.

I don't think that the mirroring script needs to be updated, there's no need to blacklist cruby tarballs. We just want CDN be used for its right purpose - putting a cache in front of another CDN makes no sense.

hsbt:
@mislav I understand your concern. If user set their local mirror and set `RUBY_BUILD_MIRROR_URL`, ruby-build should use it. @miyagawa 's patch solve this situation and use CRuby's official mirror when ruby-build downloads CRuby.

I hope to merge https://github.com/miyagawa/ruby-build/commit/01b03908600d97ec7571bf488c1882d60d2586d2

mislav:
Please don't merge this patch as-is. It will skip people's custom mirror configuration for CRuby. People who need custom mirror support will be disappointed.

mislav:
Ah I see. @miyagawa's patch is not in this PR, but is better. :+1: for something in that direction. Possibly tests would be in order?

miyagawa:
@mislav thanks for the feedback! I opened #847 for this.

mislav:
Closing in favor of #847

PR #847, opened 2015-12-11
Skip mirrors for CRuby since they're on CDN

miyagawa:
Similar to #846 but addresses @mislav's concerns on custom mirror users :)

This patch skips using the default mirror for CRuby tarballs, because they're hosted on its own CDN.

This PR does nothing for users who currently set `RUBY_BUILD_MIRROR_URL` to cache downloads for instance in their local mirror. The mirror is skipped only when a) mirror URL is not specified and b) the tarball is on cahe.ruby-lang.org, the official CDN for ruby binary downloads.

Rationale:
* #240 shows that the mirror support came out of a frustration of waiting for slow downloads of ftp.ruby-lang.org. That's not a problem now!
* I know we validate all downloads via shasum check, but the console output of  `https://dqw8nmjcqpjn7.cloudfront.net/<shasum>` is just not a good user experience over `https://cache.ruby-lang.org/.../ruby-2.3.0.tar.gz`


miyagawa:
@mislav @hsbt I have reworked the patch to make the decision to skip mirrors now data driven in definition files. I was a bit reluctant to introduce a new directive, but this is for the best to keep backward compatibilities with the existing custom definition files people might be using today.

I added tests and updated documentation as well.

bin/ruby-build
```diff
@@ -154,6 +154,10 @@ num_cpu_cores() {
   echo "${num:-2}"
 }
 
+install_package_direct() {
```
mislav: I'm not a big fan of this change. `install_package_direct` is a misnomer when you're using your own custom mirror— it's not "direct" in that case.

I was fine with the default mirror being automatically skipped for `cache.ruby-lang.org` stuff. I don't think that the user should necessarily be conscious about this "magic" happening under the hood. All they need to know is that their downloads will stay fast by default.

miyagawa: @mislav understood, i just reverted to the original method of ad-hoc check with the cache.ruby-lang.org. the improvements to test and doc are still in place.

bin/ruby-build
```diff
@@ -327,7 +327,10 @@ fetch_tarball() {
     package_url="${package_url%%#*}"
 
     if [ -n "$RUBY_BUILD_MIRROR_URL" ]; then
-      mirror_url="${RUBY_BUILD_MIRROR_URL}/$checksum"
+      if (echo "$package_url" | grep -v "cache.ruby-lang.org" > /dev/null) ||
```
mislav: Check could be expressed with

```sh
[[ $package_url == */cache.ruby-lang.org/* ]]
```

bin/ruby-build
```diff
@@ -1190,6 +1193,7 @@ fi
 
 if [ -z "$RUBY_BUILD_MIRROR_URL" ]; then
   RUBY_BUILD_MIRROR_URL="https://dqw8nmjcqpjn7.cloudfront.net"
+  RUBY_BUILD_DEFAULT_MIRROR=1
```
mislav: This is OK, but the `else` branch of this conditional has to also set `RUBY_BUILD_DEFAULT_MIRROR`, but to an empty string. This is so that RUBY_BUILD_DEFAULT_MIRROR can't accidentally be inherited from the environment.

miyagawa: Done a09ecc2

bin/ruby-build
miyagawa: Done 326734e

bin/ruby-build
```diff
@@ -327,7 +327,10 @@ fetch_tarball() {
     package_url="${package_url%%#*}"
 
     if [ -n "$RUBY_BUILD_MIRROR_URL" ]; then
-      mirror_url="${RUBY_BUILD_MIRROR_URL}/$checksum"
+      if ! [[ $package_url == */cache.ruby-lang.org/* ]] ||
```
mislav: Nitpick: I think the condition would more easier read as:

```sh
if [[ -z "$RUBY_BUILD_DEFAULT_MIRROR" || $package_url != */cache.ruby-lang.org/* ]]; then
```

test/mirror.bats
```diff
@@ -101,3 +101,20 @@ export RUBY_BUILD_MIRROR_URL=http://mirror.example.com
...
+
+
+@test "package URL with ruby-lang CDN with default mirror URL will bypasses mirror" {
+  export RUBY_BUILD_MIRROR_URL=
+  local checksum="ba988b1bb4250dee0b9dd3d4d722f9c64b2bacfc805d1b6eba7426bda72dd3c5"
+
+  stub shasum true "echo $checksum"
+  stub curl "-q -o * -*S* https://cache.ruby-lang.org/* : cp $FIXTURE_ROOT/\${5##*/} \$3"
+
+  install_fixture definitions/skip-mirror
```
mislav: Nitpick: instead of saving the fixture into a new file which will only be used once here, you can use `run_inline_definition`

```sh
  run_inline_definition <<DEF
install_package "package-1.0.0" "https://cache.ruby-lang.org/packages/package-1.0.0.tar.gz#ba988b1bb4250dee0b9dd3d4d722f9c64b2bacfc805d1b6eba7426bda72dd3c5" copy
DEF
```

test/mirror.bats
```diff
@@ -101,3 +101,20 @@ export RUBY_BUILD_MIRROR_URL=http://mirror.example.com
...
+  echo "$output" >&2
```
mislav: This is not necessary for the test, right?

mislav: Ah, I see it's being used in other tests for some reason (probably debugging). Nevertheless, it's not needed.

test/mirror.bats
```diff
@@ -101,3 +101,20 @@ export RUBY_BUILD_MIRROR_URL=http://mirror.example.com
...
+  [ "$status" -eq 0 ]
```
mislav: `assert_success` reads nicer

test/mirror.bats
```diff
@@ -101,3 +101,20 @@ export RUBY_BUILD_MIRROR_URL=http://mirror.example.com
...
+  [ -x "${INSTALL_ROOT}/bin/package" ]
```
mislav: Prefix this line with `assert` to get a better message when this line fails

mislav:
I've had a few more nitpicks, but overall looks :+1: 

Thanks for being patient.

miyagawa:
@mislav feedback greatly appreciated! I reflected all the points you made. Let me know if it needs a squash, otherwise, merge away :)

PR #852, opened 2015-12-17
Convert MRI definitions to use bz2 tarballs instead of gz ones

deepj:
`bzip2` offers more effective compression algorithm than `gzip`. An user can download less data.


mislav:
Good idea; thanks!

PR #854, opened 2015-12-25
Ruby 2.3.0 branch and trunk

hsbt:
We branched stable version of ruby 2.3.0 today. I switched it in `2.3.0-dev` definition.

and added 2.4.0-dev definition. It version still uses `2.3.0`. but matz will bump version `2.4.0` at few days later :octocat:


sferik:
:tada: :christmas_tree: :gift: :confetti_ball:

hsbt:
Thank you! We prepare 2.3.0 package now. I will add its definition later.

PR #857, opened 2015-12-26
Add rbx-2.6.0, rbx-2.7.0, rbx-2.8.0, rbx-2.9.0

mislav:
Also, support Rubinius linking against Homebrew openssl due to OS X not shipping with system OpenSSL headers anymore.


hsbt:
:+1:

PR #863, opened 2016-01-01
Remove Rubinius

sferik:


hsbt:
`require_llvm` and `locate_llvm` should be removed too.

jasonkarns:
Is it really prudent to remove the functions from ruby-build itself? One key feature of ruby-build is that the definition lookup path can be configured, so one can provide their own custom definitions. Even if the rubinius definitions are removed, removing the rubinius-specific functions would break any custom definitions that rely on them. This would be a pretty big compatibility-breaking change. Might there be value in keeping the rubinius functions a while longer? Or perhaps a deprecation phase would be beneficial to warn users that their definitions would break soon (with whatever release would actually remove the functions, assuming the rubinius removal actually moves forward).

peteygao:
What is the reasoning for dropping Rubinius support in `ruby-build`? I actively use Rubinius and have in the past week installed Rubinius 3.16 via `ruby-build`. In fact I would have installed the latest version `3.19` if #911 was already merged.

hsbt:
I will close this.

We can passively maintain rubinius. ref. https://github.com/rbenv/ruby-build/pull/1035

PR #864, opened 2016-01-01
Remove builds of JRuby that bundle Graal

chrisseaton:
Thanks very much for accepting our definition files for JRuby that bundled Graal (our new JIT compiler for Java). This was really useful when there was no other way to get Graal, but now there are official experimental releases available from Oracle https://github.com/jruby/jruby/wiki/Downloading-GraalVM and soon it will be possible to just use an early access of Java 9 https://github.com/jruby/jruby/wiki/Using-Graal-in-JDK-9-EA-Builds, so these bundled versions aren't necessary any more and we can remove them.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -639,29 +639,10 @@ build_package_jruby() {
   fix_jruby_shebangs
 }
 
-graal_architecture() {
-  if [ "$(uname -m)" != "x86_64" ]; then
-    echo "no nightly builds available" >&2
-    exit 1
-  fi
-
-  case "$(uname -s)" in
-  "Darwin") echo "macosx-x86_64";;
-  "Linux") echo "linux-x86_64";;
-  *)
-    echo "no nightly builds available" >&2
-    exit 1;;
-  esac
-}
-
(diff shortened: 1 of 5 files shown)


mislav:
Thanks for the update. What would be the steps for ruby-build/rbenv users to continue trying out the GraalVM after this is removed? Could we instead update the definitions to download the new build from Oracle?

chrisseaton:
I'm afraid the new builds are click-through licensed as we're now basing our public builds on Oracle's JDK rather than Open JDK.

I've put up an FAQ with the answer for anyone saying 'how do I do it now?'

https://github.com/jruby/jruby/wiki/Truffle-FAQ#how-do-i-get-jrubytruffle

jasonkarns:
@chrisseaton would it make sense to restore these since graalvm has a community edition now?

chrisseaton:
You can now `rbenv install truffleruby-1.0.0-rc10` (and the same with other major Ruby version managers).

TruffleRuby isn't part of JRuby any more, so specifically build of JRuby with Truffle in it isn't a thing any more.

Issue #862, opened 2015-12-31
Remove Rubinius


sstephenson:
This issue will track the removal of Rubinius from ruby-build, as requested by @brixen, Rubinius’ maintainer, in https://github.com/rbenv/ruby-build/pull/860#issuecomment-168059883:

> Honestly, given the amount of time I've wasted with rbenv, RVM, and chruby bugs and insisting on telling me how Rubinius should build and package, it would be better if none of them installed Rubinius. 
- [ ] Remove rbx-\* definition files
- [ ] Remove Rubinius support functions
- [ ] Document the reasoning behind this change on the wiki

brixen:
:+1:

sferik:
I’ve opened https://github.com/rbenv/ruby-build/pull/863, which should check the first two boxes.

As for the third, I don’t know what to say on wiki besides, “Support for Rubinius was removed at the request of its maintainer” with a link back to [this comment](https://github.com/rbenv/ruby-build/pull/860#issuecomment-168059883). I guess I’ll start with that and others can edit it if they have more to say.

mislav:
I'm a little bit sad that this would undo many hours or debugging, user support, and workarounds that I've invested in bringing easier Rubinius support to ruby-build and rbenv. Examples:

* 3e5c31f - Support Rubinius linking against Homebrew openssl
* 3b6f96d - Lookup LLVM 3.5 from Homebrew on `require_llvm` on OS X
* 9a5256e - Work around Rubinius LLVM incompatibilities on Yosemite
* 392679a - Fix `irb`, `rake`, `rdoc`, `ri` for rbx-2.2.1

However, even if after all this, our rbx support is deemed unwelcome, then I'm :+1: if someone is going to put in work to remove this, document it, and provide support to people who will undoubtedly come to the Issues section to ask about this.

brixen:
I didn't request that Rubinius be removed; that's @sstephenson's revisionism, just as this is https://twitter.com/sstephenson/status/682300372766068737.

What I said was, "given the amount of time I've wasted with rbenv, RVM, and chruby bugs and insisting on telling me how Rubinius should build and package, it would be better if none of them installed Rubinius."

You are welcome to choose to remove support for Rubinius, just as you are welcome to follow the repeated requests from me to follow my recommendations for the project. Take ownership for your decision.

brixen:
[...]

PR #866, opened 2016-01-04
Add metadata for Rubinius 3.0 aka rbx-3.0.0.

keltia:
I'm aware of issue #862 but in the meantime, hoping it will get resolved w/o removing rubinius support, here is the file for rbx-3.0.0 — a simple copy from 2.11.0 w/ SHA and filenames edited. It builds and installs fine.

Thanks.


mislav:
Thanks!

Issue #870, opened 2016-01-07
Can't install on CentOS 7


jandudulski:
```bash
$ rbenv install -k 2.3.0
/home/user/.rbenv/plugins/ruby-build/bin/ruby-build: line 207: pushd: ruby-2.3.0: No such file or directory

BUILD FAILED (CentOS Linux 7 using ruby-build 20151230)
```

mislav:
That's bizarre. Is there really no other output? Is the error the same every time?

jimnanney:
BUILD FAILED
tar: Child returned status 2
tar: Error is not recoverable: exiting now
(Red Hat Enterprise Linux Server 7.2 using ruby-build 20151230)
 
Inspect or clean up the working tree at /opt/james.nanney/tmp/ruby-build.20160107094924.4078
Results logged to /opt/james.nanney/tmp/ruby-build.20160107094924.4078.log
 
Last 10 log lines:
/opt/james.nanney/tmp/ruby-build.20160107094924.4078 ~
tar (child): bzip2: Cannot exec: No such file or directory
tar (child): Error is not recoverable: exiting now
tar: Child returned status 2
tar: Error is not recoverable: exiting now

Looks like bzip2 is missing from my rhel 7 install, maybe same on centos?

jimnanney:
sudo yum install -y bzip2 worked for me.

mislav:
Ouch, thanks for reporting. This is probably caused by #852 /cc @deepj 

I had no idea that some platforms don't support bz2 decompression out of the box through tar. I'm not sure what to do about this. We could definitely try to improve the error message so at least you get a hint that you should install bz2. Or, we could try falling back transparently to gz tarballs? However, I'm not sure if every `*.bz2` download within ruby-build has a corresponding `.tar.gz` download.

jimnanney:
Forked and added a warning to the log when bzip2 is not in path, how do I run the test suite?

deepj:
@mislav This suprises me. If I'm not wrong RVM uses bz2 tarballs as well.

mislav:
> how do I run the test suite?

Yeah, sorry that this is not more clear. I add `bats` to my PATH:

```sh
ln -s /path/to/bats ~/bin/bats
```

Then simply `bats test` in the ruby-build project.

jimnanney:
Thanks! I'm not sure how to write a test for it, but sending a pull request to echo a warning to the log.

jandudulski:
[...]

PR #871, opened 2016-01-07
add message when bzip2 is missing and package type is bz2

jimnanney:
When installing on a system without bzip2, build would fail with little help. Added a message to the log file that states bzip2 is missing from path, and when build fails, try installing bzip2.

#870 has further details

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -342,6 +342,9 @@ fetch_tarball() {
   local package_filename="${package_name}.tar.gz"
 
   if [ "$package_url" != "${package_url%bz2}" ]; then
+    if ! type -p bzip2 >/dev/null; then
+      echo "warning: bzip2 not found; consider installing \`bzip2\` package" >&4
+    fi
     package_filename="${package_filename%.gz}.bz2"
     tar_args="${tar_args/z/j}"
   fi


bin/ruby-build
```diff
@@ -342,6 +342,10 @@ fetch_tarball() {
   local package_filename="${package_name}.tar.gz"
 
   if [ "$package_url" != "${package_url%bz2}" ]; then
+    if [ -z "$(which bzip2)" ]; then
```
mislav: This check could be `if ! type -p bzip2 >/dev/null; then`. It's the way we do such checks elsewhere in the codebase.

bin/ruby-build
```diff
@@ -342,6 +342,10 @@ fetch_tarball() {
...
+      echo "bzip pacakage detected, however no bzip2 found within path" >&4
+      echo "If build fails, consider installing bzip2 pacakge" >&4
```
mislav: There are some spelling errors in the warning message. Could we rewrite the warning message as just one line, maybe something like:

> warning: bzip2 not found; consider installing \`bzip2\` package

jimnanney:
Thanks for the feedback. Never considered type before.

bin/ruby-build
```diff
@@ -342,6 +342,9 @@ fetch_tarball() {
   local package_filename="${package_name}.tar.gz"
 
   if [ "$package_url" != "${package_url%bz2}" ]; then
+    if ! type -p bzip2 >/dev/null; then
+      echo "warning: bzip2 not found; consider installing `bzip2` package"
```
mislav: One more thing: you need to escape both `` ` `` because within double quotes they are considered shelling out to a process.

jimnanney:
Thanks again, cleared up the quotes and squashed it to one commit. Sorry to make so many mistakes. Really appreciate all of your help and review.

PR #877, opened 2016-01-11
Explicitly specify destination refspec

### LINKED ISSUE #876: brew-publish script incorrectly uses branch upstream
First off, I realize this is an odd issue for a number of reasons.

As I'm not a maintainer of ruby-build, I don't do any ruby-build releases, and so I don't run the brew-publish script for ruby-build. I am, however, a maintainer of node-build which, being a direct port of ruby-build, has an identical brew-publish script which I *do* run as part of node-build's release process.

The issue I have is when the branch is created (`git checkout -q -B "$branch" origin/master` https://github.com/rbenv/ruby-build/blob/master/script/brew-publish#L37). On my machine, doing a branch creation/checkout with a remote branch as its starting point automatically sets up the remote branch as the upstream for the branch. Thus, the subsequent push (`git push -u "$fork_remote" "$branch"`) attempts to push the branch to `$fork_remote/master`, rather than pushing to `origin/$branch`

``` sh
[system] [2.1.5] /usr/local (node-build-v2.1.1 $> origin/master)
$ git push -u jasonkarns node-build-v2.1.1 
To git@github.com:jasonkarns/homebrew.git
 ! [rejected]        node-build-v2.1.1 -> master (non-fast-forward)
error: failed to push some refs to 'git@github.com:jasonkarns/homebrew.git'
hint: Updates were rejected because a pushed branch tip is behind its remote
hint: counterpart. Check out this branch and integrate the remote changes
hint: (e.g. 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```

I'm on git version 2.6.4. I'm not sure if it's my version of git that is doing the automatic upstream setup on `git checkout -B`, or if there's a config setting somewhere doing it (my 5 minute spelunking didn't uncover anything obvious).

I do have `push.default` set to `upstream`, so when the remote branch isn't specified, it chooses the configured upstream as the target (which during checkout is set to `master`). I'm assuming that the current script relies on everyone using the default configuration of `simple`?

Do any of the ruby-build maintainers have this issue?

I can submit a PR that makes the destination refspec explicit (as below, though this is untested)

``` diff
- git push -u "$fork_remote" "$branch"
+ git push -u "$fork_remote" "$branch":$"branch"
```

But I wanted to confer first due to the extenuating circumstances of this being a maintainer-run script.

jasonkarns:
closes #876

### DIFF
--- a/script/brew-publish
+++ b/script/brew-publish
@@ -46,7 +46,7 @@ rm -f "${formula}.bak"
 git commit -m "${brew_name} ${version#v}" -- "$formula"
 # hackish way of getting the git remote name for user's fork
 fork_remote="$(hub fork 2>&1 | grep -oE 'remote:? \S+' | tail -1 | awk '{print $2}')"
-git push -u "$fork_remote" "$branch"
+git push -u "$fork_remote" HEAD
 hub pull-request -m "${brew_name} ${version#v}"
 
 git checkout -q -


script/brew-publish
```diff
@@ -46,7 +46,8 @@ rm -f "${formula}.bak"
 git commit -m "${brew_name} ${version#v}" -- "$formula"
 # hackish way of getting the git remote name for user's fork
 fork_remote="$(hub fork 2>&1 | grep -oE 'remote:? \S+' | tail -1 | awk '{print $2}')"
-git push -u "$fork_remote" "$branch"
+git push -u "${fork_remote}" "${branch}:${branch}"
```
mislav: Since we're currently on the branch that we want to publish, how about using `HEAD` as refspec? Git documentation says that it will always publish the same-named branch on the remote.

jasonkarns: cool. updated as such (and tested via latest node-build release. works as advertised)

script/brew-publish
```diff
@@ -46,7 +46,7 @@ rm -f "${formula}.bak"
 git commit -m "${brew_name} ${version#v}" -- "$formula"
 # hackish way of getting the git remote name for user's fork
 fork_remote="$(hub fork 2>&1 | grep -oE 'remote:? \S+' | tail -1 | awk '{print $2}')"
-git push -u "$fork_remote" "$branch"
+git push -u "${fork_remote}" HEAD
```
mislav: Ummm why the curly braces now? #nitpick

jasonkarns: consistency with https://github.com/rbenv/ruby-build/pull/877/files#diff-b746614a124d886e7bcb47a43fce3a17R46 and https://github.com/rbenv/ruby-build/pull/877/files#diff-b746614a124d886e7bcb47a43fce3a17R50

Can remove if desired

mislav: Both of the other strings have string interpolation with multiple variables; that's why they have curl braces

jasonkarns: I figured those were stylistic since the braces aren't strictly necessary (due to the spaces).

braces now off!

mislav: Yeah they are stylistic because it's easier to read where string interpolation begins and ends. It's never neccessary when just quoting a single variable, though.

PR #879, opened 2016-01-15
Clean up usage documentation

jasonkarns:
- Add missing `-4` and `-6` flags to ruby-build usage
- Add missing `--version` flag to ruby-build and rbenv-install usage
- Order flags alphabetically
- Add missing `-s` flag to rbenv-install usage

How should rbenv-install usage show that `-f` and `-s` are mutually exclusive? (Or rather that `-s` actually takes precedence?) Related, should the code be changed to error out (with usage help) when both `-f` and `-s` are present?

### DIFF
--- a/bin/rbenv-install
+++ b/bin/rbenv-install
@@ -5,6 +5,7 @@
 # Usage: rbenv install [-f|-s] [-kpv] <version>
 #        rbenv install [-f|-s] [-kpv] <definition-file>
 #        rbenv install -l|--list
+#        rbenv install --version
 #
 #   -l/--list          List all available versions
 #   -f/--force         Install even if the version appears to be installed already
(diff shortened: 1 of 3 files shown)


bin/rbenv-install
```diff
@@ -2,9 +2,10 @@
 #
 # Summary: Install a Ruby version using ruby-build
 #
-# Usage: rbenv install [-f] [-kvp] <version>
-#        rbenv install [-f] [-kvp] <definition-file>
+# Usage: rbenv install [-fs] [-kpv] <version>
```
mislav: Use `[-f|-s]` notation to indicate that they are mutually exclusive

bin/ruby-build
```diff
@@ -1,14 +1,16 @@
 #!/usr/bin/env bash
 #
-# Usage: ruby-build [-kvp] <definition> <prefix>
+# Usage: ruby-build [-kpv46] <definition> <prefix>
```
mislav: I wouldn't add further documentation for `-46` because those are flags that I've never approved and am not sure what are they even used for. https://github.com/rbenv/ruby-build/pull/599#issuecomment-151804316

jasonkarns: Is there a desire to either fix or remove support for those flags?

mislav: I believe they "work", but I'm just not sure what they are doing and when should they be used.

jasonkarns:
updated

mislav:
Thanks!

PR #880, opened 2016-01-15
Rbenv cleanup

jasonkarns:
rbenv 1.0 now handles `--help` automatically so we don't need to handle it manually. (also remove the tests for it) Downside is that rbenv doesn't handle `-h`. However, I think it's preferable to lose the `-h` flag and slim the code down. Doing so also makes rbenv-install/rbenv-uninstall more consistent with the other rbenv plugins that have only ever supported `--help`.

Additionally, rbenv ensures `RBENV_ROOT` is set for the plugins, so there is no need to provide the `~/.rbenv` fallback.

### DIFF
--- a/bin/rbenv-install
+++ b/bin/rbenv-install
@@ -74,9 +74,6 @@ unset HAS_PATCH
 parse_options "$@"
 for option in "${OPTIONS[@]}"; do
   case "$option" in
-  "h" | "help" )
-    usage 0
-    ;;
   "l" | "list" )
     echo "Available versions:"
     definitions | indent
--- a/bin/rbenv-install
+++ b/bin/rbenv-install
@@ -24,10 +24,6 @@
 set -e
 [ -n "$RBENV_DEBUG" ] && set -x
 
-if [ -z "$RBENV_ROOT" ]; then
-  RBENV_ROOT="${HOME}/.rbenv"
-fi
-
 # Add `share/ruby-build/` directory from each rbenv plugin to the list of
 # paths where build definitions are looked up.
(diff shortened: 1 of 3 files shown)


mislav:
I don't want to drop `-h` because that would be a backwards-incompatible change. So, since we have to support `-h`, let's just keep supporting `--help` as well since it's not that much extra code. The bulk of the logic is in `usage()` within ruby-build which we have to keep anyway.

jasonkarns:
The main reason I wanted to kill the help processing (aside from just making the plugins consistent with other rbenv plugins), is that the `usage` function is copypasta in both rbenv-install and rbenv-uninstall. Since ruby-build's `usage` must work outside of rbenv, it can't rely on just calling `rbenv help <cmd>`.

With the `--help` processing removed from `-install` and `-uninstall`, then the logic of `usage` within those plugins can be slimmed down. It would no longer need to branch on the optional status code argument. Every invocation of `usage` would print to `STDERR` and `exit 1`, which trims down not only the `usage` function itself (removing a conditional), but also every invocation of it (no more args, no more output redirection).

This additional cleanup is in yet another branch locally, that I intended to submit as a PR if this one were merged. What I'll do is extract the `RBENV_ROOT` removal into a separate PR, and then merge in the extra usage cleanup so that we can get a better picture of what this change looks like holistically.

mislav:
> With the `--help` processing removed from `-install` and `-uninstall`, then the logic of `usage` within those plugins can be slimmed down.

Sure but `usage` logic is really simple right now and we don't actually gain much from cleaning it up.

PR #881, opened 2016-01-16
Add sha2 512 support for rubinius

### LINKED ISSUE #878: Sha 512 support
While working on the [rubinius scraper](https://github.com/jasonkarns/ruby-build-update-defs/pull/2) I realized that ruby-build doesn't yet have support for sha512.

While sha512 support isn't *solely* for rubinius support (which is still in question, maybe? #863), lack of sha512 is a blocker for the rubinius scraper, since rubinius only publishes sha512 checksums for their builds. (I'd prefer the scraper not generate its own checksums, but pull them from the host site.)

I'm happy to work on a PR, but opened this issue in case someone else has some cycles to spare.

jasonkarns:
closes #878

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -212,7 +212,7 @@ make_package() {
   popd >&4
 }
 
-compute_sha2() {
+compute_sha2_256() {
   local output
   if type shasum &>/dev/null; then
     output="$(shasum -a 256 -b)" || return 1
@@ -223,6 +223,23 @@ compute_sha2() {
     echo "${output##* }"
   elif type sha256sum &>/dev/null; then
     output="$(sha256sum --quiet)" || return 1
+    echo "${output% *}"
+  else
+    return 1
+  fi
+}
+
+compute_sha2_512() {
+  local output
+  if type shasum &>/dev/null; then
+    output="$(shasum -a 512 -b)" || return 1
+    echo "${output% *}"
+  elif type openssl &>/dev/null; then
+    local openssl="$(command -v "$(brew --prefix openssl 2>/dev/null || true)"/bin/openssl openssl | head -1)"
+    output="$("$openssl" dgst -sha512 2>/dev/null)" || return 1
+    echo "${output##* }"
+  elif type sha512sum &>/dev/null; then
+    output="$(sha512sum --quiet)" || return 1
     echo "${output% *}"
   else
     return 1
@@ -263,7 +280,9 @@ verify_checksum() {
   # MD5
   32) checksum_command="compute_md5" ;;
   # SHA2 256
-  64) checksum_command="compute_sha2" ;;
+  64) checksum_command="compute_sha2_256" ;;
+  # SHA2 512
+  128) checksum_command="compute_sha2_512" ;;
   # unknown checksum algorithm, return failure
   *) return 1 ;;
   esac
@@ -1225,7 +1244,7 @@ else
   RUBY_BUILD_DEFAULT_MIRROR=
 fi
 
-if [ -n "$RUBY_BUILD_SKIP_MIRROR" ] || ! has_checksum_support compute_sha2; then
+if [ -n "$RUBY_BUILD_SKIP_MIRROR" ] || ! has_checksum_support compute_sha2_256; then
   unset RUBY_BUILD_MIRROR_URL
 fi
 
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -246,16 +246,16 @@ compute_md5() {
 
 verify_checksum() {
   local checksum_command
+  local filename="$1"
+  local expected_checksum="$(echo "$2" | tr [A-Z] [a-z])"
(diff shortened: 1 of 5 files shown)


mislav:
I don't get how this "adds" SHA2 support when we had support for it previously. This to me seems just like a refactoring. Care to elaborate what feature does this add?

jasonkarns:
It adds support for 512 bit checksums. Presently only 256 bit checksums are
supported. Rubinius only publishes 512 bit checksums on their site. So
currently the 256 bit checksum must be computed when the definition file is
created. Supporting 512 bit allows the scraper to generate a definition
file based solely on the static metadata published on rubunius' release
page, without needing to compute the checksums on demand.

mislav:
So, at which point does this call `shasum -a 512`?

jasonkarns:
Maybe I should push the whole branch :facepalm:

jasonkarns:
sorry about that

test/cache.bats
```diff
@@ -57,14 +56,15 @@ setup() {
   export RUBY_BUILD_SKIP_MIRROR=
   local checksum="ba988b1bb4250dee0b9dd3d4d722f9c64b2bacfc805d1b6eba7426bda72dd3c5"
 
-  stub shasum true "echo invalid" "echo $checksum"
+  stub shasum true true "echo invalid" true "echo $checksum"
```
mislav: It's hard to follow all these arguments to stubbed `shasum`, and the extra `true`s only make it worse. Can we avoid adding them? What causes the necessity for extra arguments?

test/fixtures/definitions/with-invalid-checksum
```diff
@@ -1 +1 @@
-install_package "package-1.0.0" "http://example.com/packages/package-1.0.0.tar.gz#invalid" copy
+install_package "package-1.0.0" "http://example.com/packages/package-1.0.0.tar.gz#invalidbb4250dee0b9dd3d4d722f9c64b2bacfc805d1b6eba7426bda72dd3c5" copy
```
mislav: Can the invalid checksum be spaced with zeroes instead so that it's clear that it's not an actual checksum? Will be nicer to read I think.

bin/ruby-build
```diff
@@ -229,6 +229,23 @@ compute_sha2() {
...
+  local output
+  if type shasum &>/dev/null; then
+    output="$(shasum -a 512 -b)" || return 1
+    echo "${output% *}"
+  elif type openssl &>/dev/null; then
+    local openssl="$(command -v "$(brew --prefix openssl 2>/dev/null || true)"/bin/openssl openssl | head -1)"
+    output="$("$openssl" dgst -sha512 2>/dev/null)" || return 1
+    echo "${output##* }"
+  elif type sha512sum &>/dev/null; then
+    output="$(sha512sum --quiet)" || return 1
```
mislav: Is there evidence that a `sha512sum` exists on some systems?

mislav:
The refactoring looks good, but due to #862 I'm not sure whether we should be adding 512 support if we might be dropping Rubinius altogether.

Also, Rubinius publishes checksum from the same location it provides its main downloads. Those checksums are useless, because if the downloads get compromised, then checksums will naturally be compromised as well. That's why I'm not so excited about adding 512 support, but I'm :+1: on refactoring checksum code if it's going to be cleaner.

mislav:
Also, [`script/mirror` assumes](https://github.com/rbenv/ruby-build/blob/288261985c210d355e6cd2d0ab4a89861beafa19/script/mirror#L31-L32) that all checksums in ruby-build used as mirror location are sha256.

test/cache.bats
jasonkarns: Part of the refactor was removing the `HAS_X_SUPPORT` variables (where X is md5, sha2, etc). Previously, each potential checksum algorithm was tested for regardless whether it would ever need to be used. Now there's a general `has_checksum_support` function. This refactor simultaneously reduced in some areas but also increased in others the number of times an algorithm was tested for existence. In cases where the checksum algo is never needed, it's never checked for support. (e.g. https://github.com/rbenv/ruby-build/pull/881/files#diff-9fa50c58f7de44021ff445803deb773cL13) However, since there is no longer a variable the saves off the result of the support test, algo support is tested repeatedly in some areas.

To resolve this, I can update the `has_checksum_support` function to memoize the results which should remove the necessity for all those extra stubbings.

bin/ruby-build
jasonkarns: It's included in GNU coreutils: http://www.gnu.org/software/coreutils/manual/html_node/sha2-utilities.html#sha2-utilities

jasonkarns:
I can extract the refactoring into a separate PR that can be merged sooner. Then the 512-bit support can be merged or not depending on the fate of rubinius support

jasonkarns:
closing until rubinius' fate has been decided (#862)

PR #882, opened 2016-01-16
Update rubinius builds openssl from 1.0.1q to 1.0.2e

jasonkarns:
Updated via the [rubinius scraper](https://github.com/jasonkarns/ruby-build-update-defs/pull/2)!!!


mislav:
Yay!

Perhaps we should add a linting test to ruby-build that verifies that all openssl installs are referencing the same version (usually the latest)? That way we would ensure that some definitions are referencing outdated versions of openssl.

PR #883, opened 2016-01-17
Remvoed patch/tiny version from rubinius definitions

hsbt:
rubinius changed versioning scheme from 2.6 later.

 * http://rubinius.com/2016/01/07/rubinius-3-0-the-third-epoch/
 * https://rubinius-releases-rubinius-com.s3.amazonaws.com/index.txt

I removed redundant number from rubinius definitions.


mislav:
I agree that this follows Rubinius' version policy more closely, but is also a backwards-incompatible change. Users who were previously installing `rbx-X.Y.Z` definitions will find those to not be working anymore after ruby-build update.

Also, we're not sure if we'll keep supporting Rubinius #862 due to it being a moving target lately and not suitable for hardcoded definitions that we maintain right now.

Also, in my head I've entertained some sort of system in which `X.Y` and `X.Y.0` are recognizes as identical versions within ruby-build and possibly rbenv. That would make these kinds of discrepancies a non-issue.

I'll keep this open until a path forward is more clear.

hsbt:
> I'll keep this open until a path forward is more clear.

ok, I see.

NOTE: I worried to increase rubinius definitions. Because it released each couple of day. We need to have two hundreds definitions one year later. We will support another format for definition maybe.

jasonkarns:
It's also worth considering how rbenv-aliases would handle `rbx-3.0`

As it stands now, rbenv-aliases already creates a symlink from `rbx-3.0` to `rbx-3.0.0`. I'm not sure how rbenv-aliases would handle rbx-3.0 as a true version (probably just wouldn't do anything), but the extra .0 shouldn't be an issue for current rbenv-aliases users.

deepj:
I'm not sure if the following issue is related to this issue but it seems any rubinius 2.5.x (1) and later is not available publicly on old place -> https://s3.amazonaws.com/releases.rubini.us/ I've meet with that in PR#896 (see log).

1) except 2.5.8 can be found on new place https://rubinius-releases-rubinius-com.s3.amazonaws.com/rubinius-2.5.8.tar.bz2

mislav:
> except 2.5.8 can be found on new place

Good catch! Can you send a PR that updates the old URLs?

deepj:
Unfortunately, I've found Rubinius 2.5.8 only on their new main download site. The rest is available only on Github. See #897

PR #884, opened 2016-01-17
Rename CRuby development version.

hsbt:
CRuby development branch didn't use patch/tiny version number after 2.1 release. I removed redundant number.


mislav:
Can we just keep it the old way since it still works? Changing it will invalidate the previous git caches in `RBENV_ROOT/cache` directory, and we don't really gain anything by dropping the `.0` portion.

hsbt:
For example, ruby binary of 2.2.0-dev shows version with `ruby 2.2.5p236 (2016-01-18 revision 53567)` It doesn't contain "2.2.0". It confuse ruby-build user who didn't know CRuby version scheme. 

But this pull request is not strong opinion to me. If you reject this, I will close.

mislav:
Users never see the `ruby-2.2.0-dev` string anywhere. It's only used as a download & cache location for the git repository before checking out a specific branch or tag.

hsbt:
> Users never see the ruby-2.2.0-dev string anywhere.

fmm, It's doubt to me. Users invoke `rbenv install 2.2.0-dev` and expect to install 2.2.0 series.

mislav:
> fmm, It's doubt to me. Users invoke `rbenv install 2.2.0-dev` and expect to install 2.2.0 series.

Yes, but the `2.2.0-dev` string references the **name** of the ruby-build definition file. The first argument to `install_git` (`"ruby-2.2.0-dev"` in this case) is simply a name of the directory where the git repository is to be cloned, and users never really see this string.

So, if we change the string to `"ruby-2.2-dev"`, we don't help users in any way, since they don't see the change, but we do invalidate previous git caches that people might have on their system. Because of that, this change doesn't seem worth it to me.

hsbt:
ok, I close this for your concerns.

PR #886, opened 2016-01-18
rbenv ensures RBENV_ROOT is set for us

jasonkarns:
Handled by rbenv itself: https://github.com/rbenv/rbenv/blob/3997a394d975136f468be196941492d3d0ef5143/libexec/rbenv#L53-L58

Extracted from #880


mislav:
This breaks direct `rbenv-install` usage, but let's see if anyone/anything actually invokes it that way. :+1:

PR #887, opened 2016-01-18
Checksum refactor

jasonkarns:
Per the discussion from #881, this PR is the refactor of checksum handling.

- `has_checksum_support` function tests for support of the provided checksum algorithm, memoizing results
- `verify_checksum` now picks the algorithm based on checksum length. unexpected lengths no longer default to sha1(256bit), but rather error out. (They would error out eventually, anyway, when the resulting checksum doesn't match.)
- `HAS_X_SUPPORT` variables are no longer necessary
- algorithm support is now checked lazily; it is no longer checked if it wouldn't be used

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -265,14 +265,16 @@ verify_checksum() {
   [ -e "$filename" ] || return 0
 
   case "${#expected_checksum}" in
-  # If expected checksum is empty, return success
-  0) return 0 ;;
-  # MD5
+  0) return 0 ;; # empty checksum; return success
   32) checksum_command="compute_md5" ;;
-  # SHA2 256
   64) checksum_command="compute_sha2" ;;
-  # unknown checksum algorithm, return failure
-  *) return 1 ;;
+  *)
+    { echo
+      echo "unexpected checksum length: ${#expected_checksum} (${expected_checksum})"
+      echo "expected 0 (no checksum), 32 (MD5), or 64 (SHA2-256)"
+      echo
+    } >&4
+    return 1 ;;
   esac
 
   # If chosen provided checksum algorithm isn't supported, return success
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -246,6 +246,12 @@ compute_md5() {
   fi
 }
 
+declare -a HAS_CHECKSUM_SUPPORT
+has_checksum_support() {
+  local checksum_command="$1"
+  return ${HAS_CHECKSUM_SUPPORT[$checksum_command]:=$(echo test | "$checksum_command" >/dev/null; echo $?)}
+}
+
 verify_checksum() {
   local checksum_command
   local filename="$1"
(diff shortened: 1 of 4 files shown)


bin/ruby-build
```diff
@@ -244,24 +244,33 @@ compute_md5() {
...
-  [ -n "$expected_checksum" ] || return 0
+  case "${#expected_checksum}" in
+  # If expected checksum is empty, return success
+  0) return 0 ;;
+  # MD5
+  32) checksum_command="compute_md5" ;;
+  # SHA2 256
+  64) checksum_command="compute_sha2" ;;
+  # unknown checksum algorithm, return failure
+  *) return 1 ;;
```
mislav: Can you add a test for the case of checksum being of unsupported length? Should we also output a friendly error message about it if we're going to abort the build process?

bin/ruby-build
```diff
@@ -244,24 +244,33 @@ compute_md5() {
   fi
 }
 
+declare -a HAS_CHECKSUM_SUPPORT
+has_checksum_support() {
+  local checksum_command="$1"
+  return ${HAS_CHECKSUM_SUPPORT[$checksum_command]:=$(echo test | "$checksum_command" >/dev/null; echo $?)}
```
mislav: Bash 3 doesn't support associative arrays, so I can't imagine `HAS_CHECKSUM_SUPPORT[$checksum_command]` working. What you could do as an alternative to an array is create variable names dynamically:

```sh
has_checksum_var="HAS_CHECKSUM_SUPPORT_${checksum_command}"
if [ -n "${!has_checksum_var+defined}" ]; then
  return "${!has_checksum_var}"
else
  local supported=0
  echo test | "$checksum_command" >/dev/null || supported=$?
  eval "${has_checksum_var}=${supported}"
fi
```

bin/ruby-build
jasonkarns: Added error message (printed to STDERR), but noticed that the other error message within the verify function is printing to fd4. However, invocations of the `verify_checksum` are already redirecting STDERR to fd4. Shouldn't the functions simply use STDOUT and STDERR and let the caller worry about redirecting as necessary? It feels wrong that the verify function should need to know about fd4.

bin/ruby-build
jasonkarns: Used indirection as suggested, but opted for `printf -v` to assign the variable instead of `eval`. I'm allergic to eval.

bin/ruby-build
```diff
@@ -244,24 +244,39 @@ compute_md5() {
...
+  case "${#expected_checksum}" in
+  0) return 0 ;; # empty checksum; return success
+  32) checksum_command="compute_md5" ;;
+  64) checksum_command="compute_sha2" ;;
+  *)
+    { echo
+      echo "unexpected checksum length: ${#expected_checksum} (${expected_checksum})"
+      echo "expected 0 (no checksum), 32 (MD5), or 64 (SHA2-256)"
+      echo
+    } >&2
```
mislav: If other warning messages go to fd4 right now, let's cargo-cult that and output to fd4 manually.

bin/ruby-build
```diff
@@ -244,24 +244,39 @@ compute_md5() {
   fi
 }
 
+has_checksum_support() {
+  local checksum_command="$1"
+  local has_checksum_var="HAS_CHECKSUM_SUPPORT_${checksum_command}"
+
+  if [ -z "${!has_checksum_var+defined}" ]; then
+    printf -v "${has_checksum_var}" '%s' $(echo test | "$checksum_command" >/dev/null; echo $?)
```
mislav: TIL `printf -v` :astonished:

bin/ruby-build
jasonkarns: You're gonna make me cry. The caller is already redirecting to 2 to 4

bin/ruby-build
mislav: Since we're certain that the subexpression will always output a single integer, let's use it as a format string directly instead of `%s`. Also, double-quote it, even if it's not necessary, for consistency.

bin/ruby-build
mislav: Not all callers!

mislav: Subsequent PR to clear up the redirect mess welcome :wink:

bin/ruby-build
jasonkarns: done

bin/ruby-build
```diff
@@ -246,24 +246,39 @@ compute_md5() {
   fi
 }
 
+has_checksum_support() {
+  local checksum_command="$1"
+  local has_checksum_var="HAS_CHECKSUM_SUPPORT_${checksum_command}"
+
+  if [ -z "${!has_checksum_var+defined}" ]; then
+    printf -v "${has_checksum_var}" "$(echo test | "$checksum_command" >/dev/null; echo $?)"
```
mislav: Unnecessary curly braces :wink: :trollface:

jasonkarns: fixed

mislav:
Thanks for your patience!

PR #890, opened 2016-01-20
Leverage assertion test helpers

jasonkarns:
The current test suite uses the `assert` inconsistently. In some cases, `assert` is used. In others, the bash test built-in is used (or other 'naked' command).

The assert helper provides a marginal improvement in the error output when it fails, but this is minimal. Rather, I think the assert (and refute) test helpers improve the clarity of the test itself in signifiying the purpose of the commands. It helps clarify the arrange/act/assert sub-sections of each test.


mislav:
Looks good thanks!

PR #892, opened 2016-01-21
Add Rubinius 3.6--3.14

deepj:
I'm not sure if these commits are still welcome here after this debate #862.


mislav:
They're welcome, but due to daily releases of Rubinius, we can't keep up with hardcoded definition files anymore. I'll hold of merge until we figure out a better strategy.

PR #896, opened 2016-01-28
Upgrade to OpenSSL 1.0.1r and 1.0.2f

deepj:


mislav:
Would it work if we had the latest OpenSSL version everywhere? I presume 1.0.2f is the latest? Is MRI compatible with it?

deepj:
Yeah, that is a thing in question. I thought about it because support of OpenSSL 1.0.1 will be drop in the end of this year. Unfortunately, I'm not the right person who could test it properly if old Rubies are compatible with 1.0.2. But yes, I can change it to OpenSSL 1.0.2 without some kind of testing.

PR #897, opened 2016-01-29
New urls for old Rubinius versions and unavailable versions removed

deepj:
I found new urls for old Rubinius binaries on the Github. It seems the latest binary is for 2.2.2 (in ruby-build). See https://github.com/rubinius/rubinius/releases?after=v2.2.2. The rest is probably gone and nowhere publicly. So there is a question if Rubinius 2.2.1 and later should be removed.


mislav:
Yeah please add to this PR the removal of any rbx versions which aren't downloadable anymore.

deepj:
@mislav I hope it's ok now

mislav:
Thanks!

PR #899, opened 2016-01-30
Remove JRuby 9.0.0.0-dev since its snapshots are unavailable anymore

deepj:


hsbt:
Note: snapshot archive provided from http://ci.jruby.org/ now. but 9.0.0.0 is no available new url.

deepj:
@hsbt Thanks, I forgot to mention it as well.

mislav:
Thanks!

PR #901, opened 2016-02-01
add rbx-2.71828182

jasonkarns:
scraped via [ruby-build-update-defs](https://github.com/jasonkarns/ruby-build-update-defs) plugin


hsbt:
:+1: Thanks!

mislav:
Uhhh what version of rbx is that? Is it smaller or larger than rbx 2.8?

hsbt:
It's larger than rbx 2.8

jasonkarns:
According to this post
http://rubinius.com/2015/09/22/major-minor-maximize-delivering-features-minimize-trouble/
it was supposed to be the last 2.x release. I don't know if that ended up
being true or not.

PR #903, opened 2016-02-11
Restrict find command to subdirectories

### LINKED ISSUE #902: build fails when inaccessible directory exists in $PREFIX_DIR
I encountered the following error when attempting to install ruby on our RHEL 6 server using a prefix directory:

```
find: `/swadm/lost+found': Permission denied

BUILD FAILED (Red Hat Enterprise LInux Server release 6.7 (Santiago) using ruby-build 20160130)
```

The /swadm/lost+found directory is owned by root and we are installing ruby using a different user, which is why the find command run by the `fix_directory_permissions` method is generating the error.  I will be submitting a pull request that addresses this error shortly.

DebbieGillespie:
Fixes #902

The find command executed by `fix_directory_permissions` is too
inclusive.  It searches everything in the `$PREFIX_DIR`, which includes
directories that ruby-build does not care about, such as `lost+found`.
By restricting the find command to the `$PREFIX_DIR/bin` directory and
the `$PREFIX_DIR/share` directory, ruby-build will only change
permissions on the directories it touches, ignoring anything else that
might be in the `$PREFIX_DIR`, which could be anything.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -694,7 +694,7 @@ after_install_package() {
 
 fix_directory_permissions() {
   # Ensure installed directories are not world-writable to avoid Bundler warnings
-  find "$PREFIX_PATH" -type d \( -perm -020 -o -perm -002 \) -exec chmod go-w {} \;
+  find "$PREFIX_PATH/bin" "$PREFIX_PATH/share" -type d \( -perm -020 -o -perm -002 \) -exec chmod go-w {} \;
 }
 
 fix_rbx_gem_binstubs() {


mislav:
Ouch, thanks for catching this! ruby-build is mostly used to install to a user-owned location such as `~/.rbenv/versions/*`, and obviously in the case of installing to `/usr/local` as prefix we shouldn't be mucking things up. I think even after your change, doing a global `chmod` on `bin/` and `share/` directories might be too much. Maybe we should explore if we can get rid of this "fix permissions" hack altogether?

DebbieGillespie:
Perhaps a better approach would be to set `umask` to the desired permissions - that way the files that ruby-build creates would be set with appropriate permissions when they were created, thus removing the need to change the permission of the files after the fact.  I agree that doing a global `chmod` on `bin/` and `share/` is not desirable since those directories contain files that ruby-build did not create - however it is better than doing a chmod on the entire `$PREFIX_DIR`.

PR #904, opened 2016-02-13
Use `aria2c` the "ultra fast" downloader if available

yyuu:
[aria2](https://aria2.github.io/) is the "ultra fast" downloader and it supports parallel download with Range header, resuming failed download attempts, and so on. I received a feature request for this as yyuu/pyenv#534 from @luzfcb who is living in Brazil. I think this feature could be useful for all users who suffered by unstable internet connection.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -313,45 +313,27 @@ http() {
 }
 
 http_head_aria2c() {
-  options=""
-  [ -n "${IPV4}" ] && options="--disable-ipv6=true"
-  [ -n "${IPV6}" ] && options="--disable-ipv6=false"
-  aria2c --dry-run ${options} "$1" >&4 2>&1
+  aria2c --dry-run ${ARIA2_OPTS} "$1" >&4 2>&1
 }
 
 http_get_aria2c() {
-  options=""
-  [ -n "${IPV4}" ] && options="--disable-ipv6=true"
-  [ -n "${IPV6}" ] && options="--disable-ipv6=false"
-  aria2c -o "${2:--}" ${options} "$1"
+  aria2c -o "${2:--}" ${ARIA2_OPTS} "$1"
 }
 
 http_head_curl() {
-  options=""
-  [ -n "${IPV4}" ] && options="--ipv4"
-  [ -n "${IPV6}" ] && options="--ipv6"
-  curl -qsILf ${options} "$1" >&4 2>&1
+  curl -qsILf ${CURL_OPTS} "$1" >&4 2>&1
 }
 
 http_get_curl() {
-  options=""
-  [ -n "${IPV4}" ] && options="--ipv4"
-  [ -n "${IPV6}" ] && options="--ipv6"
-  curl -q -o "${2:--}" -sSLf ${options} "$1"
+  curl -q -o "${2:--}" -sSLf ${CURL_OPTS} "$1"
 }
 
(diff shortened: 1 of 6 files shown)


bin/ruby-build
```diff
@@ -300,42 +300,40 @@ http() {
...
     "http_${method}_wget" "$url" "$file"
   else
-    echo "error: please install \`curl\` or \`wget\` and try again" >&2
+    echo "error: please install \`aria2c\`, \`curl\` or \`wget\` and try again" >&2
     exit 1
   fi
 }
 
+http_head_aria2c() {
+  aria2c --dry-run ${ARIA2_OPTS} "$1" >&4 2>&1
```
mislav: Does `--dry-run` really perform a HEAD request?

bin/ruby-build
```diff
@@ -1219,6 +1217,10 @@ if [ -n "$RUBY_BUILD_SKIP_MIRROR" ] || ! has_checksum_support compute_sha2; then
   unset RUBY_BUILD_MIRROR_URL
 fi
 
+ARIA2_OPTS="${RUBY_BUILD_ARIA2_OPTS} ${IPV4+--disable-ipv6=true} ${IPV6+--disable-ipv6=false}"
+CURL_OPTS="${RUBY_BUILD_CURL_OPTS} ${IPV4+--ipv4} ${IPV6+--ipv6}"
+WGET_OPTS="${RUBY_BUILD_WGET_OPTS} ${IPV4+--inet4-only} ${IPV6+--inet6-only}"
```
mislav: Interesting additions. Not sure if we need to support passing in outside aria2/curl/wget configuration via environment variables. To be honest, I don't even know why we have the ipv4/v6 options and what they are for.

bin/ruby-build
yyuu: Yes. In aria2, it seems "dry run" means "not perfirming any GET requests". I confirmed it with tcpdump.

mislav:
Thanks for the addition! I'll have to look into this aria2 thing. I definitely want to support resumed downloads, but I thought wget already has a feature for that. We are probably not using it, though.

Is the feature of resumed downloads going to get used with the addition of aria2c support? Won't the build process be aborted on download failed and temporary files removed immediately?

bin/ruby-build
yyuu: At least for aria2, some user may want to specify custom options for splitting & parallelism for better download performance. Because it'd be depending on environment, I'd like to allow user to specify custom configurations via environment.

For ipv4/ipv6 options, I just moved them and I have never used it actually....

yyuu:
The aria2c command have retry feature (default: 5) and enabled by default. It also persists downloading content and the state in '*.aria2' file on the disk.

As you mentioned, the resuming only works if there is RUBY_BUILD_CACHE_PATH. aria2c will persists downloading content and its state in '*.aria2' file on disk.

yyuu:
Any updates on this?

hsbt:
@yyuu Sorry, our late response. I'm ok to support `aria2c` and merge it.

@mislav If you still have some concern, Please create another issue.

hsbt:
Thank you!

mislav:
I'm not opposed to the change, I just wish it didn't have to change literally all occurences of `curl` in tests. I'd rather that it added new tests that verify that `aria2c` downloads also work.

PR #934, opened 2016-04-14
GNU Make is always required for JRuby on FreeBSD

aycabta:


mislav:
Thanks for the suggestion! Does JRuby need a make process at all? I thought it downloads prebuilt one. How come gmake is always required? Which FreeBSD did you test on? Thanks

aycabta:
> Does JRuby need a make process at all? I thought it downloads prebuilt one. How come gmake is always required?

[JRuby install process](https://github.com/rbenv/ruby-build/blob/c97caf3705024386b70f2c9a7e29135c411bbcdf/bin/ruby-build#L652-L675) runs `gem install jruby-launcher`, the gem [has Makefile](https://github.com/jruby/jruby-launcher/blob/master/Makefile) that needs GNU Make syntax.

> Which FreeBSD did you test on?

FreeBSD 10.2-RELEASE. FYI, I know that [BSD Make after FreeBSD 10 is compatible
enough to build **C**Rubies](https://github.com/rbenv/ruby-build/commit/0e58079d660570765339fcd25895bae88aba1907), but it's not enough to build `jruby-launcher`.

danehammer:
I need this as well. The chef-ruby_build cookbook uses ruby-build (of course) and supports jruby, which it can't install on freebsd due to not using gmake.

aycabta:
Are there any other necessary things for this Pull Request?

I thought about test code, but [Travis CI doesn't support FreeBSD](https://docs.travis-ci.com/user/multi-os/).

hsbt:
It seems good. So we need to prepare FreeBSD box :)

aycabta:
@hsbt: http://ftp.freebsd.org/pub/FreeBSD/releases/VM-IMAGES/

Check this out.

PR #940, opened 2016-04-26
added definitions of 2.2.5 and 2.3.1

hsbt:
* https://www.ruby-lang.org/en/news/2016/04/26/ruby-2-2-5-released/
* https://www.ruby-lang.org/en/news/2016/04/26/ruby-2-3-1-released/


hsbt:
:octocat:

jeremy:
🤘

PR #944, opened 2016-05-03
Upgrade to OpenSSL 1.0.2h

deepj:


deepj:
Is it possible to rerun this on Travis? It's failed on:
```
curl: (35) Unknown SSL protocol error in connection to www.openssl.org:443
```

PR #949, opened 2016-05-11
Old crubies (1.8.5, 1.8.6, 1.8.7, 1.9.0, 1.9.1, 1.9.2, and 1.9.3)

jasonkarns:
Thanks to @hsbt, ruby-lang.org now has a manifest of published releases: http://cache.ruby-lang.org/pub/ruby/index.txt (Discussion leading to said manifest from: https://github.com/jasonkarns/ruby-build-update-defs/issues/3)

This has enabled a scraper to be written for cruby for use by the [ruby-build-update-defs plugin](https://github.com/jasonkarns/ruby-build-update-defs).

There are quite a few published releases of ruby available on ruby-lang.org (per the manifest) that don't have corresponding build definitions here in ruby-build. I used the scraper (tweaked per minor version) to generate the build definitions that are in this PR. The build-defs are chunked per minor version (1.8.5, 1.8.6, 1.8.7, 1.9.0, 1.9.1, 1.9.2, and 1.9.3)

If all these build definitions are added, then adding new build definitions to ruby-build becomes a simple matter of: `rbenv update-version-defs --cruby --destination share/ruby-build` (assuming the plugin is installed and the command is run from root of this repo). Indeed, the entire scrape/branch/pr process has been scripted for node-build using this scraper: https://github.com/nodenv/node-build/blob/master/script/submit-definitions

Some notes:

- 1.8.5: There weren't any existing build definitions for any 1.8.5 ruby so 1.8.6-p383 was used as the template:
    - require_gcc
    - warn_eol auto_tcltk standard
    - rubygems-1.3.7
- 1.8.6: (as the other 1.8.6 rubies)
    - require_gcc
    - warn_eol auto_tcltk standard
    - rubygems-1.3.7
- 1.8.7: (as the other 1.8.7 rubies)
    - require_gcc
    - warn_eol auto_tcltk standard
    - rubygems-1.6.2
- 1.9.0: There weren't any existing build definitions for any 1.9.0 ruby so 1.9.1-p378 was used as the template:
    - require_gcc
    - yaml-0.1.6
    - warn_eol standard
    - rubygems-1.3.7
- 1.9.1: (as the other 1.9.1 rubies)
    - require_gcc
    - yaml-0.1.6
    - warn_eol standard
    - rubygems-1.3.7
- 1.9.2: (as the other 1.9.2 rubies)
    - require_gcc
    - yaml-0.1.6
    - warn_eol standard
    - rubygems-1.8.23
- 1.9.3-p105 (modeled after 1.9.3-p125)
    - [ -n "$CC" ] || export CC=cc
    - openssl-1.0.2h
    - yaml-0.1.6
    - warn_eol standard
    - rubygems-1.8.23
- 1.9.3-p426 (modeled after 1.9.3-p429 and all the other 1.9.3 rubies)
    - openssl-1.0.2h
    - yaml-0.1.6
    - warn_eol standard

The 1.8.5 rubies and 1.9.3-p105 are the only definitions that I'm not confident in.

### DIFF
--- a/share/ruby-build/1.8.5-p113
+++ b/share/ruby-build/1.8.5-p113
@@ -1,0 +1,3 @@
+require_gcc
+install_package "ruby-1.8.5-p113" "https://cache.ruby-lang.org/pub/ruby/1.8/ruby-1.8.5-p113.tar.bz2#216600f9ad07648c501766a25069009c5c543010821da2ad916dd2ca808efd01" warn_eol auto_tcltk standard
+install_package "rubygems-1.3.7" "https://rubygems.org/rubygems/rubygems-1.3.7.tgz#388b90ae6273f655507b10c8ba6bee9ea72e7d49c3e610025531cb8c3ba67c9d" ruby
--- a/share/ruby-build/1.8.5-p114
+++ b/share/ruby-build/1.8.5-p114
@@ -1,0 +1,3 @@
+require_gcc
+install_package "ruby-1.8.5-p114" "https://cache.ruby-lang.org/pub/ruby/1.8/ruby-1.8.5-p114.tar.bz2#c503ae8eb47db72f78fb7a79fe1874ffef40a7094f7e803bacbf994a924244d9" warn_eol auto_tcltk standard
+install_package "rubygems-1.3.7" "https://rubygems.org/rubygems/rubygems-1.3.7.tgz#388b90ae6273f655507b10c8ba6bee9ea72e7d49c3e610025531cb8c3ba67c9d" ruby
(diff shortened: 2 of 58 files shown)


hsbt:
@jasonkarns Awesome!

I'm going to investigate some packages like '1.9.0-0' to 1.9.0-5', 1.8.6 and 1.8.7. I think it's better to add `-p0` now. Can you wait few days?

jasonkarns:
Certainly, no rush. I have no need for these definitions. The main benefit to having them all in ruby-build is that the scraper won't attempt to re-generate them each time. The way the scraper works is that it will generate a definition for each release that is in the release listing that _doesn't_ already exist in ruby-build. As none of these already exist, they will continue to be regenerated each time the scraper is run.

Having these added will prevent the scraper from regenerating them each time.

hsbt:
I confirmed it. It's official release name.

 * https://www.ruby-lang.org/en/news/2007/03/12/ruby-1-8-6-released/
 * https://www.ruby-lang.org/en/news/2008/05/31/ruby-1-8-7-has-been-released/
 * https://www.ruby-lang.org/ja/news/2008/03/01/ruby-1-9-0-1-snapshot-released/ (available only Japanese language)

`1.9.0-1` is same as preview release in the past.

mislav:
Thanks for discovering these old versions, but I don't think adding the EOLed versions was strictly necessary. Users weren't asking for this before, and now we'll have to support them. Who knows if these 1.8.x-1.9.0 versions install correctly on modern versions of CentOS, Debian, Ubuntu, OS X, and so on.

hsbt:
@mislav I agreed to remove EOL'ed versions of Ruby includes 1.9.3 and 2.0.0. We should discuss another issue. It seems big issue for ruby-build.

PR #951, opened 2016-05-13
Add script wrapper for running the ruby scraper

jasonkarns:
This script depends on the ruby-build-update-defs plugin. Assuming it is installed (doesn't matter how), then this script can be used to scrape new rubies and write the definitions to `share/ruby-build`. A second script is forthcoming that automates the branch+pr process for newly-scraped rubies.

### DIFF
--- a/script/scrape-rubies
+++ b/script/scrape-rubies
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+# Usage: script/scrape-rubies
+#
+# All parameters are passed transparently to `rbenv update-version-defs`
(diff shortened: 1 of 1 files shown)


mislav:
This looks alright, but is very simple and little useful by itself yet. I'll hold off merging until we solidify what the process for "scraping" ruby definitions should be like.

hsbt:
I added https://github.com/rbenv/ruby-build/blob/master/script/update-cruby. I'm happy to sync feature from released ruby packages to it.

I'm closing this.

PR #953, opened 2016-05-16
Mirror checksums

jasonkarns:
As hinted from #878, this adds mult-digest support to the mirroring script.

In its present form, it adds MD5 and SHA-512 support to the existing SHA-256 behavior. The md5 support is isolated to its own commit, so it can be dropped from this PR if desired. (It was added for completeness sake.)

I'm not really sure how to test this, so it hasn't really been tested :/ Other than the fact it's mostly copied from https://github.com/rbenv/ruby-build/blob/8dc4568ea9f02db55ddf26de9b48d01e0bdb1d05/bin/ruby-build#L267-L278

### DIFF
--- a/script/mirror
+++ b/script/mirror
@@ -15,8 +15,8 @@ test_mirrored() {
 }
 
 compute_checksum() {
-  local digest="$1"
-  local output="$(openssl dgst -"$digest")"
+  local digest_type="$1"
+  local output="$(openssl dgst -"$digest_type")"
   echo "${output##* }" | tr '[:upper:]' '[:lower:]'
 }
 
@@ -25,16 +25,16 @@ download_package() {
 }
 
 download_and_verify() {
-  local checksum digest
+  local checksum digest_type
   local url="$1"
   local file="$2"
   local expected="$3"
   download_package "$url" "$file"
 
   case "${#expected}" in
-  32) digest="md5" ;;
-  64) digest="sha256" ;;
-  128) digest="sha512" ;;
+  32) digest_type="md5" ;;
+  64) digest_type="sha256" ;;
+  128) digest_type="sha512" ;;
   *)
     { echo "Error: unexpected checksum length: ${#expected} (${expected})"
       echo "expected 32 (MD5), 64 (SHA-256), or 128 (SHA-512)"
@@ -42,7 +42,7 @@ download_and_verify() {
     return 1 ;;
   esac
 
-  checksum="$(compute_checksum "$digest" < "$file")"
+  checksum="$(compute_checksum "$digest_type" < "$file")"
   if [ "$checksum" != "$expected" ]; then
     echo "Error: $url doesn't match its checksum $expected" >&2
     return 1
--- a/script/mirror
+++ b/script/mirror
@@ -16,7 +16,7 @@ test_mirrored() {
 
 compute_sha256() {
   local output="$(openssl dgst -sha256)"
-  echo "${output##* }" | tr '[A-Z]' '[a-z]'
+  echo "${output##* }" | tr '[:upper:]' '[:lower:]'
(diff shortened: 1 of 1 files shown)


script/mirror
```diff
@@ -14,22 +14,35 @@ test_mirrored() {
   curl -qsSfIL "$RUBY_BUILD_MIRROR_URL/$1" >/dev/null 2>&1
 }
 
-compute_sha2() {
-  local output="$(openssl dgst -sha256)"
-  echo "${output##* }" | tr '[A-Z]' '[a-z]'
+compute_checksum() {
+  local digest="$1"
```
mislav: I feel like this doesn't need a local variable, but if it does, it's better that it's called something like `type` rather than ambiguous "digest".

jasonkarns:
Kept the local var; I'm not a fan of using positional args without local names because it's not abundantly clear what the variable _is_. I went with `digest_type` instead since `type` is a keyword. And renamed the other use of `$digest` as well. I wasn't particularly happy with `$digest` before, anyway, since most people would assume `digest` to be the actual checksum itself.

Any comment on allowing the mirror to handle MD5?

mislav:
I'm fine with handling MD5 since it doesn't cost us much.

jasonkarns:
Just realized I don't think I commented since I made the requested variable name change. I think this is ready to merge unless there is a way to test the mirror script.

reedloden:
If MD5 is not needed, best not to include it at all. It's a known-broken algorithm that has numerous problems (and known collisions). Best to stick to SHA-2 and other hash algorithms that are not yet broken.

jasonkarns:
@reedloden I'd agree for things that are producing md5. However, this script ought to be able to handle as wide a range of checksums as possible, IMO. That way it can consume the widest possible set of build definition files. Especially because this script operates on git history by taking in ranges of commits as input. Therefore, even though it exists _in_ the repository (and is therefore kept in lockstep with the definition files it consumes), it operates on ranges of history so it rather exists _outside_ of repo history as well. I think it makes sense to be able to operate on old commits that do contain md5 checksums.

mislav:
@jasonkarns Can we find a way to share the checksum functions between these executables? Now that the checksum approach is more sophisticated, I'm not looking forward to maintaining different implementations in different executables.

I'm going to be away for 2 weeks from now, so I won't be around to review/accept PRs. To recognize your substantial contributions to the rbenv family of projects, I've invited you to the org, so you have write access. However, please be wary of merging larger changes (including your own) to either projects without approval of other people in the org.

jasonkarns:
@mislav bin/ruby-build already has the `lib` function which is used to share functions between `bin/ruby-build` and `bin/rbenv-install`. We could move the checksum function(s) into `lib()` so that the mirror script can use them.

However, this brings up another thought I've had for nodenv.

I've long thought that there should be some organizational distinction (probably just different files) between "private" functions that ruby-build uses internally and "public" functions that are intended to be used by build definitions. I was going to suggest making `lib/internal.sh` and `lib/build.sh` files to contain those functions (respectively) that would then be sourced by `bin/ruby-build`.

The current `lib()` function is crazy clever (nearly a quine!), but I'm not sure if it has any value over just having a separate file that's sourced. Thoughts? Simple solution is just to move the checksum functions into lib and then decide about potential extraction later.

jasonkarns:
Actually, I recognize that keeping ruby-build functional as a single file makes the `install` step trivial. So I'm still for moving the checksum functions into `lib()` soon and having a bigger conversation around lib extraction much later.

mislav:
@jasonkarns I'm sorry I never got around to merging this, but I waited to see if ruby-build will need sha-512 support for anything specific, and since it hasn't, I will close this until actually needed.

PR #965, opened 2016-06-02
`aria2c` doesn't support writing remote content to stdout

yyuu:
I've implemented `aria2c` support in #904. However, I finally realized that `aria2c` doesn't support writing remote content to stdout (https://github.com/aria2/aria2/issues/190).

I confirmed that basically this must be not a problem for ruby-build since it doesn't rely on `http get` without filename ([the only usage of http get](https://github.com/rbenv/ruby-build/blob/3304f96296c9912eaee27347e87e93a3e8fc5a29/bin/ruby-build#L412) is specifying its destination filename). It'd be better to be fixed though.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -317,7 +317,12 @@ http_head_aria2c() {
 }
 
 http_get_aria2c() {
-  aria2c -o "${2:--}" ${ARIA2_OPTS} "$1"
+  local out="${2:-$(mktemp "out.XXXXXX")}"
+  if aria2c --allow-overwrite=true -o "${out}" ${ARIA2_OPTS} "$1" >&4; then
+    [ -n "$2" ] || cat "${out}"
+  else
+    false
+  fi
 }
 
 http_head_curl() {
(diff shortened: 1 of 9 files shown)


mislav:
Thanks for catching this. Like my comment in https://github.com/rbenv/ruby-build/pull/904#issuecomment-221599563, I don't like that every tweak to aria2c support needs to touch all tests. Is there a way we can circumvent this? Maybe go back to stubbing `curl` instead, and `aria2c` as uninstalled on the system, and then have just 1 tests that stubs aria2c as supported and tests its invocation?

yyuu:
It sounds nicer than current test impl..... I'll try to do it later.

yyuu:
I've prepared test helper to remove `aria2c` from `$PATH`. Now we can use `curl` stub during test as same as before. I've prepared dedicated test case for `aria2c` and it'd be sufficient to check if `aria2c` is working.

@mislav could you review the changes again?

mislav:
Thanks for doing all the work and being patient!

PR #979, opened 2016-08-13
Remove suite-wide setup

jasonkarns:
Bats doesn't support both suite-wide setup (in helper) *and* file-wide
setup. Which means the existance of any file-level setup() function
overwrites any setup() function from test_helper. This can be confusing,
and (IMO) easier to simply avoid the overwriting and remove any
_implied_ suite-wide setup function from test_helper.

This turns out to not be so bad for the recently added setup function,
because the only test files that actually need aria2c removed from PATH
are those that actually invoke curl. These can be found because they are
the only test files that stub curl; half of which already had
file-specific setup() functions. So the only ones which need a
file-local setup() function added were: checksum.bats and mirror.bats

Along these lines, rbenv.bats and hooks.bats were removing aria2c from
PATH but don't actually need to. (curl isn't stubbed in these tests so
the existance of aria2c wouldn't affect the tests)

Lastly, fixed a tab/spaces whitespace mixup that was introduced by:
750c086d11d8cc6ccb041f5c19278c994a50854f

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1222,9 +1222,9 @@ if [ -z "$MAKE" ]; then
       export MAKE="gmake"
     else
       if [ "$(uname -r | sed 's/[^[:digit:]].*//')" -lt 10 ]; then
-	export MAKE="gmake"
+        export MAKE="gmake"
       else
-	export MAKE="make"
+        export MAKE="make"
       fi
     fi
   else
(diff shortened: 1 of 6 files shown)


jasonkarns:
I've confirmed that the test suite still passes locally with these changes even if aria2c is installed. However, I wouldn't mind @yyuu validating this PR.

mislav:
Approved 👍

yyuu:
LGTM 😀

Issue #980, opened 2016-08-17
Use brew bump-formula-pr


MikeMcQuaid:
https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/bump-formula-pr.rb may give you a better experience than using your own version (as we'll notice quicker if it breaks).

hsbt:
@MikeMcQuaid Thank you for your information. I tried to create new version with `bump-formula-pr`. It's very useful for me. I replace our original script to it.

PR #996, opened 2016-09-16
Use bump formula pr

hsbt:
To use official "pull-request" tool of homebrew

ref. https://github.com/rbenv/ruby-build/issues/980

### DIFF
--- a/script/brew-publish
+++ b/script/brew-publish
@@ -8,11 +8,6 @@ set -e
 brew_name="${1?}"
 version="${2?}"
 
-if ! type -p hub >/dev/null; then
-  "ERROR: You have to install hub to proceed." >&2
-  exit 1
-fi
-
 if [ -n "$3" ]; then
   gh_project="github.com/${3}"
 else
--- a/script/brew-publish
+++ b/script/brew-publish
@@ -29,24 +29,4 @@ if [ -z "$checksum" ]; then
   exit 1
 fi
 
-pushd "$(brew --prefix)"
-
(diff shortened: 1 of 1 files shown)


mislav:
Thanks! I've cut the script down a bit further in 073762c12869b323646f94ed6a6bfee6ccb289c2

hsbt:
Thank you too!

jasonkarns:
Just FYI, I extracted this script out to https://github.com/jasonkarns/brew-publish/blob/master/bin/brew-publish a while back (for nodenv-*)

It's been updated to use bump-formula-pr as well. Any desire to leverage this instead of having the script embedded directly? Would also be useful for rbenv and any other rbenv-* plugins that are included in homebrew.

mislav:
> Any desire to leverage this instead of having the script embedded directly?

How would we leverage this vs. embedding? I like the embedding approach since it means that we don't have to pull in a 3rd party repo to gain access to a simple shell script.

jasonkarns:
Seeing as there isn't a "standard" bash package manager, I would imagine git submodule or subtree are the only options. Since it would only be for maintainers of ruby-build, and not end users, I could see git-submodule being palatable, if barely. Personally, I think the better solution would be git-subtree. The file would be "embedded" in the repo without needing any extra commands to fetch. But at the same time, it could be trivially kept up to date with the "upstream" version of the script.

It could be used on all the rbenv repos in a consistent fashion, thus making all plugin releases consistent.

PR #998, opened 2016-09-18
Require LLVM 3.6 when building Rubinius 3.30+

eugeneius:
From 3.30 onwards, Rubinius requires LLVM 3.6 or newer: https://github.com/rubinius/rubinius/commit/2dd51a1c285d1adae7224d4c966f664a11ed447e

Previously discussed here: https://github.com/rbenv/ruby-build/pull/765#issuecomment-105488542


mislav:
Thanks!

PR #1002, opened 2016-09-25
print when using homebrew packages

chulkilee:
When a user uses homebrew packages (e.g. openssl), then to upgrade them the user must update homebrew packages before rebuilding using ruby-build.


chulkilee:
We may just update the wiki page, but I think it would be nice if ruby-build print messages when using them. I'm not sure it's "right" way to do print messages (e.g. STDOUT or not, etc.).

hsbt:
Thank you. It seems useful information.

Issue #1009, opened 2016-10-17
Ruby 2.3.1 build issues on freebsd11.0-release


cod3:
When I upgraded the freebsd to 11 from 10, ruby ​​is no longer able to build.

generating x86_64-freebsd11.0-fake.rb
x86_64-freebsd11.0-fake.rb updated
./miniruby -I./lib -I. -I.ext/common  ./tool/runruby.rb --extout=.ext  -- --disable-gems -r./x86_64-freebsd11.0-fake ./tool/rbinstall.rb --make="make" --dest-dir=""  --extout=".ext"  --mflags=" "  --make-flags=" .MAKE.LEVEL.ENV=MAKELEVEL"  --data-mode=0644  --prog-mode=0755  --installed-list .installed.list  --mantype="doc" --install=all --rdoc-output=".ext/rdoc"
/tmp/ruby-build.20161017164820.14002/ruby-2.3.1/lib/rubygems/core_ext/kernel_require.rb:55:in `require': /tmp/ruby-build.20161017164820.14002/ruby-2.3.1/.ext/x86_64-freebsd11.0/openssl.so: Undefined symbol "SSLv2_method" - /tmp/ruby-build.20161017164820.14002/ruby-2.3.1/.ext/x86_64-freebsd11.0/openssl.so (LoadError)

$uname -a
FreeBSD localhost 11.0-RELEASE-p1 FreeBSD 11.0-RELEASE-p1 #0 r306420: Thu Sep 29 01:43:23 UTC 2016     root@releng2.nyi.freebsd.org:/usr/obj/usr/src/sys/GENERIC  amd64

hsbt:
It caused `SSLv2_method` was removed at FreeBSD 11.

I'm going to investigate this.

hsbt:
addressed https://bugs.ruby-lang.org/issues/12849

rhenium:
I have no idea why it occurs as it must not happen (see https://bugs.ruby-lang.org/issues/4556).

Could you show me mkmf.log and extconf.h generated by the compile time? They should be in `/tmp/ruby-build.20161017164820.14002/ruby-2.3.1/ext/openssl`.

cod3:
I have uploaded this file
/tmp/ruby-build.20161017164820.14002/ruby-2.3.1/ext/openssl/mkmf.log

https://raw.githubusercontent.com/cod3/test/master/mkmf.log

rhenium:
[...]

PR #1015, opened 2016-11-04
support FreeBSD 11-RELEASE

yuichiro-naito:
ruby-build fails when I use on FreeBSD 11-RELEASE.

```
$ rbenv install 2.3.1
Downloading ruby-2.3.1.tar.bz2...
-> https://cache.ruby-lang.org/pub/ruby/2.3/ruby-2.3.1.tar.bz2
Installing ruby-2.3.1...

BUILD FAILED (FreeBSD 11.0-RELEASE-p2 using ruby-build 20160913-13-g8ef0c34)

Inspect or clean up the working tree at /tmp/ruby-build.20161104153337.41276
Results logged to /tmp/ruby-build.20161104153337.41276.log

Last 10 log lines:
                              rake 10.4.2
installing rdoc:              /home/yuichiro/.rbenv/versions/2.3.1/share/ri/2.3.0/system
installing capi-docs:         /home/yuichiro/.rbenv/versions/2.3.1/share/doc/ruby
The Ruby readline extension was not compiled.
Configure options used:
  --prefix=/home/yuichiro-n/.rbenv/versions/2.3.1
  CFLAGS= -O3 -Wno-error=shorten-64-to-32 
  LDFLAGS=-L/home/yuichiro-n/.rbenv/versions/2.3.1/lib 
  CPPFLAGS=-I/home/yuichiro-n/.rbenv/versions/2.3.1/include 
```

The reason of this error,
FreeBSD 11-RELEASE has changed not to export readline shared library.
Application should use readline library in Ports Collection.

c.f.
  https://www.freebsd.org/releases/11.0R/relnotes.html

In FreeBSD Ports Collection,
ruby interpreter compiled with libedit by default.

To follow this configuration,
this patch prefers to link libedit first, readline next.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -931,13 +931,14 @@ use_freebsd_pkg() {
       # check if 11-R or later
       release="$(uname -r)"
       if [ "${release%%.*}" -ge 11 ]; then
-	  if pkg info -e libedit > /dev/null; then
+	  # prefers readline to compile most of ruby versions
+	  if pkg info -e readline > /dev/null; then
+	      # use readline from Ports Collection
+	      package_option ruby configure --with-readline-dir="/usr/local"
+	  elif pkg info -e libedit > /dev/null; then
 	      # use libedit from Ports Collection
 	      package_option ruby configure --enable-libedit
 	      package_option ruby configure --with-libedit-dir="/usr/local"
-	  elif pkg info -e readline > /dev/null; then
-	      # use readline from Ports Collection
-	      package_option ruby configure --with-readline-dir="/usr/local"
 	  fi
       fi
   fi
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -920,15 +920,21 @@ use_homebrew_yaml() {
 }
 
 use_freebsd_pkg() {
-  # check if FreeBSD 11-R or later
+  # check if FreeBSD
   if [ "FreeBSD" = "$(uname -s)" ]; then
+      # use openssl if installed from Ports Collection
+      if [ -f /usr/local/include/openssl/ssl.h ]; then
+	  package_option ruby configure --with-openssl-dir="/usr/local"
+      fi
+
+      # check if 11-R or later
       if [ "$(uname -r | sed 's/[^[:digit:]].*//')" -ge 11 ]; then
 	  if $(pkg info -e libedit); then
-	      # use if libedit is installed
+	      # use libedit from Ports Collection
 	      package_option ruby configure --enable-libedit
 	      package_option ruby configure --with-libedit-dir="/usr/local"
 	  elif $(pkg info -e readline); then
-	      # use if readline is installed
+	      # use readline from Ports Collection
 	      package_option ruby configure --with-readline-dir="/usr/local"
 	  fi
       fi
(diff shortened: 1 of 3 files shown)


hsbt:
Thank you. We need to add `--with-openssl-dir` option as same reason.

And We're investigate this issue on ruby core side.

yuichiro-naito:
Do you like to use openssl library from Ports Collection ?
Have a look at my second patch.

hsbt:
Seems good.

Sorry, my information is not enough. I pointed out https://bugs.ruby-lang.org/issues/12849 and https://github.com/rbenv/ruby-build/issues/1009

hsbt:
Our CI is fragile status. I will merge when it fixed. Maybe It caused Travis boxes.

yuichiro-naito:
I'm sorry that my response is late.
I'm not familiar with Travis ci.
My last commit fixed all test cases have been passed.
Will you check the test code?

bin/ruby-build
```diff
@@ -553,6 +553,7 @@ build_package_standard() {
   local PACKAGE_CFLAGS="${package_var_name}_CFLAGS"
 
   [ "$package_var_name" = "RUBY" ] && use_homebrew_readline || true
+  [ "$package_var_name" = "RUBY" ] && use_freebsd_pkg || true
```
mislav: This is repeating a condition from the line before. Combining the two would be a good approach:

```sh
if [ "$package_var_name" = "RUBY" ]; then
  use_homebrew_readline || use_freebsd_pkg || true
fi
```

bin/ruby-build
```diff
@@ -918,6 +919,28 @@ use_homebrew_yaml() {
...
+use_freebsd_pkg() {
+  # check if FreeBSD
+  if [ "FreeBSD" = "$(uname -s)" ]; then
+      # use openssl if installed from Ports Collection
+      if [ -f /usr/local/include/openssl/ssl.h ]; then
+	  package_option ruby configure --with-openssl-dir="/usr/local"
+      fi
+
+      # check if 11-R or later
+      if [ "$(uname -r | sed 's/[^[:digit:]].*//')" -ge 11 ]; then
```
mislav: I feel that this approach of extracting the release number is brittle. Instead, if the output of `uname -r` is always going to be a number followed by a dash, I suggest the following approach:

```sh
release="$(uname -r)"
release="${release%%-*}"
[ "$release" -ge 11 ]
```

bin/ruby-build
```diff
@@ -918,6 +919,28 @@ use_homebrew_yaml() {
...
+	  if $(pkg info -e libedit); then
```
mislav: If you're only interested in the exit status of `pkg info -e` but want to discard its output, you should simply redirect to null:

```sh
if pkg info -e libedit >/dev/null; then
```

bin/ruby-build
yuichiro-naito: ```uname -r``` returns like ```11.0-RELEASE-p2``` string.
This includes major and minor release number.
For comparing with ```-ge```, numbers must be integer.
I will fix to cut off dot and following characters.

bin/ruby-build
yuichiro-naito: Yes, I'm interested in exit status only.
```pkg info -e``` shows no messages to STDOUT.
So I didn't take care of the output.
Directing ```/dev/null``` emphasizes this point,
and more informational way.
I will fix it.

yuichiro-naito:
I found older ruby (before 2.1) needs readline library.
So I fixed to use readline first, libedit second.
Latter ruby (not before 2.1) works with readline, too.
This fix has no problem for most of ruby versions.
Will you check the latest code?

mislav:
Thank you! 🙇

PR #1031, opened 2016-12-05
Homebrew upgrade for HEAD package.

mxl:
From `brew upgrade --help`:

>     If --fetch-HEAD is passed, fetch the upstream repository to detect if
>     the HEAD installation of the formula is outdated. Otherwise, the
>     repository's HEAD will be checked for updates when a new stable or devel
>     version has been released.


mislav:
Thanks!

PR #1038, opened 2016-12-21
Run test on freebsd

yuichiro-naito:
I fixed to run all tests have been passed on FreeBSD 11-R.
I think it is useful for future development of ruby-build.
Will you merge this PR ?

### DIFF
--- a/script/brew-publish
+++ b/script/brew-publish
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # Usage: script/brew-publish <formula-name> <repo> <version>
 set -euo pipefail
 
--- a/test/build.bats
+++ b/test/build.bats
@@ -62,6 +62,7 @@ assert_build_log() {
   cached_tarball "yaml-0.1.6"
   cached_tarball "ruby-2.0.0"
 
+  stub uname '-s : echo Linux'
   stub brew false
   stub_make_install
   stub_make_install
@@ -69,6 +70,7 @@ assert_build_log() {
   install_fixture definitions/needs-yaml
   assert_success
 
+  unstub uname
   unstub make
 
   assert_build_log <<OUT
@@ -85,6 +87,7 @@ OUT
   cached_tarball "yaml-0.1.6"
   cached_tarball "ruby-2.0.0"
 
+  stub uname '-s : echo Linux'
   stub brew false
   stub_make_install
   stub_make_install
@@ -93,6 +96,7 @@ OUT
   TMPDIR="$TMP" install_fixture --patch definitions/needs-yaml <<<""
   assert_success
 
+  unstub uname
   unstub make
   unstub patch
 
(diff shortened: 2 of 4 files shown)


script/brew-publish
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
```
mislav: Where is bash located on freebsd?

test/build.bats
```diff
@@ -355,6 +367,7 @@ apply -p1 -i /my/patch.diff
 exec ./configure "\$@"
 CONF
 
+  stub uname '-s : echo Linux'
```
mislav: What happens if `uname` is not stubbed?

test/test_helper.bash
```diff
@@ -3,7 +3,7 @@ export TMP="$BATS_TEST_DIRNAME/tmp"
 if [ "$FIXTURE_ROOT" != "$BATS_TEST_DIRNAME/fixtures" ]; then
   export FIXTURE_ROOT="$BATS_TEST_DIRNAME/fixtures"
   export INSTALL_ROOT="$TMP/install"
-  PATH=/usr/bin:/usr/sbin:/bin/:/sbin
+  PATH="/usr/bin:/usr/sbin:/bin/:/sbin:$PATH"
```
mislav: Which important directory was missing from this list? I specifically didn't want to include the whole `$PATH`.

script/brew-publish
yuichiro-naito: bash is installed in `/usr/local/bin/bash`.
It's an optional package and not included in FreeBSD base system.

test/build.bats
yuichiro-naito: When `uname -s` returns `FreeBSD`,
ruby-build adds configure option '--with-readline-dir=/usr/local'.
It is a result of my previous Pull Request.

Some error message are shown like below.

```
✗ yaml is installed for ruby
   (from function `assert_equal' in file test/test_helper.bash, line 139,
    from function `assert_output' in file test/test_helper.bash, line 148,
    from function `assert_build_log' in file test/build.bats, line 58,
    in test file test/build.bats, line 76)
     `assert_build_log <<OUT' failed
   expected: yaml-0.1.6: --prefix=${TMP}/install
   make -j 2
   make install
   ruby-2.0.0: --prefix=${TMP}/install
   make -j 2
   make install
   actual:   yaml-0.1.6: --prefix=${TMP}/install
   make -j 2
   make install
   ruby-2.0.0: --prefix=${TMP}/install --with-readline-dir=/usr/local
   make -j 2
   make install
```

test/test_helper.bash
yuichiro-naito: `/usr/local/bin` is the most important.

I'm guessing some people might compile and install bash
to other directory like `/opt/pkg/bin`.
FreeBSD ports tree supports $PREFIX definition.

c.f. https://www.freebsd.org/doc/en/books/porters-handbook/porting-prefix.html

I don't use this option,
and if you don't want to include whole $PATH,
just adding `/usr/local/bin` satisfies me.

mislav: You you change the list to:

```
/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin
```

yuichiro-naito:
Thanks for the comment.
I fixed the code.

mislav:
Thank you!

PR #1046, opened 2017-01-19
gcc is installed on FreeBSD

yuichiro-naito:
I fixed a problem about gcc is installed on FreeBSD-11.

The problem is simple.
When I install gcc on my FreeBSD,
ruby-build fails to compile for all ruby versions.

ruby-build adds "-Wno-error=shorten-64-to-32" option to CFLAGS,
when cc is clang. (clang is the default compiler of FreeBSD-11)
But if the configure script of ruby detects gcc,
ruby is compiled with "gcc -Wno-error=shorten-64-to-32",
it will fail because gcc doesn't supports "-Wno-error=shorten-64-to-32" option.

My patch checks if gcc is not installed,
and then adds "-Wno-error=shorten-64-to-32" option to CFLAGS.

Will you merge this Pull Request?

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1236,7 +1236,7 @@ if [ -n "$noexec" ]; then
   exit 1
 fi
 
-# Apply following work around, if gcc is not available.
+# Apply following work around, if gcc is not installed.
 if [ -z "$(locate_gcc)" ]; then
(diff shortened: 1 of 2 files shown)


hsbt:
I've been confirmed this pull request fixed build failure with gcc on FreeBSD 11.

PR #1050, opened 2017-01-27
Fix to build Rubinius on FreeBSD

yuichiro-naito:
ruby-build fails to build Rubinius on FreeBSD-11.
The reasons of this problem are following 3 issues.

1. llvm package

FreeBSD llvm packages are named like following.

  llvm37
  llvm38
  llvm39
  llvm-devel

For each package,
clang and llvm-config are renamed to be installed as following.

  clang37     llvm-config37
  clang38     llvm-config38
  clang39     llvm-config39
  clang-devel llvm-config-devel

I fixed that ruby-build checks which package is installed
and then make configure options by the result.

2. readline package

readline package is installed in '/usr/local',
I fixed to add configure option '--with-opt-dir=/usr/local'.

3. cc & c++ wrapper

It's complicated to explain.

Rubinius build tool makes symbolic links for cc and c++ compiler.
If the configure options specifies to use clang37,
cc is linked to '/usr/local/bin/clang37' and
c++ is linked to '/usr/local/bin/clang++37'.

Unfortunately '/usr/local/bin/clang37' is a wrapper script on FreeBSD.
The wrapper script makes the clang compiler path
like "/usr/local/llvm37/bin/$0".

As the result of these,
cc is executed as "/usr/local/llvm37/bin/cc".
No compiler is found in the path.

Same things happens in FreeBSD ports.
I fixed to use the wrapper (written in C) as same as FreeBSD ports.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -601,6 +601,45 @@ build_package_ree_installer() {
   } >&4 2>&1
 }
 
+post_configure_hook_rbx() {
+    if [ "FreeBSD" = "$(uname -s)" ]; then
+	cat > build-tool.c <<_EOS_
+/* LLVM wrapper script invokes the actual binary based on the argv[0]
+ * which rubinius messes up, by calling clang39 as cc, and thus breaking
+ * the LLVM wrapper. This hack workarounds this limitation, by invoking
+ * LLVM wrapper script with expected argv[0]
+ */
+
+#include <string.h>
+#include <unistd.h>
+
+#define PATH_CC "${CLANG_BIN}"
+#define PATH_CXX "${CLANGCPP_BIN}"
+
+int
+main(int argc, char** argv)
+{
+	char* basename = strrchr(argv[0] , '/');
+
+	if(NULL == basename)
+		basename = argv[0];
+	else
+		basename++;
+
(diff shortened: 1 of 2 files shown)


yuichiro-naito:
Will you merge this pull request ?

hsbt:
I couldn't review this. I'm not familiar with rubinius build environment.

mislav:
@yuichiro-naito Thank you for the patch. It sounds like you did a lot of investigation into these build problems.

However, we don't include such patches in the source of ruby-build itself. We do ocassionally perform some workarounds for common problems, but this workaround seems very complex and hard for us to maintain over time, since we don't understand most what's going on in here.

Did you raise these issues with the Rubinius team, and do they have some easier workarounds to suggest, or are they working on fixing FreeBSD build problems in upcoming releases?

yuichiro-naito:
Thanks for the comment.

I know this pull request is complicated and
fixing many part of build process for particular reason.

When I think what causes this problem,
Rubinius build process and FreeBSD compiler environment are the factors
not ruby-build.

Fixing ruby-build was the easiest way for me.
I understand it is hard to maintain for the future.
I'm going to suggest FreeBSD or Rubinius project about this problem.

PR #1059, opened 2017-02-23
`--quiet` option isn't sensible sha256sum(1) of GNU coreutils

yyuu:
The option is available only if verifying digest, not available when computing digest.

>       --quiet
>              don't print OK for each successfully verified file


yyuu:
Is there any issue merging this? Actually ruby-build's usage of `sha256sum` is just corrupted and won't work with `sha256sum` of GNU coreutils / BusyBox.

hsbt:
@yyuu Sorry, my late response. I'm ok to merge this. Can you resolve a conflict?

yyuu:
Rebased!

hsbt:
Thanks!

PR #1067, opened 2017-03-22
Add definition for Ruby 2.4.1

koic:
Refer: http://blade.nagaokaut.ac.jp/cgi-bin/scat.rb/ruby/ruby-core/80279


hsbt:
Thanks always.

PR #1072, opened 2017-03-30
Add definition for Ruby 2.3.4

koic:
Refer: https://www.ruby-lang.org/en/news/2017/03/30/ruby-2-3-4-released


hsbt:
Thanks always

PR #1073, opened 2017-03-31
ruby-build 20170322

ivanovaleksey:
Hi,
it seems to me that `RUBY_BUILD_VERSION` should be bumped according to recent release, shouldn't it?


hsbt:
It's my missing updates. Thanks.

Issue #1079, opened 2017-04-24
Rubinius 2.n LLVM installation instructions on macOS are out of date


chrisseaton:
When I try to `rbenv install rbx-2.11` on macOS, I'm told to `brew install llvm@3.5`, but this doesn't work any more `No available formula with the name "llvm@3.5"`.

Why do I want to install such an old version of Rubinius? Because the JIT was removed from newer versions and the new one isn't in place yet.

What do I do instead of `brew install llvm@3.5`.

cc @brixen

hsbt:
I understood your concern. But instruction of llvm-3.5.x installation is not ruby-build responsibility scope, I think.

chrisseaton:
But this code and the error message is part of ruby-build isn't it?

https://github.com/rbenv/ruby-build/blob/master/bin/ruby-build#L862-L893

```bash
require_llvm() {
  local llvm_version="$1"
  if is_mac -ge 1010; then
    if [[ "$RUBY_CONFIGURE_OPTS" != *--llvm-* ]]; then
      case "$llvm_version" in
      3.2 )
        package_option ruby configure --prebuilt-name="llvm-3.2-x86_64-apple-darwin13.tar.bz2"
        ;;
      3.[56] )
        local llvm_config="$(locate_llvm "$llvm_version")"
        if [ -n "$llvm_config" ]; then
          package_option ruby configure --llvm-config="$llvm_config"
        else
          local homebrew_package="llvm@$llvm_version"
          { echo
            colorize 1 "ERROR"
            echo ": Rubinius will not be able to compile using Apple's LLVM-based "
            echo "build tools on OS X. You will need to install LLVM $llvm_version first."
            echo
            colorize 1 "TO FIX THE PROBLEM"
            echo ": Install Homebrew's llvm package with this"
            echo -n "command: "
            colorize 4 "brew install $homebrew_package"
            echo
          } >&3
          return 1
        fi
        ;;
      esac
    fi
  fi
}
```

It's not code that is in the Rubinius tarball.

`ruby-build` is the only place that this message can be fixed.

hsbt:
ok, I removed/modified this code.

Your original concern caused by Homebrew. So, [llvm@3.5](https://github.com/Homebrew/homebrew-core/blob/master/Formula/llvm@3.5.rb) is not available now, Only 4.0-3.7.

hsbt:
Duplicates of #862

PR #1082, opened 2017-05-02
End to support old rbx versions.

hsbt:
Related with https://github.com/rbenv/ruby-build/issues/1079

homebrew stopped to distribute llvm-3.5 and 3.6. Unfortunately, We will not support old rbx versions that are required llvm-3.5 and 3.6. Also, we will not support version required llvm-3.2.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -879,20 +879,6 @@ require_llvm() {
   fi
 }
 
-locate_llvm() {
-  local llvm_version="$1"
-  local package llvm_config
-  shopt -s nullglob
-  for package in `brew list 2>/dev/null | grep "^llvm"`; do
-    llvm_config="$(echo "$(brew --prefix "$package")/bin/llvm-config"*)"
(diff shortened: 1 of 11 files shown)


jasonkarns:
Meta-question: Since ruby-build supports pulling build definitions from other /share locations (other plugins, other locations specified by `RUBY_BUILD_DEFINITIONS`), are we concerned that removing the `locate_llvm` (or any other functions within ruby-build) will break build definitions that exist outside of ruby-build itself? I've personally been thinking about a process for deprecating functions:

1. extract deprecated functions to a separate file (optional)
2. add deprecation warning to the function
3. create a "backcompat" type plugin to house deprecated functions (and perhaps even contains removed build definitions for old rubies?)
4. move function from ruby-build into the "backcompat" plugin

This type of process would provide a safe transition period for any external build definitions relying on functions that ruby-build itself considers dead. And at the very least would provide an archive of functions that _used_ to be available in ruby-build. Thoughts?

hsbt:
This branch was stale status. I will consider with current develop environment on macOS.

PR #1095, opened 2017-06-08
Ignore LibreSSL bundled with macOS 10.13

osyoyu:
macOS 10.13 High Sierra has replaced OpenSSL 0.9.8 with LibreSSL 2.2.7. Even though, the header files (`ssl.h` etc.) aren't included, so we can't use this to build Ruby.

This pull request changes `has_broken_mac_openssl()` to ignore LibreSSL bundled with macOS 10.13.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -959,7 +959,10 @@ use_homebrew_readline() {
 
 has_broken_mac_openssl() {
   is_mac &&
-  [[ "$(/usr/bin/openssl version 2>/dev/null || true)" = "OpenSSL 0.9.8"?* ]] &&
+  [[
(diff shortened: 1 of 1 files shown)


hsbt:
Did you install Xcode-beta or invoke `xcode-select --install` ?

has_broken_mac_openssl don't evaluate that macOS contains `ssl.h`. It's problem for forked openssl in macOS.

hsbt:
Sorry, development files like `ssl.h` are not provided after El Capitan(10.11). This pull request seems ok to me. 

Note: We have two problem about openssl on macOS. First one is broken openssl forked Apple, Inc. Second one is that header files is not provided after macOS 10.11.

PR #1097, opened 2017-06-08
SI-740: reverse curl and wget preference

chaiken-verticloud:
curl does not work on the Altiscale CentOS platform (through at least Prometheus 6.7.9), probably due to an old version of the openssl RPM.  Putting wget in front of curl allows the code to work on Altiscale CentOS without any degradation in functionality.


chaiken-verticloud:
We can manage this issue locally at Altiscale, assuming that the owners of ruby-build don't want to make this change.

PR #1100, opened 2017-06-10
Fix installing error on macOS 10.13

Watson1978:
Now, `rbenv install` fails on macOS 10.13.

macOS 10.13 beta has bundled LibreSSL 2.2.7 as OpenSSL library.
```
$ sw_vers -productVersion
10.13
$ /usr/bin/openssl version
LibreSSL 2.2.7
```

However, seems that Ruby has supported LibreSSL 2.3+, not support 2.2.7
 (https://github.com/ruby/ruby/blob/aab0d67a1ff5190ff7a951e40cee742210302aed/ext/openssl/History.md#supported-platforms)

So, if system bundled older LibreSSL, this patch will use OpenSSL installed by homebrew to fix installing error.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -957,9 +957,18 @@ use_homebrew_readline() {
   fi
 }
 
+is_broken_openssl() {
+  openssl_version="$(/usr/bin/openssl version 2>/dev/null || true)"
+  if [[ "$openssl_version" = "OpenSSL 0.9.8"?* ]] || [[ "$openssl_version" = "LibreSSL 2.2"? ]]; then
+    return 1
+  else
+    return 0
+  fi
+}
+
 has_broken_mac_openssl() {
   is_mac &&
-  [[ "$(/usr/bin/openssl version 2>/dev/null || true)" = "OpenSSL 0.9.8"?* ]] &&
+  is_broken_openssl &&
   [[ "$RUBY_CONFIGURE_OPTS" != *--with-openssl-dir=* ]] &&
   ! use_homebrew_openssl
 }


Watson1978:
Sorry, seems this duplicates #1095 :(

hsbt:
Can you try to install Ruby 2.2.x? If it is caused by OpenSSL support version, We can install old versions of Ruby.

hsbt:
I update #1095 . Maybe, High Sierra is not provided header files like `ssl.h`. We can not install all CRuby.

Watson1978:
Failed to install 2.2.7 without this PR.

```
$ rbenv install 2.2.7
Downloading ruby-2.2.7.tar.bz2...
-> https://cache.ruby-lang.org/pub/ruby/2.2/ruby-2.2.7.tar.bz2
Installing ruby-2.2.7...
ruby-build: use readline from homebrew

BUILD FAILED (OS X 10.13 using ruby-build 20170523-7-g2de7a29)

Inspect or clean up the working tree at /var/folders/fs/gltv6wcd3m70hcgqcp2njt_80000gn/T/ruby-build.20170610130621.75637
Results logged to /var/folders/fs/gltv6wcd3m70hcgqcp2njt_80000gn/T/ruby-build.20170610130621.75637.log

Last 10 log lines:
installing rdoc:              /Users/watson/.rbenv/versions/2.2.7/share/ri/2.2.0/system
installing capi-docs:         /Users/watson/.rbenv/versions/2.2.7/share/doc/ruby
The Ruby openssl extension was not compiled.
Configure options used:
  --prefix=/Users/watson/.rbenv/versions/2.2.7
  --with-readline-dir=/usr/local/opt/readline
  CC=clang
  LDFLAGS=-L/Users/watson/.rbenv/versions/2.2.7/lib
  CPPFLAGS=-I/Users/watson/.rbenv/versions/2.2.7/include
```

hsbt:
Thanks. I'm ok to ignore LibreSSL 2.2.7 same as old versions of macOS. I'm going to merge #1095. Thank you for your investigation.

Watson1978:
Ok, thanks :)

PR #1104, opened 2017-06-12
Improve check for openssl on macOS

mislav:
- Only call `openssl version` once;

- Make the check work with future version of LibreSSL after 2.2.7. This assumes that Apple might keep updating LibreSSL shipping with macOS, but that they will never provide header files.

Followup to https://github.com/rbenv/ruby-build/pull/1095

/cc @hsbt @osyoyu

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -958,11 +958,9 @@ use_homebrew_readline() {
 }
 
 has_broken_mac_openssl() {
-  is_mac &&
-  [[
-    "$(/usr/bin/openssl version 2>/dev/null || true)" = "OpenSSL 0.9.8"?* ||
-    "$(/usr/bin/openssl version 2>/dev/null || true)" = "LibreSSL 2.2.7"
-  ]] &&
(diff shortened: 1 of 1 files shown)


mislav:
@jeremy That's a good point. If there wasn't ever a `/usr/bin/openssl` that passed this check (since it was "0.9.8" since forever), nor will it ever (now that we blacklisted LibreSSL), I guess we can drop the `openssl version` check entirely?

hsbt:
my understanding is following summary about OpenSSL on macOS.

 * < 10.11(El Capitan): macOS has openssl-0.9.8, but it's forked binary for Apple, so It's broken.
 * >= 10.11(El Capitan): macOS was not provide OpenSSL header.
 * = 10.13(High Sierra): macOS changed LibreSSL from OpenSSL, but it was not provided header files same as El Capitan.(I do not confirm it yet)

So, OpenSSL/LibreSSL of macOS is completely broken. I agreed to use jeremy's condition for openssl veriication.

PR #1106, opened 2017-06-13
Clarify install guidance

jeremy:
* Move Homebrew to the top, common case for new devs
* Recommend install method according to how rbenv is installed
* Spell out exactly how to get new build definitions


rlue:
Is installing ruby-build via homebrew really such a common use case that it deserves to be placed first? 

[The Homebrew formula for `rbenv` already includes ruby-build](https://github.com/rbenv/rbenv#homebrew-on-mac-os-x); I think it would be wiser to suggest promote that avenue over this one.

Thoughts?

mislav:
@jeremy Thanks a lot for this suggestion and sorry for the late response. I've just merged an alternate PR that offered a more significant overhaul to the README. One part of this is that Homebrew is always the preferred installation method offered to macOS users.

PR #1107, opened 2017-06-16
Add definition for JRuby 9.1.11.0 and 9.1.12.0

koic:
These JRubies has been released.

- http://jruby.org/2017/06/14/jruby-9-1-11-0.html
- http://jruby.org/2017/06/15/jruby-9-1-12-0.html

Thanks.


zoras:
@koic @hsbt thoughts on auto updating homebrew ruby-build formula on addition of new definitions?
https://github.com/Homebrew/homebrew-core/blob/master/Formula/ruby-build.rb

hsbt:
I have no plan to auto-update mechanism yet.

PR #1114, opened 2017-07-10
Make `README` more succinct

rlue:
I ran into a big headache with `ruby-build` because I didn't have my build environment set up the way I should have. The `README` made mention of this, but I missed it because the `README` is long and I am lazy.

I believe this version conveys practically all the same information as the original in far fewer words and a clearer visual hierarchy. ruby-build is very straightforward to use; its documentation should reflect that.

If these changes are to the maintainers' liking, I'd even ask your approval to take it a step further and move most of the "Advanced Usage" section (along with the "Installation > Upgrading" section) to the wiki, replacing it with links to the relevant articles. But I won't get ahead of myself.

What do you folks think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ definitions.
 
 The build process may be configured through the following environment variables:
 
+| Variable                 | Function                                                                                         |
 | ------------------------ | ------------------------------------------------------------------------------------------------ |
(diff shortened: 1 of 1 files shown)


wilson:
+300, this is a fantastic diff.
As soon as a maintainer notices, I hope it gets merged.

mislav:
Thank you! Much appreciated.

PR #1120, opened 2017-08-02
Quote JRuby check on FreeBSD to avoid "unary operator" syntax issue.

wilson:
Currently, if you pass an argument like '-v' or '-k' to ruby-build on a FreeBSD system, you get a shell syntax warning about '=' being used as a unary operator.

This is happening because of some missing quotes, and no longer complains once they are added.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1281,7 +1281,7 @@ fi
 
 if [ -z "$MAKE" ]; then
   if [ "FreeBSD" = "$(uname -s)" ]; then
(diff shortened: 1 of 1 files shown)


hsbt:
I can reproduce this.

```
$ rbenv install -v jruby-9.1.13.0
/path/to/.rbenv/plugins/ruby-build/bin/ruby-build: line 1284: [: =: unary operator expected
```

I'm ok to merge this. Thanks.

PR #1136, opened 2017-10-09
Relax java version regexp

### LINKED ISSUE #1135: Incompatible with Java 9
Java 9's version is `9` - just `9`, not `1.9` - but `ruby-build` checks for `1.[789]`, so it thinks Java 9 is incompatible.

https://github.com/rbenv/ruby-build/blob/73988db2ae1cb0339f7e8fde959baf243f19f05e/bin/ruby-build#L691

juanibiapina:
This allows the java version to be specified as "9", instead of "1.9".

Closes #1135

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -773,7 +773,7 @@ fix_rbx_irb() {
 
 require_java7() {
   local version="$(java -version 2>&1 | grep '\(java\|openjdk\) version' | head -1)"
-  if [[ $version != *1.[789]* ]]; then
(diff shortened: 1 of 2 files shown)


juanibiapina:
Actually jruby-9.1.13.0 doesn't install with java 9 because jruby-openssl has the exact same problem with the version. Maybe this shouldn't be merged unless the whole thing works.

hsbt:
@juanibiapina Thanks for your patch. If you have improvement for this patch, Please submit again. I hope to care JRuby and rbenv users.

PR #1147, opened 2017-11-07
Build Ruby against the right OpenSSL on macOS if pkg-config is present

krbullock:


hsbt:
What's resolve by this pull-request? Please give your use-case to me.

krbullock:
I have a case where pkg-config may be installed thru Fink at /sw/bin/pkg-config, and may return paths for the OpenSSL installed there. This was causing recent Ruby versions to fail to compile via ruby-build for me. Setting `PKG_CONFIG_PATH` to ensure that ruby-build builds Ruby against the intended OpenSSL library seemed the best way to fix it, and ought to work 🤞 no matter what the environment.

hsbt:
I see. I will merge this.

krbullock:
Thank you!

PR #1162, opened 2017-12-15
Fix rbenv install on Solaris / Illumos

klimkjar:
rbenv install crashes on Solaris with an empty log file. Adding support
for the proper Solaris getconf call in num_cpu_cores fixed it. Tested
and working under OmniOS CE r151024.


hsbt:
Thanks!

PR #1174, opened 2018-01-22
Add snapshot and stable-snapshot

znz:
They are useful on environments without autoconf etc.
Based on share/ruby-build/2.5.0.
When autoconf is old (for example CentOS 6), 2.5.0-dev can not install.


hsbt:
This is useful. But I'm not sure that we can sustainably maintain it. 

----

In Japanese.

便利とは思うのですが、snapshot が指すバージョンがその年で異なることと、依存関係などが変わった時に継続的にメンテナンスするのが大変そうという感触を持ってます。

hsbt:
It's hard to maintain with OpenSSL version. I will close this.

PR #1189, opened 2018-03-29
remove spaces before newline

toy:


jeremy:
@toy Thanks for the PR. Could you run the test suite as well? It needs an update to account for the whitespace change.

toy:
@jeremy I've pushed the change so tests are green

PR #1198, opened 2018-04-26
Allow overriding HTTP client type based on environment variable of `RUBY_BUILD_HTTP_CLIENT`

yyuu:
This allows users to choose appropriate HTTP client for ruby-build based on environment variable of `RUBY_BUILD_HTTP_CLIENT`. Since rbenv is based on shim script architecture, checking some command's availability based on the visibility in `$PATH` (e.g. `type aria2c`) is not actually appropriate way to do so.

The issue was originally reported as pyenv/pyenv#1126 that there was some problem in `aria2c` package of FreeBSD and `python-build` was always preferred to use it since it was on the top of the list. I think similar issue can happen with `ruby-build` too. This provides users to prepare some workaround for situations like that.


mislav:
Thank you! 🙇

PR #1226, opened 2018-08-21
More portable rename after tarball extraction

eregon:
This should solve https://github.com/rbenv/ruby-build/commit/d155b41e#commitcomment-30215955.

Renaming is needed for TruffleRuby, so just reverting the commit is not enough.
Specifically, https://github.com/oracle/truffleruby/releases/download/vm-1.0.0-rc5/truffleruby-1.0.0-rc5-linux-amd64.tar.gz contains a toplevel directory `truffleruby-1.0.0-rc5-linux-amd64` which is not the same as `$package_name` which is `truffleruby-1.0.0-rc5`

So in this PR, I rename the extracted directory if it is not the same as `$package_name`
* make_package() assumes the archive was extracted to $package_name.
* fetch_tarball() does not check the name of the toplevel directory in
  the archive, so rename after if it's not exactly "$package_name".

@mislav @xaque208 Please review this PR.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -381,10 +381,7 @@ fetch_tarball() {
     download_tarball "$package_url" "$package_filename" "$checksum"
   fi
 
-  mkdir "$package_name"
-
-  { if tar $tar_args "$package_filename" -C "$package_name" --strip-components=1; then
-  ls >&2
+  { if tar $tar_args "$package_filename"; then
       if [ -z "$KEEP_BUILD_PATH" ]; then
         rm -f "$package_filename"
       else
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -350,6 +350,7 @@ fetch_tarball() {
   local package_url="$2"
   local mirror_url
   local checksum
+  local extracted_dir
 
   if [ "$package_url" != "${package_url/\#}" ]; then
     checksum="${package_url#*#}"
@@ -382,6 +383,11 @@ fetch_tarball() {
   fi
 
   { if tar $tar_args "$package_filename"; then
+      if [ ! -d "$package_name" ]; then
+        extracted_dir=$(find_extracted_directory)
+        mv "$extracted_dir" "$package_name"
+      fi
+
       if [ -z "$KEEP_BUILD_PATH" ]; then
         rm -f "$package_filename"
       else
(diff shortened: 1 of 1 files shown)


xaque208:
Thanks again for putting this up @eregon.  I'll check out the sha and can test have a test.

bin/ruby-build
```diff
@@ -381,10 +381,12 @@ fetch_tarball() {
     download_tarball "$package_url" "$package_filename" "$checksum"
   fi
 
-  mkdir "$package_name"
+  { if tar $tar_args "$package_filename"; then
+      if [ ! -d "$package_name" ]; then
+        extracted_dir=$(find_extracted_directory)
```
mislav: Needs `"` around `$(...)`

bin/ruby-build
```diff
@@ -394,6 +396,16 @@ fetch_tarball() {
...
 }
 
+find_extracted_directory() {
+  for f in *; do
+    if [ -d "$f" ]; then
+      echo "$f"
+      return
+    fi
+  done
+  echo "Extracted directory not found"
```
mislav: This failure message should go to `>&2`, and the function should `return 1` to make it exit with a non-zero status

bin/ruby-build
eregon: AFAIK this is correct when using assignment: https://unix.stackexchange.com/questions/118433/quoting-within-command-substitution-in-bash
I also used ShellCheck to make sure I didn't introduce any new warning.

But I'll change, it works fine with the `"` too and it's more consistent.
The above `tarball_filename` also doesn't use `"`, I'll add them.

bin/ruby-build
eregon: Won't the return value be ignored since this is called like
`extracted_dir="$(find_extracted_directory)"` ?
Is there a good pattern for dealing with this (checking return value and returning a string)?

eregon: Actually returning 1 works as expected and fails the build :smiley: 
```
...
Extracted directory not found

BUILD FAILED (Fedora 26 using ruby-build 20180618-15-g4ff9ea3)
```

eregon: Interestingly this doesn't error out if the caller is
```bash
local extracted_dir=$(find_extracted_directory)
```
but it works if declaration and assignment of `extracted_dir` are separate.
Bash is fun.

eregon:
@mislav I fixed the PR to address your review, could you re-review?

PR #1224, opened 2018-08-15
Upgrade to OpenSSL 1.0.2p and 1.1.0i

deepj:
Is it possible to release a new version of ruby-build after merging this, please? :)

PR #1228, opened 2018-08-26
Fix missing space since openssl-1.1.0i update

sodabrew:
Corrects error from PR #1224


sodabrew:
Apologies for the lazyweb PR, but if you'd kindly Squash Merge with the title `Fix missing space since openssl-1.1.0i update` that would be awesome.

deepj:
@sodabrew thanks for the fix. I haven't noticed it. My mistake 😢  My apologies! @hsbt @sodabrew

PR #1231, opened 2018-09-04
Strip binaries after build

fauno:
Adds the `strip_binaries` build flag to a package definition.  it runs
`strip` to the `ruby` binary and libraries, removing unneeded
information and making the files smaller and faster.

It was tested with 2.5.1 but it should work with any version.

<https://en.wikipedia.org/wiki/Strip_(Unix)>

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1136,8 +1136,12 @@ build_package_strip_binaries() {
   local major_version=${DEFINITION_PATH##*/}
   major_version=${major_version%.*}.0
 
-  strip --strip-all "${PREFIX_PATH}/bin/ruby"
-  find "${PREFIX_PATH}/lib/ruby/${major_version}" -type f -name "*.so" | xargs strip --strip-unneeded
+  if command -v strip >/dev/null 2>&1 ; then
+    strip --strip-all "${PREFIX_PATH}/bin/ruby"
+    find "${PREFIX_PATH}/lib/ruby/${major_version}" -type f -name "*.so" | xargs strip --strip-unneeded
+  else
+    colorize 1 "You need to install \`strip\` from binutils to reduce library size"
+  fi
 }
 
 rake() {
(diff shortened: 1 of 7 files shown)


hsbt:
Does it work with macOS?

fauno:
> Does it work with macOS?

it should if you have `strip` installed, it's part of `binutils` so i
guess you should have it installed alongside `gcc` for building ruby.

mislav:
Thanks for the suggestion. I'm curious how @hsbt feels about this, because he's the more invested maintainer of this project than I am right now, but I don't think it's a good fit for the ruby-build project.

Our tool _doesn't_ aim out of the box to make the smallest possible and optimized installation of Ruby; we just aim to make the _most default_, basic installation of Ruby (an equivalent to `./configure && make install`) and let other people control aspects of it by environment variables, command-line arguments, and `after_install` callbacks in the `install.bash` hook [within rbenv plugins](https://github.com/rbenv/rbenv/wiki/Authoring-plugins).

Most people have plenty of disk space and don't need to strip down their installations of Ruby to bare minimum. Those who do, e.g. when installing Ruby in a container and wanting to keep the container size small, will want to run `strip` and other post-processing steps after the installation has finished, but that should ideally be done not by us, but by their own scripts and build definitions. The idea is, is someone cares so much about install size, they should roll out their own optimized installation.

You might think: “but it costs us nothing to simply run the `strip` tool if it's available on the system, and have everything normally if `strip` is unavailable!” And that's somewhat true; the implementation that you've contributed looks fine to me and it gracefully falls back when `strip` is unavailable. But, you have to think about the fact that the project will have to maintain every bit of its functionality for years to come, and that we want to be careful about adding features that we don't think are absolutely necessary and that they would be better served by a custom script or a pluginin.

a-vrma:
`strip` is installed with the command line tools (`/Library/Developer/CommandLineTools`) in macOS.
Any user with a C compiler will also have the `strip` tool.

a-vrma:
There is no --strip-all option in the macOS `strip` though.

a-vrma:
Also in my single test on macOS it didn't change more than a kb.

jlduran:
@fauno any reason why 2.6 is not stripped?

mislav:
I stand by my above rationale and closing this as #wontfix. Thank you for suggesting the approach! For people who find this thread, it should be easy to create your own build definitions for Rubies you want stripped and re-use the `build_package_strip_binaries` code from this PR.

fauno:
none, i'm still using 2.5 and didn't pay attention to other versions
yet, i'll push changes to 2.6 also :)

-- 
 D

PR #1245, opened 2018-10-22
pass arguments of --list to definitions()

glaszig:
this is just a quick hack but it pains me there's no built-in filter method for `rbenv-install --list` even though internally it exists. this patch just exposes it.

```
$ rbenv install -l 2.6
Available versions:
  2.2.6
  2.6.0-dev
  2.6.0-preview1
  2.6.0-preview2
  rbx-2.2.6
  rbx-2.6
```

### DIFF
--- a/bin/rbenv-install
+++ b/bin/rbenv-install
@@ -78,7 +78,7 @@ for option in "${OPTIONS[@]}"; do
     ;;
   "l" | "list" )
     echo "Available versions:"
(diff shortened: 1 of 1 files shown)


mislav:
Thanks for the contribution, but it's easy enough to just pipe the output to `grep` or another search tool yourself! That's the philosophy behind command-line tools, anyway. That way you can also control the pattern better, e.g. you can use `^2.6` for a better match.

PR #1253, opened 2018-11-11
Update EOL Ruby message for macOS > 10.9 Mavericks

searls:
_Please go easy on me, I'm a first time contributor to the xenv suite of tools_ 🙇

I was trying to install Ruby 1.8.7 on macOS Mojave today and found that this message was out of date. `brew/taps` complains it no longer exists and the `apple-gcc42` package complains it hasn't worked on the last 4 macOS releases. Installing this gcc package did make `rbenv instal 1.8.7-p375` work for me (though I can't speak for whether or not it will link SSL correctly, basic Ruby scripts seem to work)

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -838,9 +838,9 @@ require_gcc() {
 
         colorize 1 "TO FIX THE PROBLEM"
         if type brew &>/dev/null; then
-          echo ": Install Homebrew's apple-gcc42 package with this"
+          echo ": Install Homebrew's GCC package with this"
           echo -n "command: "
-          colorize 4 "brew tap homebrew/dupes ; brew install apple-gcc42"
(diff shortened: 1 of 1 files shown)


searls:
OK, can confirm that OpenSSL doesn't link correctly by default. 

For what it's worth, this full command does work:

```
CC=gcc-4.9 \
CFLAGS="-O2 -fno-tree-dce -fno-optimize-sibling-calls -I$(brew --prefix openssl)/include" \
LDFLAGS="-L$(brew --prefix openssl)/lib" \
rbenv install 1.8.7-p375
```

Is that something that we should just paste into the message or is there a build script change to be made to address those flags?

searls:
@hsbt OK, wiki updated!

PR #1257, opened 2018-11-24
Upgrade to OpenSSL 1.0.2q and 1.1.0j

deepj:


hsbt:
@deepj Thanks always.

PR #1259, opened 2018-11-25
Fix grep syntax for OpenBSD (wasn't picking up version before)

ckuttruff:


ckuttruff:
Just confirming that this of course works as expected on linux:

```
root@orghub:~# cat /etc/issue
Ubuntu 18.04.1 LTS \n \l

root@orghub:~# java -version 2>&1 | grep -E '(java|openjdk) version' | head -1
openjdk version "1.8.0_181"
root@orghub:~# 
```

Just a different between GNU grep / OpenBSD grep with the default syntax supported (sans extended regex)

hsbt:
The original reporter is left from GitHub.

PR #1275, opened 2019-01-15
Add TruffleRuby 1.0.0-rc11

deepj:


eregon:
> I'm thinking to add a small script under `script/` to automate this, so there is no copy-paste involved.

See https://github.com/rbenv/ruby-build/pull/1276

PR #1283, opened 2019-02-04
Use shasum -a 256 instead of sha256sum in update-truffleruby

deepj:
@eregon ^^

### DIFF
--- a/script/update-truffleruby
+++ b/script/update-truffleruby
@@ -14,7 +14,7 @@ add_platform() {
   platform="$1"
   basename="truffleruby-${version}-${platform}.tar.gz"
   url="https://github.com/oracle/truffleruby/releases/download/vm-${version}/${basename}"
-  sha256=$(curl -L "$url" | sha256sum | cut -d ' ' -f 1)
+  sha256=$(curl -L "$url" | shasum -a 256 | cut -d ' ' -f 1)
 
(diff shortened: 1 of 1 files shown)


eregon:
Mmh, but this doesn't work on Linux:

```
$ shasum -a 256  README.md
zsh: shasum: command not found...
```
FWIW, the `ruby-versions` update script also expects `sha256sum` to be available:
https://github.com/postmodern/ruby-versions/blob/5f172745e188222d2f66da61141a825fb4eb42e3/update.sh#L73

eregon:
Maybe using openssl is more portable, as shown in https://apple.stackexchange.com/a/230919/61281
That works on Linux (but the output is swapped, the digest is last).

mislav:
See https://github.com/rbenv/ruby-build/blob/7205ae477579a279df2a041606398c79033d0fc8/bin/ruby-build#L221-L232

deepj:
@mislav thanks for the tip! I’ll change it

jasonkarns:
Starting to think the sha function from bin/ruby-build should be moved up into the lib() wrapper so that it can be leveraged by the CI tests, mirroring scripts, and scraper scripts. It's already duplicated once I believe, by the mirror script.

mislav:
@jasonkarns That's not a bad idea! 👍

PR #1288, opened 2019-03-13
Add support for ruby 2.5.4 and 2.6.2

em77:
Ruby [2.5.4](https://www.ruby-lang.org/en/news/2019/03/13/ruby-2-5-4-released/) and [2.6.2](https://www.ruby-lang.org/en/news/2019/03/13/ruby-2-6-2-released/) have been released.


AlexWayfer:
> Add support for ruby 2.5.4

Thanks! But I think this PR should be renamed after that.

Issue #1292, opened 2019-03-17
Install of JRuby 1.6.8 fails when building jruby-launcher


jayjlawrence:
Yes, old unsupported version, I have technical debt that will go away eventually. Below is the output that explains the issue on OS X and Linux (Ubuntu 16.04 using Open JDK).

The (optional) install of ruby-launcher fails under jruby-1.6.8. It is not clear if jruby-launcher ever supported 1.6.8. This is not a show-stopper so I'd like to just leave it out.

This issue is to describe a PR that parameterizes the building of jruby-launcher and sets this for 1.6.8. I tried to follow the lead with other code in the ruby-build shell script.

Offending output on Linux:

```
BUILD FAILED (Ubuntu 16.04 using ruby-build 20180618-5-gad8f3a4)

Inspect or clean up the working tree at /tmp/ruby-build.20190317123348.22262
Results logged to /tmp/ruby-build.20190317123348.22262.log

Last 10 log lines:
        /home/deploy/.rbenv/versions/jruby-1.6.8/bin/jruby extconf.rb
OpenJDK 64-Bit Server VM warning: You have loaded library /home/deploy/.rbenv/versions/jruby-1.6.8/lib/native/i386-Linux/libjffi-1.0.so which might have disabled stack guard. The VM will try to fix the stack guard now.
It's highly recommended that you fix the library with 'execstack -c <libfile>', or link it with '-z noexecstack'.
NameError: uninitialized constant RbConfig
  const_missing at org/jruby/RubyModule.java:2647
         (root) at extconf.rb:2


Gem files will remain installed in /home/deploy/.rbenv/versions/jruby-1.6.8/lib/ruby/gems/1.8/gems/jruby-launcher-1.1.9-java for inspection.
Results logged to /home/deploy/.rbenv/versions/jruby-1.6.8/lib/ruby/gems/1.8/gems/jruby-launcher-1.1.9-java/./gem_make.out
```

and on OS X

```
ERROR:  Error installing jruby-launcher:
	ERROR: Failed to build gem native extension.

        /Users/HOME/.rbenv/versions/jruby-1.6.8/bin/jruby extconf.rb
NameError: uninitialized constant RbConfig
  const_missing at org/jruby/RubyModule.java:2647
         (root) at extconf.rb:2


Gem files will remain installed in /Users/HOME/.rbenv/versions/jruby-1.6.8/lib/ruby/gems/1.8/gems/jruby-launcher-1.1.9-java for inspection.
[...]

PR #1293, opened 2019-03-17
Make building of jruby-launcher gem optional

jayjlawrence:
The Jruby-launcher gem is not compatible with all versions of jruby. As well it may not always be desirable to have this gem automatically installed by ruby-build. Updated jruby-1.6.8 as it requires this per #1292.

 It may be preferable to have the jruby-launcher separated out from the build_package_jruby function along the lines of other platforms. This patch tries to follow current code patterns (but I may have misjudged) and to be of minimal impact.

Took the liberty of modifying the code. Was unsure about the test suite.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ The build process may be configured through the following environment variables:
 | `RUBY_CONFIGURE_OPTS`    | Additional `./configure` options (applies only to Ruby source).                                  |
 | `RUBY_MAKE_OPTS`         | Additional `make` options (applies only to Ruby source).                                         |
 | `RUBY_MAKE_INSTALL_OPTS` | Additional `make install` options (applies only to Ruby source).                                 |
+| `JRUBY_DISABLE_LAUNCHER` | Do not install the jruby-launcher gem when installing JRuby                                      |
 
 #### Applying Patches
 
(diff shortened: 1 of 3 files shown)


nirvdrum:
It looks like you've only applied this to JRuby 1.6.8. Is that the only version that's problematic or is that the only one you've tested with? I'm trying to ascertain whether the other jruby-1.6.x and jruby-1.5.x definitions should also disable the launcher.

jayjlawrence:
This didn't sit right with me so I had a look at jruby-launcher. We just need a require 'rbconfig' in the extconf to make the underlying issue go away. I've submitted jruby/jruby-launcher#23 to address that. They will need to push a new gem.

Nonetheless I believe that automatically installing jruby-launcher as part of the JRuby build is undesirable. It should be a post build task. 

I would recommend phasing it out - announce for 6 months to all jruby installs "Installing jruby-launcher (Phased out after Sept 1, 2019 )" . Then "No longer installing jruby-launcher by default; put it in your Gemfile or other post build tasks"

If you want I can do the above - subject to your adjustments.

nirvdrum:
I think @headius or @enebo should weigh in on this, since it impacts how many people install and run JRuby.

jayjlawrence:
I dug a bit deeper and addressed the build error for jruby-launcher. Submitted jruby/jruby-launcher/pull/23 to address that issue. I expect that this should improve for most versions of jruby. I confirmed with 1.6.8.

I still think that at a minimum the ruby-build user should be able to opt-out of the installation of jruby-launcher.

mislav:
Thank you, but see https://github.com/rbenv/ruby-build/issues/1292#issuecomment-549730708 please.

PR #1347, opened 2019-09-19
Look for Homebrew openssl@1.1

gfguthrie:
ruby-build formula [depends on openssl@1.1](https://github.com/Homebrew/homebrew-core/blob/master/Formula/ruby-build.rb#L11), let’s look for that version

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1021,7 +1021,7 @@ has_broken_mac_openssl() {
 }
 
 use_homebrew_openssl() {
-  local ssldir="$(brew --prefix openssl 2>/dev/null || true)"
+  local ssldir="$(brew --prefix openssl@1.1 2>/dev/null || true)"
   if [ -d "$ssldir" ]; then
     echo "ruby-build: using openssl from homebrew"
     package_option ruby configure --with-openssl-dir="$ssldir"


hsbt:
I didn't understand why homebrew depends on openssl@1.1. 

@igas Can you elaborate this?

igas:
Homebrew relies on the latest versions of software by default. Recently we moved all the formulas that build with openssl 1.1 to it. If ruby-build does not support 1.1, please submit PR to stick to `openssl@1.0`. It would be great to improve `test` block so we do not accidentally bump it in the future and break formula again. I'm happy to help from homebrew side if you need an assistance.

gfguthrie:
For my part, I'll add that all of the Ruby versions that ruby-build supports that are not EOL already build with 1.1, and so that is why I thought this would be a good change. The version of openssl can always be specified as part of the build if the user wishes to do something different, but perhaps it's best for latest versions to now be the default.

hsbt:
@igas This change broke the old ruby versions. see https://github.com/rbenv/ruby-build/issues/1353

igas:
I see, I don't think there are good way to handle this.

I propose to drop dependency all together on brew side. On the ruby-build we can look for which version of ruby is requested and install (if needed) `openssl` or `openssl@1.1` accordingly and use it in path.

What do you think guys?

gfguthrie:
There are two ways to go IMO:

1) Stick with how the formula/ruby-build currently are, where we're consistent with Homebrew philosophy of favoring current versions. (There is a workaround for installing EOL versions of Ruby with Homebrew openssl, it probably should just be documented somewhere other than my comment on the issue).

2) Remove the dependency from the formula on any version of openssl, and also remove looking for any Homebrew version of openssl in ruby-build. That way all versions of Ruby will install properly with the version of openssl installed by ruby-build, and instead the workaround to use Homebrew openssl will apply to everyone who wants to only have one version of openssl installed.

yyyc514:
https://github.com/rbenv/ruby-build/issues/1355

I'm just a single data point but this PR breaks my build of Ruby 2.6.5 (and 2.6.4)... I'm guessing maybe others... I just reverting to the commit prior (which builds against 1.0) and all seems to work fine. My 2.6.3 was also built against 1.0 prior to this change and I had no issues.

yyyc514:
Well, take back what I said... when I removed ALL of homebrew stuff from my path now it compiles just fine, so it's something about the build process not liking something I have in Homebrew.  I will add more comments to the other thread. :-)  Back to the prior discussion.

igas:
I'm happy to adjust brew side to what upstream will choose.

yyyc514:
My issue (as documented on the other thread) is that Brew's`pkg-config` was in my path and Ruby (when built against OpenSSL 1.1) blows up for some reason if it can see that `pkg-config`.  But I'm happy now and I documented it for others... so...

PR #1348, opened 2019-09-23
Add example of mirrored URL.

ross-bragg:
I found the documentation on how to install ruby from a mirrored URL extremely unclear and found issues from folks who had also had that problem. This PR adds an example of a package downloaded from cache.ruby-lang.org and its corresponding download url on a mirror.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ The build process may be configured through the following environment variables:
 | `RUBY_BUILD_CURL_OPTS`   | Additional options to pass to `curl` for downloading.                                            |
 | `RUBY_BUILD_WGET_OPTS`   | Additional options to pass to `wget` for downloading.                                            |
 | `RUBY_BUILD_MIRROR_URL`  | Custom mirror URL root.                                                                          |
-| `RUBY_BUILD_SKIP_MIRROR` | Always download from official sources, not mirrors. (Default: unset)                             |
+| `RUBY_BUILD_SKIP_MIRROR` | Bypass the download mirror and fetch all package files from their original URLs.                  |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -123,7 +123,10 @@ URL specified in the definition file.
 You can point ruby-build to another mirror by specifying the
 `RUBY_BUILD_MIRROR_URL` environment variable--useful if you'd like to run your
 own local mirror, for example. Package mirror URLs are constructed by joining
-this variable with the SHA2 checksum of the package file.
+this variable with the SHA2 checksum of the package file. For example, the default
+URL for `ruby-2.3.0.tar.bz2` is http://cache.ruby-lang.org/pub/ruby/ruby-2.3.0.tar.bz2.
+That package would be mirrored at `$RUBY_BUILD_MIRROR_URL/ec7579eaba2e4c402a089dbc86c98e5f1f62507880fd800b9b34ca30166bfa5e`,
```
mislav: I'm sorry, but I don't think this example clarifies anything over what was already being said.

“Package mirror URLs are constructed by joining [RUBY_BUILD_MIRROR_URL] with the SHA2 checksum of the package file” → `<RUBY_BUILD_MIRROR_URL>/<SHA2>`. Was this unclear?

ross-bragg: Yes. I wasn't expecting moving from the usual "file with an extension" to download to "arbitrary path with no extension". It took me some time and stepping through the script to realize it wasn't `$RUBY_BUILD_MIRROR/$SHASUM/$PACKAGE`.

mislav:
Thanks for explaining. I've pushed some commits to change the wording further. I ended up restructuring the whole section. Care to take a look?

ross-bragg:
That's a lot better, yeah. Thanks!

PR #1362, opened 2019-10-09
Don't use Homebrew's openssl when --with-openssl-dir= is specified

genkami:
`use_homebrew_openssl()` overwrites `$RUBY_CONFIGURE_OPTS` so we can't use custom openssl when openssl@1.1 is installed via Homebrew.

I fixed this problem.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1016,18 +1016,21 @@ has_broken_mac_openssl() {
   is_mac || return 1
   local openssl_version="$(/usr/bin/openssl version 2>/dev/null || true)"
   [[ $openssl_version = "OpenSSL 0.9.8"?* || $openssl_version = "LibreSSL"* ]] &&
-  [[ "$RUBY_CONFIGURE_OPTS" != *--with-openssl-dir=* ]] &&
   ! use_homebrew_openssl
 }
 
 use_homebrew_openssl() {
   local ssldir="$(brew --prefix openssl@1.1 2>/dev/null || true)"
-  if [ -d "$ssldir" ]; then
+  if [ -d "$ssldir" ] && ! openssl_dir_specified; then
     echo "ruby-build: using openssl from homebrew"
     package_option ruby configure --with-openssl-dir="$ssldir"
   else
     return 1
   fi
+}
+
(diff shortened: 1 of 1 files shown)


bin/ruby-build
```diff
@@ -1016,20 +1016,23 @@ has_broken_mac_openssl() {
   is_mac || return 1
   local openssl_version="$(/usr/bin/openssl version 2>/dev/null || true)"
   [[ $openssl_version = "OpenSSL 0.9.8"?* || $openssl_version = "LibreSSL"* ]] &&
-  [[ "$RUBY_CONFIGURE_OPTS" != *--with-openssl-dir=* ]] &&
```
mislav: You've removed this line. This has served as a guard that Homebrew openssl never gets used if an explicit `--with-openssl-dir` was passed. You have removed this and placed the guard elsewhere. 

Do you have custom Ruby install formula by any chance?

genkami: Consider the case that we have installed openssl@1.1 via Homebrew and want to build ruby with `--with-openssl-dir=` option.
Then this line is evaluated to truthy value, and thus `&&` invokes `use_homebrew_openssl`, resulting in setting `--with-openssl-dir="$ssldir"`.

mislav: > Consider the case that we have installed openssl@1.1 via Homebrew and want to build ruby with `--with-openssl-dir=` option.
> Then this line is evaluated to truthy value, and thus `&&` invokes

If you had passed your own `--with-openssl-dir=` value, then the `[[ ... ]]` check evaluates to _false_, and nothing on the right side of `&&` is invoked.

mislav:
```sh
RUBY_CONFIGURE_OPTS="--with-openssl-dir=/foo/bar"

if [[ "$RUBY_CONFIGURE_OPTS" != *--with-openssl-dir=* ]]; then
  echo 'I will try to use Homebrew openssl!'
else
  echo 'I will NOT try to use Homebrew openssl.' # <-- this gets printed
fi
```

genkami: Sorry, I completely misunderstood it...

PR #1372, opened 2019-10-31
jRuby 9.x depends on java 8

### LINKED ISSUE #1313: JRuby 9.2+ requires Java 8, not 7
Minor issue...I noticed the definitions for JRuby 9k (9.x.x.x) all have "java7" as a required prerequisite, but since JRuby 9.2 we have required Java 8. The definitions should be tweaked.

afromankenobi:
Fix #1313

### DIFF
--- a/share/ruby-build/jruby-9.1.9.0
+++ b/share/ruby-build/jruby-9.1.9.0
@@ -1,2 +1,2 @@
-require_java7
+require_java8
 install_package "jruby-9.1.9.0" "https://s3.amazonaws.com/jruby.org/downloads/9.1.9.0/jruby-bin-9.1.9.0.tar.gz#36b802050155dccd808b4f69f01db4aa1599de467be657bdf81659aab9e8084e" jruby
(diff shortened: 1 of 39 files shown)


share/ruby-build/jruby-9.1.9.0
```diff
@@ -1,2 +1,2 @@
-require_java7
+require_java8
```
mislav: Can we leave 9.0 and 9.1 definitions intact, and just update jruby 9.2+ definitions?

afromankenobi:
Yeah of course. Question, why?

share/ruby-build/jruby-9.1.9.0
afromankenobi: Updated. Please confirm! :)

mislav:
Heads up: this broke things https://github.com/rbenv/ruby-build/issues/1373

I will fix it, but for the future, it's always a good idea to test your changes before you open pull requests. Thank you!

afromankenobi:
Oh! So sorry. I just created the method for test locally but forgot to commit. Thanks for the correction

mislav:
No worries! Ideally, our test suite should have caught things like these, but right now it doesn't source any definitions to check if all the directives make sense. Ideally, the test suite should "run" all definitions, but just stub out the download/configure/make part so it doesn't actually happen.

Issue #1360, opened 2019-10-04
OpenSSL issue installing Ruby 2.6.5


jbirdjavi:
I'm getting an error installing ruby 2.6.5

[ruby-build.20191004151402.40644.log](https://github.com/rbenv/ruby-build/files/3692192/ruby-build.20191004151402.40644.log)

➜  ~ rbenv install 2.6.5
ruby-build: using openssl from homebrew
Downloading ruby-2.6.5.tar.bz2...
-> https://cache.ruby-lang.org/pub/ruby/2.6/ruby-2.6.5.tar.bz2
Installing ruby-2.6.5...
ruby-build: using readline from homebrew

BUILD FAILED (OS X 10.14.6 using ruby-build 20191004)

Inspect or clean up the working tree at /var/folders/yz/51nhy_gj6xb76r9bzw5qngw00000gn/T/ruby-build.20191004151402.40644
Results logged to /var/folders/yz/51nhy_gj6xb76r9bzw5qngw00000gn/T/ruby-build.20191004151402.40644.log

Last 10 log lines:
ossl_x509req.c:266:5: warning: implicit declaration of function 'X509_REQ_get0_signature' is invalid in C99 [-Wimplicit-function-declaration]
    X509_REQ_get0_signature(req, NULL, &alg);
    ^
7 warnings generated.
1 warning generated.
7 warnings generated.
make[1]: *** [ext/openssl/all] Error 2
make[1]: *** Waiting for unfinished jobs....
linking shared-object ripper.bundle
make: *** [build-ext] Error 2

jbirdjavi:
I'm also getting an error installing ruby 2.6.4, which had installed successfully on a previous version of ruby-build.

[ruby-build.20191004152553.83968.log](https://github.com/rbenv/ruby-build/files/3692225/ruby-build.20191004152553.83968.log)

DocEmmetBrown:
I've got the exact same issue. Up to date MacOS 10.14.6.

jbirdjavi:
The workaround @DocEmmetBrown gave also works for me.  I assume that means that there are still some incompatibilities with openssl@1.1.

ashwinvidiyala:
@DocEmmetBrown Worked for me too. Thank you!

hjijin:
[...]

Issue #1353, opened 2019-10-03
Cannot install Ruby versions < 2.4 because of openssl@1.1 dependency


brandonfriess-stripe:
It seems the most recent change to ruby-build 20191002 homebrew formula to require openssl@1.1 breaks the install of Ruby versions that are < 2.4 (those require openssl 1.0.2q)

So, for macOS users who install ruby-build via homebrew, they will get openssl@1.1 installed as well.

If they then try to run `rbenv install 2.1.10` for instance, it will fail with:

```
Last 10 log lines:
The Ruby openssl extension was not compiled.
ERROR: Ruby install aborted due to missing extensions
Configure options used:
  --prefix=/Users/username/.rbenv/versions/2.1.10
  --with-openssl-dir=/usr/local/opt/openssl@1.1
  --with-readline-dir=/usr/local/opt/readline
  CC=clang
  CFLAGS= -O3 -Wno-error=shorten-64-to-32 
  LDFLAGS=-L/Users/username/.rbenv/versions/2.1.10/lib 
  CPPFLAGS=-I/Users/username/.rbenv/versions/2.1.10/include 
```

Yeah, I know older versions are EOL but they still get used in some environments.

Seems like changing https://github.com/rbenv/ruby-build/blob/master/bin/ruby-build#L1023 to only check for `openssl`  and not `openssl@1.1 could potentially fix this?

hsbt:
I understand this issue too. But I didn't know the maintenance policy of ruby-build formula on homebrew. 

I suggest you use `git clone` instead of `brew install ruby-build`

gfguthrie:
Since I helped move the 1.1 change along:

The workaround for EOL Ruby versions (or any that won't build with 1.1) where you still want to use Homebrew is as follows:
`brew install openssl`
`RUBY_CONFIGURE_OPTS="--with-openssl-dir=/usr/local/opt/openssl" rbenv install 2.3.8`

brandonfriess-stripe:
Thanks! That will work and I am sure will be helpful for others who may land on this thread.

aardvarkk:
I ran into this bug when trying to install Ruby 2.3.8 using the ASDF version manager. I was getting the following error when running:

`asdf install ruby 2.3.8`:

**ERROR: Ruby install aborted due to missing extensions**

Following the example above, I changed the command to the following and the installation worked correctly:

[...]

PR #1375, opened 2019-11-08
Don’t look for Homebrew openssl

gfguthrie:
`ruby-build` already does a lot of work to make sure that the version of ruby you're installing has the correct version of openssl also installed. Looking for any version of openssl installed by Homebrew is problematic for a few reasons:
- The user may have an incompatible version installed
- The user may not know that a particular version of openssl has been installed by Homebrew
- The openssl version split now makes this logic more difficult to maintain for all versions of ruby that `ruby-build` can install

Let's stop looking for any version of Homebrew's openssl by default so that the user always gets a working version of the ruby they're installing. Advanced users can use the `RUBY_CONFIGURE_OPTS="--with-openssl-dir=/my-openssl-directory"` option if they don't want to install the `ruby-build` defined compatible version of openssl.

#1353 #1360 #1347

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1023,6 +1023,16 @@ has_broken_mac_openssl() {
   [[ "$RUBY_CONFIGURE_OPTS" != *--with-openssl-dir=* ]]
 }
 
+use_homebrew_openssl() {
+  local ssldir="$(brew --prefix openssl@1.1 2>/dev/null || true)"
+  if [ -d "$ssldir" ]; then
+    echo "ruby-build: using openssl from homebrew"
+    package_option ruby configure --with-openssl-dir="$ssldir"
+  else
+    return 1
+  fi
+}
+
 build_package_mac_openssl() {
   # Install to a subdirectory since we don't want shims for bin/openssl.
   OPENSSL_PREFIX_PATH="${PREFIX_PATH}/openssl"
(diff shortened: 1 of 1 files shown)


bin/ruby-build
```diff
@@ -1020,18 +1020,7 @@ has_broken_mac_openssl() {
   is_mac || return 1
   local openssl_version="$(/usr/bin/openssl version 2>/dev/null || true)"
   [[ $openssl_version = "OpenSSL 0.9.8"?* || $openssl_version = "LibreSSL"* ]] &&
-  [[ "$RUBY_CONFIGURE_OPTS" != *--with-openssl-dir=* ]] &&
-  ! use_homebrew_openssl
-}
-
-use_homebrew_openssl() {
```
mislav: Can you leave this function as-is? Custom definitions may be referencing it

gfguthrie: good point - had not thought of that!

hsbt:
Looks good.

gfguthrie:
@wadkar the reasons are in the PR #1375

wadkar:
Ah, I see. Thanks for pointing me here. I understand it now. Also checked `brew info ruby-build`'s caveat (though I use basic github checkout):

```sh
$ brew info ruby-build
==> Caveats
ruby-build installs a non-Homebrew OpenSSL for each Ruby version installed and these are never upgraded.

To link Rubies to Homebrew's OpenSSL 1.1 (which is upgraded) add the following to your ~/.zshrc:
  export RUBY_CONFIGURE_OPTS="--with-openssl-dir=$(brew --prefix openssl@1.1)"

Note: this may interfere with building old versions of Ruby (e.g <2.4) that use OpenSSL <1.1.
```

I have updated the wiki accordingly. Please edit it if there are any inaccuracies.

PR #1380, opened 2019-11-19
Enable users to discover how to update the ruby-build plugin

jsoref:


mislav:
Thank you for the suggestion, but I think our current behavior is enough to help people upgrade.

1. When the definition to install is not found, this is printed: “If the version you need is missing, try upgrading ruby-build: {upgrade instructions}”
2. https://github.com/rbenv/ruby-build/wiki#updating-ruby-build

mislav:
I understand that you're concerned that some people could be confused as to why a specific version isn't listed in `ruby-build --definitions` or `rbenv install --list`, and that they might not notice that ruby-build is out of date.

I'm open to other, more sophisticated solutions to this problem, such as dynamic fetching a ruby-build definition from this repo, but I don't believe that adding extra output to stderr or instructing people to run a command which will result in an error is the way to go.

jsoref:
I can't speak so much to the rbenv flavor. For pyenv, two of us really did not figure out the problem. We decided pyenv was broken. We tried uninstalling and reinstalling pyenv repeatedly and it didn't help.

The current model assumes that cautious users will intentionally ask a tool to install something that doesn't exist, after explicitly asking for a list of things that does exist.

I can assure you that as a cautious user, I don't do that.

When I look for Bridges across a river, I don't decide to walk out into thin air where I'd like a bridge to be. Surprisingly, I have yet to drown by following this strategy. I have been moderately annoyed that the local bridge building authority hasn't put in place all possible Bridges that I might want to have. But, they did recently install a new bridge, and I've used it, without drowning.

(Actually, the local bridges are mostly over train tracks, so it's more like I haven't been hit by a train rather than drowned in water, but…)

PR #1381, opened 2019-11-20
Set default MAKE=gmake on FreeBSD again

metalefty:
as an workaround for Ruby bug 16331: https://bugs.ruby-lang.org/issues/16331

Due to this bug, build will fail with FreeBSD's make after #1368.
The bug is already fixed in MRI upstream but GNU make is still required
when building older releases of Ruby. Use GNU make rather than switching
make/gmake depending of Ruby version.

See also: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=241633

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1320,15 +1320,12 @@ fi
 
 if [ -z "$MAKE" ]; then
   if [ "FreeBSD" = "$(uname -s)" ]; then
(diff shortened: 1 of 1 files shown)


metalefty:
I'm sorry about not fixing the test together. I completely forgot it. Thanks!

PR #1383, opened 2019-11-21
Use openssl from Homebrew for TruffleRuby on macOS

eregon:
Fixes https://github.com/oracle/truffleruby/issues/1818

* The openssl built by ruby-build was not used by TruffleRuby,
  and causes https://github.com/oracle/truffleruby/issues/1818
* Improve error message when openssl from Homebrew is not available.
* Change the definition code so it checks the operating system too.

cc @hsbt


eregon:
I'll merge this now as it's kind of an urgent fix for TruffleRuby macOS ruby-build/rbenv users.

hsbt:
👍

PR #1397, opened 2020-01-11
Update so Ruby 1.9.3 thru 2.3, and rbx can be built with OpenSSL 1.0.2u

MSP-Greg:
1. Change all `share/ruby-build/2.3.x` files to use OpenSSL 1.0.2u
2. Update cert code for Ubuntu & macOS

This is for use in generating Rubies for use on GitHub Actions.  A build run is here:

https://github.com/MSP-Greg/ruby-install-builder/runs/384299585

### DIFF
--- a/share/ruby-build/1.9.3-dev
+++ b/share/ruby-build/1.9.3-dev
@@ -1,3 +1,3 @@
-install_package "openssl-1.0.2q" "https://www.openssl.org/source/openssl-1.0.2q.tar.gz#5744cfcbcec2b1b48629f7354203bc1e5e9b5466998bbccc5b5fcde3b18eb684" mac_openssl --if has_broken_mac_openssl
+install_package "openssl-1.0.2u" "https://www.openssl.org/source/openssl-1.0.2u.tar.gz#ecd0c6ffb493dd06707d38b14bb4d8c2288bb7033735606569d8f90f89669d16" mac_openssl --if has_broken_mac_openssl
 install_package "yaml-0.1.6" "http://pyyaml.org/download/libyaml/yaml-0.1.6.tar.gz#7da6971b4bd08a986dd2a61353bc422362bd0edcc67d7ebaac68c95f74182749" --if needs_yaml
 install_git "ruby-1.9.3-dev" "https://github.com/ruby/ruby.git" "ruby_1_9_3" warn_eol autoconf standard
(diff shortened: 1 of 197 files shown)


bin/ruby-build
```diff
@@ -1063,10 +1063,18 @@ build_package_mac_openssl() {
...
+    # Extract root certs from the system keychain in .pem format and rehash.
+    security find-certificate -a -p /Library/Keychains/System.keychain > "$pem_file"
+    security find-certificate -a -p /System/Library/Keychains/SystemRootCertificates.keychain >> "$pem_file"
+  else
+    # See https://github.com/rubygems/rubygems/issues/2415#issuecomment-509806259
+    # Debian-based logic:
+    rm -rf "$OPENSSLDIR/certs" "$pem_file"
+    ln -s /etc/ssl/certs "$OPENSSLDIR"
+    ln -s /etc/ssl/certs/ca-certificates.crt "$pem_file"
+  fi
```
eregon: This logic needs more tweaks, it needs to support other OS, and have some kind of fallback.
We probably want to rename `build_package_mac_openssl` to just `build_package_openssl` or so.
In general I think it would be better to have 2 PRs:
* One updating OpenSSL and no other changes.
* One building OpenSSL on Linux for Ruby <= 2.3 (since it's hard to get OpenSSL 1.0.2 from Linux distributions).

hsbt:
@MSP-Greg Can you separate the pull-request to update versions of OpenSSL and change the certificate logic at least?

MSP-Greg:
Thanks all.

I'm revised the PR, removing the cert logic, leaving the `--if has_broken_mac_openssl` logic, and splitting it into Ruby, rbx 2 & rbx 3.

As is, it does work on Ubuntu, but as mentioned, leaving it OS specific is probably best.

eregon:
There are instructions in the wiki how to workaround on Linux for old OpenSSL: https://github.com/rbenv/ruby-build/wiki#openssl-usrincludeopensslasn1_mach102-error-error-this-file-is-obsolete-please-update-your-software (last paragraph)
I think it would still be useful for `ruby-build` to have a way to install old OpenSSL 1.0.x automatically, since it's becoming increasingly difficult to install packages for it on Linux.
That's a topic for another PR though, I agree.

MSP-Greg:
The previous commit that updated to 1.0.2q had quite a few files.  1.0.2u should be compatible with 1.0.1, but I thought I'd split the commits up in case there are issues and some need to be reverted.

MSP-Greg:
Fixed commit subject typo 'OpeSSL' => 'OpenSSL'

PR #1398, opened 2020-01-11
Create definition file truffleruby-19.3.0.2

ipepe:
I noticed that there is new version for truffleruby and created a file for installing it.

### DIFF
--- a/share/ruby-build/truffleruby-19.3.0.2
+++ b/share/ruby-build/truffleruby-19.3.0.2
@@ -1,0 +1,13 @@
+case $(uname -s) in
+Linux)
+  install_package "truffleruby-19.3.0.2" "https://github.com/oracle/truffleruby/releases/download/vm-19.3.0.2/truffleruby-19.3.0.2-linux-amd64.tar.gz" truffleruby
+  ;;
+Darwin)
+  use_homebrew_openssl
+  install_package "truffleruby-19.3.0.2" "https://github.com/oracle/truffleruby/releases/download/vm-19.3.0.2/truffleruby-19.3.0.2-macos-amd64.tar.gz" truffleruby
+  ;;
+*)
+  colorize 1 "Unsupported operating system: $(uname -s)"
+  return 1
+  ;;
+esac


ipepe:
Thanks @hsbt

share/ruby-build/truffleruby-19.3.0.2
```diff
@@ -0,0 +1,13 @@
+case $(uname -s) in
+Linux)
+  install_package "truffleruby-19.3.0.2" "https://github.com/oracle/truffleruby/releases/download/vm-19.3.0.2/truffleruby-19.3.0.2-linux-amd64.tar.gz" truffleruby
```
eregon: This is missing the checksums, I added them in 4082f9eff2f4cdd84ec1b512f0b2046458266e42

ipepe: I wasn't sure what those were. Thanks for fixing this @eregon

ipepe: For future reference, what type of checksums are those and how did You generate proper hash?

eregon:
@hsbt Please @-mention me on TruffleRuby PRs.

[19.3.0.2](https://www.graalvm.org/docs/release-notes/19_3/#19302) has AFAIK no relevant changes for TruffleRuby so that's why I didn't add it. But it doesn't hurt to add it either.

share/ruby-build/truffleruby-19.3.0.2
eregon: I use `script/update-truffleruby` documented in https://github.com/oracle/truffleruby/blob/master/doc/contributor/updating-ruby-installers.md
Then the checksums are automatically computed.

In general I prefer to push updates to Ruby installers myself because it's more efficient and consistent (I have push rights on all 3 Ruby installers).

ipepe: My bad! Interesting thing is that I just checked https://github.com/oracle/truffleruby/releases and commit for 19.3.0.2 is exactly the same as for 19.3.0 . But checksums differ between these versions.

eregon: Yeah it was rebuilt anyway, it could matter e.g., if there was a Graal compiler or Truffle fix in there.

hsbt:
ok, I will never touch with truffleruby. I'm tired to handle truffleruby.

eregon:
@hsbt Yes I meant just please assign me for TruffleRuby PRs in `ruby-build`, I'm happy to handle them :smiley:

eregon:
@hsbt I'll just follow the PRs on this repo instead.
I shouldn't ask any extra work from you.

Thank you for helping with maintaining `ruby-build` :pray:

PR #1404, opened 2020-01-31
shorter git command chains

eladeyal-intel:


jasonkarns:
I like this! Do you know which was the first version of git that supported the `-C` option, so we know where we'd be dropping support?

eladeyal-intel:
Git v1.8.5 2013 https://github.com/git/git/blob/master/Documentation/RelNotes/1.8.5.txt

eladeyal-intel:
@jasonkarns

jasonkarns:
Thanks @eladeyal-intel! I believe this is also has a secondary benefit of helping out Fish shell users who are running old versions of fish which don't support `&&`. 🚀

PR #1412, opened 2020-03-08
Added rbx-4.*

hsbt:


script/update-rbx-defs
```diff
@@ -8,7 +8,7 @@ dir = Pathname(File.expand_path("share/ruby-build"))
 
 open(file).each do |package|
   sha256, filename = package.chomp.split
-  version = filename.match(/rubinius-(3.[\d\.]+).tar.bz2/)
+  version = filename.match(/rubinius-([34].[\d\.]+).tar.bz2/)
```
mislav: Should we make this regex more permissive? Won't we run into the same problem when rbx 5.0 comes out?

hsbt: We should use `\d` instead of `[34]` 👍  It's caused by my laziness.

PR #1414, opened 2020-03-10
Add update-mri and update-jruby scripts

koic:
This PR adds update-mri and update-jruby scripts.
They are provided to update MRI and JRuby definition files.

These scripts are based on script/update-truffleruby.
https://github.com/rbenv/ruby-build/blob/v20200224/script/update-truffleruby

### DIFF
--- a/script/update-cruby
+++ b/script/update-cruby
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+
+set -e
+set -o pipefail
+
+if [ $# -ne 2 ]; then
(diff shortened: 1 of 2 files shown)


hsbt:
Can you rename `script/update-mri` to `script/update-ruby` or `script/update-cruby` ? I prefer to use cruby instead of mri.

koic:
Thanks for your feedback. I renamed script/update-mri to script/update-cruby.

PR #1419, opened 2020-03-19
s/all built-in definitions/all available definitions/

metalefty:
Suggested by mislav: https://github.com/rbenv/ruby-build/pull/1402#discussion_r389060742

> I wonder should we substitute "built-in" with something like "all
> available definitions". Since ruby-build will additionally pick up
> definitions found under the paths listed in RUBY_BUILD_DEFINITIONS,
> these definitions will get listed even if they are not "built-in".
> Therefore, "built-in" feels like a misnomer here.

Searched by `git grep built-in`.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -9,7 +9,7 @@
 #   -v/--verbose     Verbose mode: print compilation status to stdout
 #   -4/--ipv4        Resolve names to IPv4 addresses only
 #   -6/--ipv6        Resolve names to IPv6 addresses only
-#   --definitions    List all built-in definitions
+#   --definitions    List all available definitions
 #   --version        Show version of ruby-build
 #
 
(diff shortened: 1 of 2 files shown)


bin/ruby-build
```diff
@@ -9,7 +9,7 @@
 #   -v/--verbose     Verbose mode: print compilation status to stdout
 #   -4/--ipv4        Resolve names to IPv4 addresses only
 #   -6/--ipv6        Resolve names to IPv6 addresses only
-#   --definitions    List all built-in definitions
+#   --definitions    List all available definitions
```
jasonkarns: I'm concerned that "available" implies ruby-build will be checking online sources, or is otherwise aware of all ruby releases. That has actually been a point of confusion in the past.

Perhaps "all locally available definitions" ?

mislav: Good point! I'd be also fine with `all local definitions`

metalefty: Thanks for the feedback. Changed to "local".

PR #1420, opened 2020-03-20
mark maglev as EoL

metalefty:
as the last commit of mavlev is over 3 years ago

Discussed at: https://github.com/rbenv/ruby-build/pull/1402#issuecomment-596162716


metalefty:
Good point. I also think "EoL" should be based on official announcements. Marking maglev as `warn_eol` a little bit misnomer. AFAIK, there's no official announcements from Maglev team. 

I suggested alternate `warn_abandoned` tag in https://github.com/rbenv/ruby-build/pull/1402#issuecomment-596162716 . I think this is better to describe current mavlev state. However, I feel implementing `warn_abandoned` only to hide mavlev from the list is not worth the effort. That's the reason of this PR. IMO, I know marking maglev as `warn_eol` is not best but an acceptable workaround.

Differences between `warn_*`:

- `warn_eol` 
    -  officially announced as EoL
- `warn_abandoned` (suggested by me)
    - considered as abandoned
    - no active developments for a long time with no official announcements

hsbt:
I decided to merge this. The last commit of Maglev is 4 years ago.

PR #1422, opened 2020-03-26
Add definition for JRuby 9.2.11.1

koic:
JRuby 9.2.11.1 has been released.
https://www.jruby.org/2020/03/25/jruby-9-2-11-1


mislav:
Thank you!

PR #1424, opened 2020-03-31
add definition 2.4.10, 2.5.8, 2.6.6, 2.7.1 and ruby 2.4 EOL

jnozsc:
https://www.ruby-lang.org/en/news/2020/03/31/ruby-2-4-10-released/
https://www.ruby-lang.org/en/news/2020/03/31/ruby-2-5-8-released/
https://www.ruby-lang.org/en/news/2020/03/31/ruby-2-6-6-released/
https://www.ruby-lang.org/en/news/2020/03/31/ruby-2-7-1-released/
https://www.ruby-lang.org/en/downloads/branches/


mislav:
Thank you!!

JulienItard:
Hi, I dont get this new version when i try to upgrade it with homebrew, is it deployed ?

PatrickLerner:
It first needs to be released here and then it can be bumped in homebrew. It's still only on master so far. I guess mislav will get around to that in a bit :)

JulienItard:
@PatrickLerner 👍

PR #1432, opened 2020-04-09
mark ruby 2.5 as security maintenance

jnozsc:
merge when https://github.com/ruby/www.ruby-lang.org/pull/2410 and https://github.com/ruby/www.ruby-lang.org/pull/2420 get released


jasonkarns:
Is this good to merge now, @hsbt ?

PR #1451, opened 2020-06-23
compatible with macOS 11

jnozsc:
https://en.wikipedia.org/wiki/MacOS_Big_Sur

stop hardcoding `OS X` since `macOS Big Sur` version no. is 11.0

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -91,7 +91,7 @@ os_information() {
   if type -p lsb_release >/dev/null; then
     lsb_release -sir | xargs echo
   elif type -p sw_vers >/dev/null; then
-    echo "OS X $(sw_vers -productVersion)"
+    echo "$(sw_vers -productName) $(sw_vers -productVersion)"
   elif [ -r /etc/os-release ]; then
     source /etc/os-release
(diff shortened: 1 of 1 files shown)


bin/ruby-build
```diff
@@ -91,7 +91,7 @@ os_information() {
   if type -p lsb_release >/dev/null; then
     lsb_release -sir | xargs echo
   elif type -p sw_vers >/dev/null; then
-    echo "OS X $(sw_vers -productVersion)"
+    echo "$(sw_vers -productName) $(sw_vers -productVersion)"
```
hsbt, line 94: What's return value with `sw_vers -productName` on Big Sur?

jnozsc: https://twitter.com/sleeve/status/1276191458714775552
https://twitter.com/never_released/status/1275444907394490373

it seems on big sur, it will return `macOS`

hsbt: got it. It seems reasonable.

hsbt:
Note, macOS Catalina returns:

```
$ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.15.5
BuildVersion:   19F101
```

PR #1455, opened 2020-07-12
Add support for installing GraalVM

eregon:
* Supports using other GraalVM languages such as JavaScript, Python, R and Java.
* Supports running TruffleRuby on JVM and not only Native (see https://github.com/oracle/truffleruby#truffleruby-runtime-configurations).

@hsbt Can you review?

I tested it locally on Linux (all seems fine) and verified URLs generated by the script are correct.


hsbt:
Does this definition provide the truffleruby with JVM environment? I know that graalvm provides the several language. But they are out of scope of `ruby-build`.

Can we rename `graalvm` to `truffleruby-graalvm` ?

eregon:
It provides mostly two new capabilities:
* Can run TruffleRuby in JVM mode (the existing `truffleruby` definition is only TruffleRuby Native). Different performance trade-offs, and the ability to use Java interop.
* Can use additional languages from GraalVM (JavaScript, Python, R), which can be installed with the `bin/gu` utility. JavaScript is always available since it's part of the base GraalVM.

> I know that graalvm provides the several language. But they are out of scope of `ruby-build`.

That's why this PR only installs the Ruby component by default.
R and Python are additional components, which users can install later if wanted with `gu install R`/`gu install python` (e.g., could be useful for PyCall.rb).

> Can we rename `graalvm` to `truffleruby-graalvm` ?

I think it would be rather confusing and very long to type.

FWIW, `ruby-install` seemed OK with the name `graalvm`: https://github.com/postmodern/ruby-install/pull/373
I'd rather have a consistent name in Ruby installers.

hsbt:
I didn't accept this name with your comment. 

> I think it would be rather confusing and very long to type.

We can complete it with shell-integration. 

> I'd rather have a consistent name in Ruby installers.

It's your convenience.

bin/ruby-build
```diff
@@ -748,7 +748,18 @@ build_package_truffleruby() {
   build_package_copy
 
   cd "${PREFIX_PATH}"
-  "${PREFIX_PATH}/lib/truffle/post_install_hook.sh"
+  "${PREFIX_PATH}/lib/truffle/post_install_hook.sh" || return $?
```
jasonkarns, line 751: What is the purpose for the `|| return $?` ? It's the last line of the function so the last line's exit status will be the exit status for the function, no?

share/ruby-build/graalvm-20.1.0
```diff
@@ -0,0 +1,13 @@
+case $(uname -s) in
+Linux)
+  install_package "graalvm-20.1.0" "https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-20.1.0/graalvm-ce-java8-linux-amd64-20.1.0.tar.gz#4fac212b37cd548831fd6587dd4d59dc068068815aa20323b47fde9529d6bb6e" graalvm
+  ;;
+Darwin)
+  use_homebrew_openssl
+  install_package "graalvm-20.1.0" "https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-20.1.0/graalvm-ce-java8-darwin-amd64-20.1.0.tar.gz#3b9fd8ce84c9162a188fde88907c66990db22af0ff6ae2c04430113253a9a634" graalvm
```
jasonkarns, line 7: How do we choose between the java8 package and the java11 package? When should/would we switch to java11?

bin/ruby-build
```diff
@@ -748,7 +748,18 @@ build_package_truffleruby() {
...
+
+build_package_graalvm() {
+  build_package_copy
+
+  cd "${PREFIX_PATH}"
+  bin/gu install ruby || return $?
+
+  local ruby_home
+	ruby_home=$(bin/ruby -e 'print RbConfig::CONFIG["prefix"]')
+	"$ruby_home/lib/truffle/post_install_hook.sh" || return $?
```
jasonkarns, line 762: What is the purpose for the || return $? ? It's the last line of the function so the last line's exit status will be the exit status for the function, no?

share/ruby-build/graalvm-20.1.0
jasonkarns, line 7: How do we choose between the java8 package and the java11 package? When should/would we switch to java11?

eregon:
@jasonkarns It's a good point about the executables and the `bin/` directory.

There is actually a directory `jre/languages/ruby/bin` with only the Ruby executables:
```
$ ls jre/languages/ruby/bin   
bundle  bundler  gem  irb  rake  rdoc  ri  ruby  truffleruby
```

It might be nice if we could use that instead of the top-level `bin` directory, but it doesn't seem very easy. I'll try something with moving GraalVM under `graalvm/` and then symlink that ruby `bin/` directory.
One issue is it would then no longer include `gu` which as mentioned above can be useful to install GraalVM Python/R.
`jvisualvm`, `jstack`, etc can also be useful to profile & debug TruffleRuby, so it seems inconvenient if they are not in PATH.
I'm not sure there is a simple answer here.

Personally I've been using `graalvm/bin` in PATH and it worked fine (using `chruby` which only changes PATH), but it might be indeed a bigger concern with rbenv's shims.

share/ruby-build/graalvm-20.1.0
eregon: I'm being pragmatic here, since we are in a Ruby installer, TruffleRuby is currently better tested on java 8 so I picked java 8.
I think having both would be too many variants.
Of course we can change it over time, that's easy since there are separate definition files per GraalVM version.

bin/ruby-build
eregon: You're right, it's probably unneeded if it stays the last line.

eregon:
Regarding the name, `graalvm` would make me think all GraalVM executables are available (in PATH), while `truffleruby-graalvm` or `truffleruby-jvm` (by opposition) would make me think only Ruby executables are available

eregon:
@jasonkarns I've pushed a change so now:
```
$ ls
bin  graalvm
$ ls bin/
bundle  bundler  gem  gu  irb  rake  rdoc  ri  ruby  truffleruby
```

Seems a good compromise.

@hsbt I've pushed a commit to rename to `truffleruby-graalvm` (but it fails CI, see below).

I still don't like it, and as someone working on TruffleRuby and on GraalVM I think it's not a good name.
This new definition makes it possible to access all components of GraalVM, it's not "just TruffleRuby", it includes a JVM and other languages, and I think there is no point to hide that.
Of course for a Ruby installer, Ruby is still the "primary language" and so only Ruby executables are in `bin/`, and optional components for other languages are not installed by `ruby-build`.

The official documentation to install TruffleRuby with GraalVM has always been called "Installing GraalVM":
https://github.com/oracle/truffleruby/blob/master/doc/user/installing-graalvm.md

It's fundamentally a different VM (different GC, allocator, JIT vs AOT, JIT tiers, etc), so IMHO it deserves its own name.
With REE, nobody named it `ruby-ree` (it's `ree-VERSION`), and there are far more differences between TruffleRuby JVM and Native than between REE and Ruby 1.8.7.

I think everyone knows TruffleRuby is related to GraalVM, so I believe `graalvm` is clear enough.

Also, it breaks `bin/ruby-build -l` (also seen in CI) which clearly expects anything after `-` to be the version.
I don't think it's a good idea to try to remove that reasonable assumption.
Basically `-graalvm` is considered like `-dev`.
cc @metalefty who wrote `-l`, thoughts?

Also, `bin/ruby-build --definitions` shows this confusing mix (so the new formula is between 2 versions of the `truffleruby` formula):
```
...
truffleruby-dev
truffleruby-graalvm-20.1.0
truffleruby-1.0.0-rc10
truffleruby-1.0.0-rc11
...
```

@hsbt Could you reconsider, or explain your opinion?

@mislav Any opinion on this?

>> I'd rather have a consistent name in Ruby installers.

> It's your convenience.

No, it's not for me, it's about consistency in the Ruby ecosystem.

jasonkarns:
@eregon 

> Seems a good compromise.

Yes, that's fantastic, thanks!

To the name, since `graalvm` is the vm on which ruby runs, i would give that primary position in file listings. And there is the _chance_ (though remote) that other ruby implementations could be written on graalvm; perhaps even another jruby itself.

Given that, I would expect `graalvm+truffleruby`. (leaving room for some hypothetical `graalvm+jruby` or `graalvm+otherjavaruby`)

The reason I suggest the `+` as the joiner is that I believe there is some non-trivial sorting logic in rbenv (and likely other tools, like rbenv plugins) that expects the first `-` to separate the ruby name from the version string.

Personally, I don't think additional rubies are likely to be written on graalvm, so as a user, I'd be content with `graal` or `graalvm` as the ruby name. But _if_ a more explicit name is desired, I'd kindly vote that `graalvm` be first, and that `+` be used as the joiner instead of `-`.

eregon:
Thanks for your feedback.
I'd like to take the occasion to clarify my reasoning regarding the name and what you said.

GraalVM as in https://www.graalvm.org/downloads/ will not have another Ruby implementation. Just TruffleRuby.
JRuby runs on JVM + Graal JIT. That's not GraalVM though. (and JRuby doesn't ship its own JVM, on purpose)

GraalVM includes many things like the Graal JIT (= GraalVM Compiler), AOT compilation (= GraalVM Native Image), Truffle languages including TruffleRuby (= GraalVM Ruby), etc.
GraalVM as a concept includes TruffleRuby, and so having both `graalvm` and `truffleruby` in the name just feels redundant.
As an example, the GraalVM website also documents TruffleRuby since it's one of its components:
https://www.graalvm.org/docs/reference-manual/languages/ruby/
https://www.graalvm.org/docs/Ruby/user/README

hsbt:
> it's about consistency in the Ruby ecosystem.

Are you maintainer of the Ruby ecosystem? Why did you decide it without other opinions? 

You sometimes enforce to me with the truffleruby matters. I'm tired that. (It's second time for me)

eregon:
> Are you maintainer of the Ruby ecosystem?

No. But I am the maintainer of TruffleRuby in the 3 main Ruby installers (RVM, ruby-build, ruby-install).
I think anyone can see value in some consistency between them for installing the same thing.

> Why did you decide it without other opinions?

I actually asked a few people, notably on the TruffleRuby Slack. There seems to be no good alternative.
* `graalvm` would work.
* `truffleruby-graalvm` technically doesn't work or requires major hacks.
* `truffleruby+graalvm` could be one way, but it's unprecedented in ruby-build, it feels like 2 things but TruffleRuby is part of GraalVM (discussed above), paths with `+` in the PATH looks IMHO a bit weird (and `shellescape` escapes it to `\+` which is ugly). 

> You sometimes enforce to me with the truffleruby matters.

I'm not sure to understand what you mean.
I'm sorry about it in any case.
Could you clarify? What's the second time you are thinking of?

If you don't want to review this PR, that is fine by me.
I asked your review as I think it's nice to have a review on this and you are AFAIK the only active maintainer of `ruby-build`.
But I have commit rights here, and this is TruffleRuby-related, so I feel it's also fair enough if I merge without needing your approval.

Also, I feel sometimes you want to enforce your opinion, like here with the name, even though you did not explain why you want this name. I explained in details why I think `graalvm` is better. I'm happy to discuss about it, but if you do not want to elaborate on your opinion, then I think it's not fair to decide the name on your own.

hsbt:
> But I am the maintainer of TruffleRuby in the 3 main Ruby installers (RVM, ruby-build, ruby-install).
> I actually asked a few people, notably on the TruffleRuby Slack.

I always feel Your "Ruby" words points only TrfuffleRuby. I wonder why you ask the ecosystem issue to Ruby(CRuby) and JRuby people. I was surprised your "Ruby" proposal relied on TruffleRuby.

> Could you clarify? What's the second time you are thinking of?

see https://github.com/rbenv/ruby-build/pull/1398#issuecomment-574998236 and DM of Ruby's slack.

Back to the naming issue.

> But I have commit rights here, and this is TruffleRuby-related, so I feel it's also fair enough if I merge without needing your approval.

The global naming issue is not only TruffleRuby-related issue. The people try to install Ruby with ruby-build at first. They think the interpriter or implementation name, not VM. The this isntallation triggered truffleruby, We should use "truffuleruby" words for consistency of the ruby-build policy. If graalvm will trigger Ruby(CRuby), I accept to use `graalvm` name.

My thought is here.

* CRuby related definitions can use no-prefix definition
* Other implementation like JRuby, Rubinius, and TruffleRuby needs to its prefix by historical reason and convention.

`graalvm` broke above convention.

eregon:
To some degree I feel each project should be able to decide about the name of the project.
Of course, it also needs to fit with `ruby-build` naming as it integrates with this project.

I see `graalvm-` as a prefix, just like `truffleruby-` is a prefix. This PR is not adding a `20.1.0` definition but a `graalvm-20.1.0` definition.
I see your point though, people not familiar with Ruby will probably not guess so easily what a `graalvm` formula is, and that `graalvm` is essentially a variant of the `truffleruby` definition with more stuff.
I'll try `truffleruby+graalvm` then.

eregon:
`truffleruby+graalvm` seems to work and needed only [minor tweaks](https://github.com/rbenv/ruby-build/pull/1455/files#diff-fdcfb8a18714b33b07529b7d02b54f1dR1259) to the latest & sorting versions logic.

PR #1457, opened 2020-07-17
Support specifying the complete mirror URL

### LINKED ISSUE #1032: Supporting alternative mirror url schemes
At present, ruby-build only supports the `mirrorUrl/checksum` format for mirrored package URLs. (https://github.com/rbenv/ruby-build/blob/ffef60696af59a73f935459fd128adb8bfd369a7/bin/ruby-build#L356) For node-build, we had need to support a different default mirrored URL scheme. In doing so, we have extended the mirror support to not only accept a *_MIRROR_URL environment variable, but also to accept a *_MIRROR_CMD variable with the name of a function or command which can _construct_ a full mirrored package URL (given the original package URL and checksum as arguments). In this manner, node-build can default to a common mirror scheme for node mirrors, but also be flexible enough to allow users to construct URLs as necessary for their specific mirror sites.

The core of this feature is in https://github.com/nodenv/node-build/pull/210/commits/c73fbe20bc7459a928291e85e1d30c2c28e3787b as part of https://github.com/nodenv/node-build/pull/210

We haven't yet merged this feature into node-build. If everything goes as planned and it _is_ merged to node-build, I intend to also submit this feature here to ruby-build. In hopes of gathering feedback ahead of time, I would love some comments on this feature.

uzxmx:
When using a mirror, `ruby-build` requires the package is located at `$RUBY_BUILD_MIRROR_URL/<SHA2>`, which is not a generic case.

[Here](https://www.ruby-lang.org/en/downloads/mirrors/) list all mirror sites of ruby source, and most of them provide the tarball with name format like `http://host/pub/ruby/2.7/2.7.1.tar.bz2`. To benefit from more mirror sites, and also keep compatibility, I propose adding another variable (which can be named as `RUBY_BUILD_MIRROR_PACKAGE_URL`) to specify the complete mirror URL. So a user can use it like this:

```
# I use `asdf` to manage versions, which depends on `ruby-build`.
$ RUBY_BUILD_MIRROR_PACKAGE_URL=https://cache.ruby-china.com/pub/ruby/2.7/ruby-2.7.1.tar.bz2 asdf install ruby 2.7.1
```

Fix #1032. As [this comment](https://github.com/rbenv/ruby-build/issues/1032#issuecomment-266220709) said, the download speed in China is too slow when installing a new version from official ruby CDN, so at least this feature will be especially useful for users living in China.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,28 +64,29 @@ definitions.
 
 The build process may be configured through the following environment variables:
 
-| Variable                 | Function                                                                                         |
-| ------------------------ | ------------------------------------------------------------------------------------------------ |
-| `TMPDIR`                 | Where temporary files are stored.                                                                |
-| `RUBY_BUILD_BUILD_PATH`  | Where sources are downloaded and built. (Default: a timestamped subdirectory of `TMPDIR`)        |
-| `RUBY_BUILD_CACHE_PATH`  | Where to cache downloaded package files. (Default: `~/.rbenv/cache` if invoked as rbenv plugin)  |
-| `RUBY_BUILD_HTTP_CLIENT` | One of `aria2c`, `curl`, or `wget` to use for downloading. (Default: first one found in PATH)    |
-| `RUBY_BUILD_ARIA2_OPTS`  | Additional options to pass to `aria2c` for downloading.                                          |
-| `RUBY_BUILD_CURL_OPTS`   | Additional options to pass to `curl` for downloading.                                            |
-| `RUBY_BUILD_WGET_OPTS`   | Additional options to pass to `wget` for downloading.                                            |
-| `RUBY_BUILD_MIRROR_URL`  | Custom mirror URL root.                                                                          |
-| `RUBY_BUILD_SKIP_MIRROR` | Bypass the download mirror and fetch all package files from their original URLs.                  |
-| `RUBY_BUILD_ROOT`        | Custom build definition directory. (Default: `share/ruby-build`)                                 |
-| `RUBY_BUILD_DEFINITIONS` | Additional paths to search for build definitions. (Colon-separated list)                         |
-| `CC`                     | Path to the C compiler.                                                                          |
-| `RUBY_CFLAGS`            | Additional `CFLAGS` options (_e.g.,_ to override `-O3`).                                         |
-| `CONFIGURE_OPTS`         | Additional `./configure` options.                                                                |
-| `MAKE`                   | Custom `make` command (_e.g.,_ `gmake`).                                                         |
-| `MAKE_OPTS` / `MAKEOPTS` | Additional `make` options.                                                                       |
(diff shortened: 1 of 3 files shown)


jasonkarns:
Fwiw, node-build had this same limitation and we solved it by adding a new env var "NODE-BUILD-MIRROR-CMD" which defines a user-provided command that will generate a mirror url when given the package name and checksum.

https://github.com/nodenv/node-build/blob/3c0318da0dd975420653d1cf8e14578a8709bd3b/bin/node-build#L473

This allows users full control over how mirror urls are constructed. It may be a feature worth up streaming into ruby-build.

uzxmx:
@jasonkarns I considered the `*_MIRROR_CMD` mechanism you mentioned in #1032. I agree `*_MIRROR_CMD` will be the best way to allows users full control over how mirror urls are constructed. But I have two concerns:

* Less straightforward than a simple variable that just specifies the final url. To use `*_MIRROR_CMD`, it needs either we export a function, or we write a simple script to print the final url.
* I'm a bit worried that some kinds of user mistakes may lead to bad things. Because it just executes whatever a user feeds.

So I think a simple variable is enough to satisfy all the cases.

Anyway, if you or other members prefer `*_MIRROR_CMD` way, I also vote for it.

jasonkarns:
> * Less straightforward than a simple variable that just specifies the final url. To use `*_MIRROR_CMD`, it needs either we export a function, or we write a simple script to print the final url.

Yep, I agree it's not super simple. The downside with a variable containing the full URL is that users would need to construct that URL for each ruby they install. The _CMD approach allows a one-time setup that works for future ruby installs as well. I'd say the using a full URL env var would be easy, but not necessarily "simple".

It's also worth noting that the _CMD approach and the "full URL" approach are not mutually exclusive. We could support both so users have a choice for whichever suits their needs. I'd like some other maintainers to weigh in before proceeding with a PR.

uzxmx:
> users would need to construct that URL for each ruby they install. The _CMD approach allows a one-time setup that works for future ruby installs as well.

Totally agree.

> 
> It's also worth noting that the _CMD approach and the "full URL" approach are not mutually exclusive. We could support both so users have a choice for whichever suits their needs. I'd like some other maintainers to weigh in before proceeding with a PR.

Thanks. Waiting for your and other maintainers' further feedbacks.

uzxmx:
@hsbt Sorry to ping you. Can you help to make progress on this issue? Really want to have another strategy to specify the mirror.

@jasonkarns I found another [fix](https://github.com/pyenv/pyenv/pull/1673) for this similar issue in python-build, which was merged several days ago. It uses `PYTHON_BUILD_MIRROR_URL_SKIP_CHECKSUM` to "skip" the checksum.

hsbt:
@uzxmx Sorry, too late response. I knew the situation of Ruby in China. I believe this option helps Ruby users in China.

PR #1459, opened 2020-07-26
Fix the installer of truffleruby+graalvm with macOS

hsbt:
@eregon 

I try to install truffleruby+graalvm with macOS. But it always failed because the macOS package is different from linux version.

macOS package is:

```
Contents
Contants/Home <- This path is root of graalvm
Contents/Info.plist
Contents/MacOS
(snip)
THIRD_PARTY_README_JDK
```

I resolved the directory structure with linux package on macOS. I could install `truffleruby+graalvm-20.1.0` with this patch.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -756,6 +756,12 @@ build_package_truffleruby_graalvm() {
   build_package_copy_to "${PREFIX_PATH}/graalvm"
 
   cd "${PREFIX_PATH}/graalvm"
+
+  if is_mac; then
+    mv Contents/Home/* .
+    rm -rf Contents 
+  fi
+
   bin/gu install ruby || return $?
 
   local ruby_home


eregon:
Thanks for testing!

I forgot about this strange layout on macOS :/

How about https://github.com/rbenv/ruby-build/pull/1460 ?
Could you test that on macOS? (I can also ask someone else to test if you're busy)
The advantage is it doesn't move or remove files, instead adjusting where to link `bin/` to.

hsbt:
#1460 is works with macOS 10.15.

I think `Home` and `MacOS` layout is specific for `.app` package with macOS. It doesn't need to use a slandalone binaries like truffleruby. But It's not strong opinion to remove them. I'm okay to #1460.

eregon:
> I think `Home` and `MacOS` layout is specific for `.app` package with macOS. It doesn't need to use a slandalone binaries like truffleruby. But It's not strong opinion to remove them. I'm okay to #1460.

Yeah I know it looks weird. But this is the standard layout for all JDKs on macOS.
OK, let's merge #1460, thank you for testing and investigating.

hsbt:
👍

PR #1460, opened 2020-07-26
Fix installation of truffleruby+graalvm on macOS

eregon:
See #1459.

This way avoids moving and removing files.


eregon:
@ssnickolay checked and it works.

PR #1461, opened 2020-07-28
Rename CONDUCT.md to CODE_OF_CONDUCT.md

jasonkarns:
GitHub's expected filename is `CODE_OF_CONDUCT.md`. Whether that is the best name or even the current de facto standard is to be debated. But moving forward, this is likely the filename that will become the most common within the community, as that is the name GitHub expects and suggests.

This simply renames the existing code of conduct file from `CONDUCT.md` to `CODE_OF_CONDUCT.md` to conform to GitHub's preference.


pjkendall982:
Until all bugs are fixed and all problems are worked out, when it comes to something as simple as a name, in my opinion it shouldn’t be written in stone till you are ready for your next project...... The future always leaves room for improvement.
All respect
Pat

PR #1465, opened 2020-08-10
Added support of MSYS2 environment

knovok:


mislav:
Thank you, but ruby-build is unsupported on Windows, and we don't have the resources to maintain MSYS2 support. I suspect the only place that ruby-build will run on Windows is under WSL.

PR #1474, opened 2020-09-24
[easy] promote 'rbenv install -L', an alias of '--list-all'

metalefty:
As far as I search on Twitter, the short option is not well known.


hsbt:
Thanks

PR #1476, opened 2020-09-28
Local openssl for old rubies

makandra-stefan-langenmaier:
Hi!

We have problems with missing openssl-1.0 for older Ruby versions in newer distribution and I saw multiple other issues here with a similar topic.
As I saw, there are alreay local openssl installs used for Macs, I thought this could also be used for older Ruby versions.

Let me know if this could be an approach for Linux without the correct openssl installed. Feedback is welcome.

Best regards,
Stefan

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -101,6 +101,10 @@ os_information() {
   fi
 }
 
+is_linux() {
+  [ "$(uname -s)" = "Linux" ] || return 1
+}
+
 is_mac() {
   [ "$(uname -s)" = "Darwin" ] || return 1
   [ $# -eq 0 ] || [ "$(osx_version)" "$@" ]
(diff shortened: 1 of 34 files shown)


excid3:
I would love this. Would make compiling older rubies on Ubuntu a lot easier.

excid3:
Tested this and it works great. 👍 

I did notice that Ruby 2.0.0-p648 is missing the openssl line for Linux and should probably be included. Ideally, all the older ones too, but I'm okay if it's just the latest versions of the old Rubies.

jeremy:
Nice work!

tibra:
We absolutely need this!

eregon:
FWIW, I have a similar branch at https://github.com/rbenv/ruby-build/compare/master...eregon:ruby23-openssl-linux.
I think it makes sense to compile OpenSSL 1.0.0 automatically on Linux, rather than asking everyone to do this manually:
https://github.com/rbenv/ruby-build/wiki#openssl-version-compatibility

Note that a previous PR doing a similar approach ended up being rejected: https://github.com/rbenv/ruby-build/pull/1397#pullrequestreview-341523509
So I guess we need to convince the other maintainers (cc @mislav  @hsbt).

I think the main thing that changed since then is all recent Linux distributions don't have a way to install OpenSSL 1.0.0 anymore, so it's very inconvenient to manually download and compile OpenSSL 1.0.0.
And this logic already exists for macOS anyway.

So I think what we should do is just extend the logic to deal with the slightly different certificate paths on Linux, as done in https://github.com/rbenv/ruby-build/compare/master...eregon:ruby23-openssl-linux, and then rename
`mac_openssl --if has_broken_mac_openssl` to
`openssl10 --if has_wrong_openssl`.
Duplicating the lines seems suboptimal to me at least.

makandra-stefan-langenmaier:
I'm closing that pull request as the solution caused more problems for us in the long-term and I wanted to reduce possible confusion for others. The issues were not only with Ruby but also other gems that use openssl.

eregon:
> The issues were not only with Ruby but also other gems that use openssl.

Is it the issue that some gems like database gems (mysql2, pg) link to a system library (e.g., installed by a package manager) and those libraries link against the system libssl, and then the Ruby process ends up loading 2 versions of libssl, which usually segfaults?

PR #1712, opened 2021-01-19
Run Linux CI and script/mirror verify in GitHub Actions

eregon:
* TravisCI is super slow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,10 +1,14 @@
-name: test with macos
+name: Test
 
 on: [push, pull_request]
 
 jobs:
-  build:
-    runs-on: macos-latest
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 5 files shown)


eregon:
Mmh, probably we need to migrate https://github.com/rbenv/ruby-build/blob/28bcde943de838603a47b90c2db3bf7305ff41df/script/test to commands inside the workflow.

Also we'll need to make sure the env vars and secrets that are set in [.travis.yml](https://github.com/rbenv/ruby-build/blob/28bcde943de838603a47b90c2db3bf7305ff41df/.travis.yml) are also set for GitHub Actions.

hsbt:
I'm positive to migrate this. But I didn't know a secrets of AWS credentials for the current CDN.

@mislav Can you setup it to GitHub Secrets?

eregon:
This PR now keeps TravisCI to upload to S3 (since only TravisCI has the secrets), and all tests are moved to GitHub Actions.

Issue #1730, opened 2021-02-24
JRuby 9.2.15.0 was released


danielburkat:
Please add jruby-9.2.15.0 to the list of available installations. 

https://www.jruby.org/2021/02/24/jruby-9-2-15-0.html

rolandoam:
@hsbt the merged pull request placed the jruby definition in the wrong directory. I just created a new PR to fix this

PR #1731, opened 2021-02-24
adds jruby 9.2.15.0

rolandoam:
JRuby 9.2.15.0 was released today (https://www.jruby.org/2021/02/24/jruby-9-2-15-0.html). I believe this is the file that would make it available for ruby-build env.


rolandoam:
should solve #1730

PR #1741, opened 2021-03-30
Add JRuby 9.2.17.0 definition

andrepiske:
It was released yesterday: https://www.jruby.org/2021/03/29/jruby-9-2-17-0.html


hsbt:
Thanks!

PR #1758, opened 2021-04-22
Join *_OPTS variables after *_OPTS_ARRAY variables

casperisfine:
Ref: https://github.com/rbenv/ruby-build/pull/1368#issuecomment-824624767

This for instance allows `RUBY_CONFIGURE_OPTS` to override
what's set by the definition.

For instance `enable_shared` can be disabled with `RUBY_CONFIGURE_OPTS='--disable-shared'`

cc @eregon 

NB: I tried to write a test for it, but I'm getting failures without much info to debug:
```
 ✗ ordering
   (from function `unstub' in file test/test_helper.bash, line 62,
    in test file test/build.bats, line 265)
     `unstub make' failed
```

Pointers on how best to do this are welcome.


casperisfine:
I manually tested this with:

```
$ RUBY_CONFIGURE_OPTS=" --disable-shared " bin/ruby-build 3.0.0 /tmp/rubies/3.0.0
$ /tmp/rubies/3.0.0/bin/ruby -r rbconfig -e 'p RbConfig::CONFIG["ENABLE_SHARED"]'
"no"
$ /tmp/rubies/3.0.0/bin/ruby -r rbconfig -e 'p RbConfig::CONFIG["configure_args"]'
" '--prefix=/tmp/rubies/3.0.0' '--with-openssl-dir=/tmp/rubies/3.0.0/openssl' '--enable-shared' '--with-readline-dir=/usr/local/opt/readline' '--disable-shared' 'CC=clang' 'CFLAGS= -DHAVE_UNISTD_H ' 'LDFLAGS=-L/tmp/rubies/3.0.0/lib -L/usr/local/opt/openssl/lib' 'CPPFLAGS=-I/tmp/rubies/3.0.0/include -I/usr/local/opt/openssl/include'"
```

bin/ruby-build
```diff
@@ -589,10 +589,10 @@ build_package_standard_build() {
       export CC=clang
     fi
     ${!PACKAGE_CONFIGURE:-./configure} --prefix="${!PACKAGE_PREFIX_PATH:-$PREFIX_PATH}" \
-      $CONFIGURE_OPTS ${!PACKAGE_CONFIGURE_OPTS} "${!PACKAGE_CONFIGURE_OPTS_ARRAY}" || return 1
+      "${!PACKAGE_CONFIGURE_OPTS_ARRAY}" $CONFIGURE_OPTS ${!PACKAGE_CONFIGURE_OPTS} || return 1
   ) >&4 2>&1
 
-  { "$MAKE" $MAKE_OPTS ${!PACKAGE_MAKE_OPTS} "${!PACKAGE_MAKE_OPTS_ARRAY}"
+  { "$MAKE""${!PACKAGE_MAKE_OPTS_ARRAY}" $MAKE_OPTS ${!PACKAGE_MAKE_OPTS}
```
eregon, line 595: Isn't there a missing space here?

casperisfine: Indeed. I pushed a fix.

eregon:
Is it the case that `*_OPTS_ARRAY` is only modified by `package_option`, and that's only called by some functions in `ruby-build` with values which are not controlled by the user (either hardcoded or from some system command), and is therefore never influenced by user-provided values?
As far as I could see, it seems the case, but I'm not super familiar with that logic.

If that holds then I guess it might be OK to change that order as in this PR. If not, it could cause some incompatibility.
Even if it holds, I guess it might cause some incompatibility as flags are passed in a different order, but OTOH it seems best that user-provided flags can override the defaults of `ruby-build` :+1: 

> ... '--enable-shared' '--with-readline-dir=/usr/local/opt/readline' '--disable-shared' ...

That feels a bit suboptimal, maybe `build_package_enable_shared` could check if `--disable-shared` was passed and then not add `--enable-shared` at all?
It would look a bit cleaner and if anything ever uses something like `RbConfig::CONFIG["configure_args"].include?("--enable-shared")` (a bad idea, there is `RbConfig::CONFIG["ENABLE_SHARED"]` for that) it would not break (or if the configure script does some changes on seeing --enable-shared that it doesn't under with --disable-shared, etc).
Might be a bit magic, but I think it would be a good trade-off in this case.

casperisfine:
>  is therefore never influenced by user-provided values?

As far as I can tell yes, but I might have missed something?

> That feels a bit suboptimal, maybe build_package_enable_shared could check if --disable-shared was passed and then not add --enable-shared at all?

I thought about that, but I thought it would be cleaner to have a more general fix rather than to focus on `--enable-shared` specifically.

But I'm open to implement that alternative if you think it's better.

eregon:
@hsbt @mislav It would be great if one of you can review too.

This looks fine to me, as it enables the user to override ruby-build defaults, which seems useful/expected in general (CLI arguments or in this case user-provided env vars typically override defaults in a script).

I think avoiding to add `--enable-shared` if `--disable-shared` is in `RUBY_CONFIGURE_OPTS` would be a bit cleaner for the result.

@casperisfine 
> I thought about that, but I thought it would be cleaner to have a more general fix rather than to focus on --enable-shared specifically.

I think both would be best, what you did here, and removing the extra `--enable-shared` (keeping the `--disable-shared` from `RUBY_CONFIGURE_OPTS`).

eregon:
> NB: I tried to write a test for it, but I'm getting failures without much info to debug:

Could you show the partial test you wrote/push it to a branch? It'd be nice to test this change to avoid regressions.

casperisfine:
Of course:

Here's one of the version https://github.com/casperisfine/ruby-build/commit/2f5ae705b844d37c3f8f17a3c60be00c81c719e0

I know it can't work because `--disable-shared` was introduced much later than 2.0, but if I try to make a inline definition with say 2.5.x, it fail to download the archive...

casperisfine:
> I think avoiding to add `--enable-shared` if `--disable-shared` is in `RUBY_CONFIGURE_OPTS` would be a bit cleaner for the result.

I added a commit that does just that.

eregon (approved):
This looks good to me.
There wasn't much feedback from other maintainers, I'll assume they are busy.
Let's merge this, and if there is any issue it can most likely be handled some way (example: the first commit could be reverted if it breaks something).

casperisfine:
Thank you!

mislav:
I think this changeset is risky, since it changes the order of build flags passed in. This might affect someone's builds in ways that we did not predict.

But now that it's released, let's wait to see whether we get any complaints!

PR #1759, opened 2021-04-23
Fix hashes for TruffleRuby GraalVM 21.1.0

eregon:
* Unfortunately the files were changed and reuploaded as there was an
  issue with the gu (GraalVM Component Updater) catalog URL.


eregon:
The cloudfront mirror seems to use the hash, so that should work fine.

Issue #42, opened 2011-09-17
Binary packages


sstephenson:
I'd like to add binary package support to ruby-build. I'd guess that most ruby-build installations are on one of two platforms—Mac OS X x86_64 or Linux x86_64—and most people really don't need to compile their own Ruby runtimes. Binary packages would make it much quicker to get up and running.

In order to do this, we'll need a command to build a definition into a temporary directory and make a tarball. We'll also need a way to download and install package tarballs from the internet. This implies having package repository sources, an architecture naming system, and an index of packages with checksums.

schlueter:
Any movement on this?

md5:
@schlueter +!

cybertk:
+1

mconigliaro:
To be honest, I don't even care about getting *your* binaries to work on *my* machines. If there was just a simple way to check for a precompiled version of Ruby before building (e.g. `RUBY_BUILD_BINARY_URL`?), I'd just build my own and host them myself.

md5:
@mconigliaro If you want to do that, you can use `rbenv` by itself and not use `ruby-build` since its purpose is to build Ruby for `rbenv`.

From [the `rbenv` docs](https://github.com/rbenv/rbenv#installing-ruby-versions):

> Alternatively to the `install` command, you can download and compile Ruby manually as a subdirectory of `~/.rbenv/versions/`. An entry in that directory can also be a symlink to a Ruby version installed elsewhere on the filesystem. `rbenv` doesn't care; it will simply treat any entry in the versions/ directory as a separate Ruby version.

mconigliaro:
@md5 oh yeah, that's a good point. I could always just have my config management system download and unpack a set of precompiled versions to `~/.rbenv/versions/`. Thanks!

jasonkarns:
Super old issue, but I just wanted to pop in and say that node-build (direct fork, minimally modified) has added support for installing precompiled binaries. https://github.com/nodenv/node-build/commit/474e872d9acd5cc61c909f58d1521626d8c9d385 (source has been modified slightly since that PR was merged, though)

We grab the system info via:
```
platform(){
  local arch="$(uname -m)"
  local os="$(uname -s | tr '[:upper:]' '[:lower:]')"

  case "$arch" in
    x86_64 | amd64 | i686-64 ) arch=x64 ;;
    i[36]86* | [ix]86pc )      arch=x86 ;;
  esac

  echo "${os}-${arch}"
}
```

And a build definition looks like:
[...]

PR #1760, opened 2021-04-23
Initial attempt to add support for movable rubies

technicalpickles:
This follows the pattern that rvm uses. There's a --movable flag, and that adds some extra options.

Only tested on Ubuntu so far, but I can confirm that the configure options end up in RbConfig, but have not tested beyond that yet.

cc https://github.com/rbenv/ruby-build/issues/42


eregon:
FYI, `--enable-load-relative` has several issues:
https://github.com/ruby/setup-ruby/issues/98#issuecomment-721233467

technicalpickles:
I think those issues are manageable to some extent. That comment references 2.3 and 2.1, both of which are EOL. And also truffle ruby. I think then is it should be possible to not allow movable in those cases.

eregon:
Note the shebang issue affects all versions, i.e. `gem install bundler` fails with `--enable-load-relative` on all Ruby versions. That very likely needs to be fixed or worked around.
Also I feel nobody really tests that configuration and if it fails it'll be ruby-build's "fault/responsibility" for using an unusual option.

Regarding truffleruby we already run the post-install hook and it already installs from binaries, so there is nothing to do.
To me this is yet another clear sign CRuby should make their own binary movable builds, maybe that might happen if people open an issue on https://bugs.ruby-lang.org/ and it gains some support (e.g., nodejs has this).

PR #1405, opened 2020-02-13
Add truffleruby-dev to install the latest TruffleRuby nightly build

eregon:
* Uses the latest nightly, which is automatically built.
* I checked locally and it works well with:
  `ruby-build truffleruby-dev ~/.rubies/truffleruby-dev`.

cc @hsbt @deepj 

Ref: https://github.com/oracle/truffleruby/issues/1483

PR #1773, opened 2021-06-11
Add jruby-dev

deivid-rodriguez:
* Uses the latest nightly, which is automatically built.

This is a shameless copy of https://github.com/rbenv/ruby-build/pull/1405.

### DIFF
--- a/share/ruby-build/jruby-dev
+++ b/share/ruby-build/jruby-dev
@@ -1,0 +1,13 @@
+case $(uname -s) in
+Linux)
+  install_package "jruby-head" "https://github.com/ruby/jruby-dev-builder/releases/latest/download/jruby-head-ubuntu-20.04.tar.gz" jruby
+  ;;
+Darwin)
+  use_homebrew_openssl
+  install_package "jruby-head" "https://github.com/ruby/jruby-dev-builder/releases/latest/download/jruby-head-macos-latest.tar.gz" jruby
+  ;;
+*)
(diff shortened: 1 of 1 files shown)


share/ruby-build/jruby-dev
```diff
@@ -0,0 +1,13 @@
+case $(uname -s) in
+Linux)
+  install_package "jruby-head" "https://github.com/ruby/jruby-dev-builder/releases/latest/download/jruby-head-ubuntu-20.04.tar.gz" jruby
+  ;;
+Darwin)
+  use_homebrew_openssl
+  install_package "jruby-head" "https://github.com/ruby/jruby-dev-builder/releases/latest/download/jruby-head-macos-latest.tar.gz" jruby
```
eregon, line 7: The archive is the same on Linux and macOS, but there might be differences in the future so this sounds good.
Those URLs are not guaranteed to be stable forever, but they will probably be relatively stable for ruby/setup-ruby, and we can always update them.

eregon (approved):
Thanks!

eregon:
This also fixes https://github.com/jruby/jruby/issues/6092, cc @PragTob

cc @headius (to let you know about this)

share/ruby-build/jruby-dev
eregon: Self-note, discussion about platform-specific archives or not: https://github.com/jruby/jruby/issues/6092#issuecomment-962623179

PR #1780, opened 2021-07-15
Refresh download mirror in GitHub Actions on push

### LINKED ISSUE #1778: travis-ci.org no longer builds and Ruby archives are no longer mirrored
See https://travis-ci.org/github/rbenv/ruby-build/builds, which says:

> Since June 15th, 2021, the building on travis-ci.org is ceased. Please use travis-ci.com from now on.

With https://github.com/rbenv/ruby-build/pull/1712, the only thing .travis.yml would do is to upload to the S3 mirror.
So recent Rubies are not mirrored anymore.

There are two options:
* Use travis-ci.com instead of travis-ci.org
* Move the secrets to upload to S3 to GitHub Actions, and let GitHub Actions handle the upload, cc @mislav.

mislav:
Fixes #1778

Ref. https://github.com/rbenv/ruby-build/pull/1712
Ref. #438


.github/workflows/mirror.yml
```diff
@@ -0,0 +1,18 @@
+name: Mirror
+
+on: push
```
eregon, line 3: Should it be limited to the `master` branch?

.github/workflows/mirror.yml
```diff
@@ -0,0 +1,18 @@
...
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Update download mirror
+      run: script/mirror update "${BEFORE_REF}.."
```
eregon, line 13: It would be interesting to run it once with `6b5c8148228bc3f5e58ab9c54c29d0b71be65524..` (the last time TravisCI ran).
That would both test the upload, and upload new releases added since that commit.

script/mirror
```diff
@@ -53,31 +53,38 @@ update() {
...
+      echo "Will mirror: $url"
+      [ -n "$tmp_path" ] || tmp_path="$(mktemp -d "${TMPDIR:-/tmp}/s3-sync.XXXXX")"
+      file="$tmp_path/$checksum"
       download_and_verify "$url" "$file" "$checksum"
-      ./script/s3-put "$file" "${AMAZON_S3_BUCKET?}"
     fi
   done
+  if [ -n "$tmp_path" ]; then
+    echo "Uploading..."
+    aws s3 sync --acl=public-read --size-only "$tmp_path" "s3://${AMAZON_S3_BUCKET?}"
```
eregon, line 71: What provides the `aws` command here? Is it available by default in CI?

.github/workflows/mirror.yml
mislav: You're right! I'll try to trigger the action with that ref. However, I have already pre-filled the mirror while testing on my own machine, so the output we should see is that all downloads have already been mirrored.

script/mirror
mislav: Yes it comes preinstalled on Actions runners

.github/workflows/mirror.yml
mislav: That's a good thought! I'll limit it to the default branch. 👍

mislav:
Okay I've enabled manually dispatching the mirror update job for testing:
```
gh workflow run mirror.yml -r actions-mirror -f beforeRef=6b5c8148228bc3f5e58ab9c54c29d0b71be65524
```
https://github.com/rbenv/ruby-build/runs/3087601268?check_suite_focus=true

eregon (approved):
Looks great, thanks!

hsbt:
🎉

PR #1793, opened 2021-08-10
Feature: more flexible mirror URL by a command

uzxmx:
This PR provides a more flexible way to specify the mirror URL by using a shell function than #1457.

For #1457, users need to know the complete mirror URL, and another drawback is users need to change the `RUBY_BUILD_MIRROR_PACKAGE_URL` env every time they install a different version.

This PR resolves these two drawbacks. Users can build the final mirror URL by any shell commands.

Thanks to the author of #1032, the idea comes from that thread.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,7 @@ The build process may be configured through the following environment variables:
 | `RUBY_BUILD_WGET_OPTS`          | Additional options to pass to `wget` for downloading.                                            |
 | `RUBY_BUILD_MIRROR_URL`         | Custom mirror URL root.                                                                          |
 | `RUBY_BUILD_MIRROR_PACKAGE_URL` | Custom complete mirror URL (e.g. http://mirror.example.com/package-1.0.0.tar.gz).                |
+| `RUBY_BUILD_MIRROR_CMD`         | Custom mirror command (see [below](#more-flexible-mirror-url) for the usage).                    |
 | `RUBY_BUILD_SKIP_MIRROR`        | Bypass the download mirror and fetch all package files from their original URLs.                 |
 | `RUBY_BUILD_TARBALL_OVERRIDE`   | Override the URL to fetch the ruby tarball from, optionally followed by `#checksum`.             |
 | `RUBY_BUILD_DEFINITIONS`        | Colon-separated list of paths to search for build definition files.                              |
@@ -149,6 +150,25 @@ complete URL by setting `RUBY_BUILD_MIRROR_PACKAGE_URL`. It behaves the same as
 
 The default ruby-build download mirror is sponsored by
 [Basecamp](https://basecamp.com/).
+
+#### More flexible mirror URL
+
+For more flexible mirror URL, you can provide a custom command to output the
+desired mirror URL, as shown below:
+
+```sh
+# There are two arguments:
+#   1st arg: the original URL without checksum
+#   2nd arg: the checksum
+$ cat <<'EOF' >./get_mirror_url && chmod +x ./get_mirror_url
+#!/bin/sh
+echo "$1" | sed "s/cache.ruby-lang.org/mirror.example.com/"
+EOF
+
+$ export RUBY_BUILD_MIRROR_CMD="$(pwd)/get_mirror_url"
+```
+
+After executing the above script in your shell, install a version as usual.
 
 #### Keeping the build directory after installation
 
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -474,6 +474,8 @@ fetch_tarball() {
       fi
     elif [ -n "$RUBY_BUILD_MIRROR_PACKAGE_URL" ]; then
       mirror_url="$RUBY_BUILD_MIRROR_PACKAGE_URL"
+    elif [ -n "$RUBY_BUILD_MIRROR_CMD" ]; then
+      mirror_url="$("$RUBY_BUILD_MIRROR_CMD" "$package_url" "$checksum")"
     fi
   fi
 
@@ -1537,7 +1539,7 @@ else
   unset RUBY_BUILD_CACHE_PATH
 fi
 
-if [ -z "$RUBY_BUILD_MIRROR_URL" ] && [ -z "$RUBY_BUILD_MIRROR_PACKAGE_URL" ]; then
+if [ -z "$RUBY_BUILD_MIRROR_URL" -a -z "$RUBY_BUILD_MIRROR_PACKAGE_URL" -a -z "$RUBY_BUILD_MIRROR_CMD" ]; then
   RUBY_BUILD_MIRROR_URL="https://dqw8nmjcqpjn7.cloudfront.net"
   RUBY_BUILD_DEFAULT_MIRROR=1
 else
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -142,6 +143,25 @@ complete URL by setting `RUBY_BUILD_MIRROR_PACKAGE_URL`. It behaves the same as
...
+```sh
+# There are two arguments:
+#   1st arg: the original URL without checksum
+#   2nd arg: the checksum
+mirror_cmd() {
+  echo "${1/cache.ruby-lang.org/mirror.example.com}"
+}
+
+export -f mirror_cmd
+export RUBY_BUILD_MIRROR_CMD=mirror_cmd
```
eregon, line 160: This seems quite fragile and I guess only works if the original shell is Bash too.
Isn't RUBY_BUILD_MIRROR_URL enough?

uzxmx: Yes, the example only works for bash. I'll update the example by using a more compatible way.

eregon:
Thanks, that variant avoids relying on exporting shell functions which seems far more portable.

I feel like an env var to set the mirror base URL could be simpler, but one issue is e.g.
```
https://cache.ruby-lang.org/pub/ruby/3.0/ruby-3.0.0.tar.gz
is mirrored as
https://mirror.cyberbits.eu/ruby/3.0/ruby-3.0.0.tar.gz
https://cache.ruby-china.com/pub/ruby/3.0/ruby-3.0.0.tar.gz
```
[CRuby mirror list](https://www.ruby-lang.org/en/downloads/mirrors/)
BTW many of these mirrors don't have recent releases, I wonder which are actually used.
The `China 2 (Ruby China)`  mirror at least has the same structure as `cache.ruby-lang.org/` so for those we could just reuse the URL path and just change the domain.

@hsbt What do you think?

uzxmx:
IMO, just changing the domain is definitely the simplest solution. But so far, `ruby-build` actually doesn't support  this way, some codes need to be added. For example, for `RUBY_BUILD_MIRROR_URL`, the final mirror URL is built by concatenating `RUBY_BUILD_MIRROR_URL` with the checksum, so we cannot rely on it.

https://github.com/rbenv/ruby-build/blob/57c397d7cabf9f0ff73b8c03c4c1859a3850405a/bin/ruby-build#L364-L371

Just changing the domain may also not work for other ruby variants, like jruby (e.g. one url is `https://s3.amazonaws.com/jruby.org/downloads/1.7.22/jruby-bin-1.7.22.tar.gz#554da042087bd4a787c73626c81fa354c9ce1168735032f7d954cffec85f5a4a`).

jasonkarns:
This looks mostly compatible with the implementation that node-build uses from nodenv/node-build#210

As @uzxmx mentioned different mirror urls, it's the same reason node-build opted to accept the `_CMD` env var: to be flexible enough to handle any mirror and not be limited to those that preserve identical path formats, or that accept the checksum in different ways.

mislav (changes requested):
Thank you for your contributions to ruby-build and helping make it more usable in China!

As a late reviewer to this whole extended mirror feature, I would like to voice my concerns about the complexity of the feature (incl. the original PR that already got merged).

First of all, I must admit I do not understand RUBY_BUILD_MIRROR_PACKAGE_URL. The value is the static URL from which a Ruby distribution is to be downloaded, right? But a single Ruby install can attempt to download more archives than just the Ruby version. For example, some installs will also need to fetch openssl from a remote server. Does RUBY_BUILD_MIRROR_PACKAGE_URL affects all the downloads initiated within ruby-build, or just the Ruby-related one? If it doesn't affect all downloads, then I fear that this mirror approach isn't adequate.

Second, I am confused about the internal variable RUBY_BUILD_DEFAULT_MIRROR. Why would we have separate logic for the default mirror and for user-provided mirrors? That feels like a code smell to me. Our mirroring logic should be flexible enough to handle either our own mirror or user-provided mirrors without conditionals.

Lastly, I do dig the new RUBY_BUILD_MIRROR_CMD option because it seems the most flexible one. If that gets merged, I would vote to eliminate RUBY_BUILD_MIRROR_PACKAGE_URL and RUBY_BUILD_DEFAULT_MIRROR.

bin/ruby-build
```diff
@@ -367,6 +367,8 @@ fetch_tarball() {
       fi
     elif [ -n "$RUBY_BUILD_MIRROR_PACKAGE_URL" ]; then
       mirror_url="$RUBY_BUILD_MIRROR_PACKAGE_URL"
+    elif [ -n "$RUBY_BUILD_MIRROR_CMD" ]; then
+      mirror_url="$("$RUBY_BUILD_MIRROR_CMD" "$package_url" "$checksum")"
```
mislav, line 371: This should use `command` to ensure that the name executable is called even if it coincides with a name of a bash builtin or a shell function defined in this file
```suggestion
      mirror_url="$(command "$RUBY_BUILD_MIRROR_CMD" "$package_url" "$checksum")"
```

uzxmx:
@mislav  Thanks for the suggestion. It's truly useful. Already adopted.

I agree the combination of `RUBY_BUILD_MIRROR_PACKAGE_URL` with `RUBY_BUILD_DEFAULT_MIRROR` is a bit confusing unless we read the whole logic carefully. Those names might mislead new users. If the old compatibility is not desired, I also vote to eliminate them.

mislav:
> Those names might mislead new users. If the old compatibility is not desired, I also vote to eliminate them.

Ah, I just realized that RUBY_BUILD_MIRROR_PACKAGE_URL was supported since Dec 2020. I wasn't aware of that. In the interest of backwards compatibility, it should keep being supported, but perhaps we should strip it from documentation in favor of RUBY_BUILD_MIRROR_CMD and perhaps print a deprecation warning to stderr when someone attempts to use it.

uzxmx:
@eregon  @hsbt @mislav 

Could you help to make a conclusion about what else should be done? I'm eager to make the according changes to get it merged. 😄

3UMOVERSANDPACKERS:
Ah, I just realized that RUBY_BUILD_MIRROR_PACKAGE_URL has been supported since December 2020. I wasn't aware of that. In the interest of backwards compatibility, it should keep being supported, but perhaps we should[ strip it from documentation in]() favor of RUBY_BUILD_MIRROR_CMD and perhaps print a [deprecation warning] (https://3umovers.com/packers-and-movers-in-Lahore/l) to standby when someone attempts to use it.

Hsenaai (approved):

jasonkarns:
I just recently (finally) have been getting node-build up to date with ruby build the past few weeks. I recently merged https://github.com/nodenv/node-build/pull/901 which backports the `RUBY_BUILD_MIRROR_PACKAGE_URL` env var into node-build. Which means I have node-build's `_MIRROR_CMD` var working in conjunction with `_MIRROR_PACKAGE_URL`.

I should be able to replicate that into ruby-build if that's still desired.

Dustin4444 (approved):

0llirocks:
No sure if this PR still alive but having the RUBY_BUILD_MIRROR_CMD option would solve our problems with the mirror as well.

rosstimoyhy-byte (commented):
[]()

UnknownMp391:
I just came across this PR while trying to use a Ruby mirror in China, and I have to say — the current mirror setup is extremely frustrating for users.

The `RUBY_BUILD_MIRROR_URL` approach (based on SHA2 checksum paths) is incompatible with most commonly-available community mirrors in China, such as `cache.ruby-china.com`, which organize files by version number (e.g., `/ruby-4.0.5.tar.gz`). This forces users to either manually download the tarball every time, or hack around it with `RUBY_BUILD_TARBALL_OVERRIDE`, which is not discoverable for beginners. **I am puzzled as to why `rbenv` isn't using the mirror I configured.**

This PR’s `RUBY_BUILD_MIRROR_CMD` would be a **huge improvement** — it would allow us to simply substitute the domain name while keeping the official path structure, which is exactly what mirrors like Ruby China provide.

I understand the design philosophy behind content-addressed mirrors, but for many users in restricted network environments, **flexibility** is far more important than purity. This seems like a small change that would significantly improve the experience for a large number of users.

Is there any chance this could be revived? I'd be happy to help test or contribute if needed.

Thanks for considering!

PR #1795, opened 2021-09-23
Add a definition for yjit-dev

casperisfine:
YJIT builds just like regular `ruby-head`, just pointing to a different repo.

I think it would be useful to have it available for people who want to try they apps against it.


hsbt:
Is it better to change `yjit-dev`? I think. Because the main branch will point 3.2.0 at the next year, We shouldn't rename this definition with every year without CRuby.

casperisfine:
@hsbt, good point. I made the change.

casperisfine:
Thank you!

arm5382:
👍

Issue #1798, opened 2021-09-27
jruby install failing on OpenJDK 17


bokmann:
According to [the latest release of JRuby](https://www.jruby.org/news)
"JRuby supports Java versions 8 through 17, and should work well with any supported build of OpenJDK."

On my mac, according to the last rbenv and ruby-build installed via homebrew:

`
~$ java --version
openjdk 17 2021-09-14
OpenJDK Runtime Environment Homebrew (build 17+0)
OpenJDK 64-Bit Server VM Homebrew (build 17+0, mixed mode, sharing)

~$ rbenv install jruby-9.3.0.0 
ERROR: Java 8 required, but your Java version was:
openjdk version "17" 2021-09-14
OpenJDK Runtime Environment Homebrew (build 17+0)
OpenJDK 64-Bit Server VM Homebrew (build 17+0, mixed mode, sharing)

BUILD FAILED (macOS 11.6 using ruby-build 20210925)
`

eregon:
It works  for me with the system Java 11.
I guess it's an issue of `require_java` https://github.com/rbenv/ruby-build/blob/bba976f9d8b894897bee27ba52d3fde80912549a/bin/ruby-build#L854-L869 to parse the version of openjdk 17, which seems to have changed format.
cc @deepj which might have the same issue

eregon:
For my local JDK17:
```
$ java --version
openjdk 17 2021-09-14
$ java -version 
openjdk version "17" 2021-09-14
```
`ruby-build` uses the output of `java -version`.

eregon:
The bug seems to be caused by the version 17 having no dot, and as a result the date seems to be considered part of the version incorrectly:
```
version_string: 
version_string2: 17
2021
09
14
found_version: 17
2021
09
14
```
vs
```
version_string: 
version_string2: 11.0.10
2021
01
19
found_version: 11
```

eregon:
Fixed and released.

PR #1799, opened 2021-09-28
Fix version check for Java 17

eregon:
* Fixes https://github.com/rbenv/ruby-build/issues/1798
* OpenJDK 17 reports:
  $ java -version
  openjdk version "17" 2021-09-14
  There is no dot in the version, so we need to only use the first match from grep -o.
* Clarify it is a minimum required java version, not an exact version.


hsbt:
👍

PR #1800, opened 2021-09-28
Prefer maven URLs going forward

headius:
Following up from https://github.com/ruby/setup-ruby/issues/220...

The S3 URLs used by ruby-build are still maintained, but we prefer that apps and tools switch to the Maven URLs. They are mirrored and federated and not dependent on us maintaining our own data store.

Pinging @enebo to see if we should make this change for earlier versions, perhaps back to the earliest version for which we push jruby-dist?

### DIFF
--- a/script/update-jruby
+++ b/script/update-jruby
@@ -12,8 +12,8 @@ version="$1"
 release_directory="$2"
 file="share/ruby-build/jruby-${version}"
 
-basename="jruby-bin-${version}.tar.gz"
-url="https://s3.amazonaws.com/jruby.org/downloads/${version}/${basename}"
+basename="jruby-dist-${version}-bin.tar.gz"
+url="https://repo1.maven.org/maven2/org/jruby/jruby-dist/${version}/${basename}"
 archive="$release_directory/$basename"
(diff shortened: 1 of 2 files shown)


eregon:
Thank you.
Could you update https://github.com/rbenv/ruby-build/blob/master/script/update-jruby too?
You can try to rerun it (`./script/update-jruby 9.3.0.0 .`) and it should not change the 9.3.0.0 file.

headius:
Script updated and confirmed that it does not change the 9.3.0.0 file when executed.

I looked into it, and we have been publishing the jruby-dist artifact as far back as JRuby 1.6.0 in 2011. For now I think just using the new URL going forward is fine, as we worked out long-term storage of the S3 artifacts with Engine Yard a couple years back.

eregon:
I agree, probably best to use the new URL from 9.3 onward, and ruby-build has its own mirror in front anyway.
I installed 9.3 with this change locally and it works.
Thanks for the PR.

enebo:
Sorry I was late to this but I think we should just do it 9.3.0.0 onward.  We can retroactively do another PR if the need ever arises.

PR #1817, opened 2021-11-02
Remove yjit-dev

noahgibbs:
YJIT has been merged with upstream CRuby and the repo used by ruby-build is no longer maintained properly.

At this point the best way to build YJIT is with 3.1.0-dev or similar head-of-CRuby-master.

cc: @casperisfine


casperisfine (approved):
👍 We definitely didn't expect the upstream merge to happen so soon, we wouldn't have added `yjit-dev` if we did.

hsbt:
Thanks. I'm also wondering this definition in the future. It's great to merge yjit to the upstream.

PR #1820, opened 2021-11-05
Patch OpenSSL 1.0 to work with sites that use expired Let's Encrypt CA (DST Root CA X3)

entretechno-jeremiah:
Some gem sources (_e.g._ [Rails Assets](https://rails-assets.org)) are using [the expired Let's Encrypt CA (DST Root CA X3)](https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/), which is causing gem installation to fail. This change [enables the `X509_V_FLAG_TRUSTED_FIRST` flag](https://www.openssl.org/blog/blog/2021/09/13/LetsEncryptRootCertExpire/) by default, which will make it use the new Let's Encrypt CA (ISRG Root X1) where necessary.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1091,6 +1091,22 @@ use_homebrew_openssl() {
 }
 
 build_package_mac_openssl() {
+  # Patch OpenSSL 1.0 to work with sites that use expired Let's Encrypt CA (DST Root CA X3)
+  [[ "$1" != openssl-1.0.* ]] || patch -p0 <<- EOF
+	diff -ruN crypto/x509/x509_vpm.c crypto/x509/x509_vpm.c
+	--- crypto/x509/x509_vpm.c	2019-12-20 07:02:41.000000000 -0600
+	+++ crypto/x509/x509_vpm.c	2021-11-05 09:59:12.062606932 -0500
+	@@ -537,7 +537,7 @@
+	      "default",                 /* X509 default parameters */
+	      0,                         /* Check time */
+	      0,                         /* internal flags */
+	-     0,                         /* flags */
+	+     X509_V_FLAG_TRUSTED_FIRST, /* flags */
+	      0,                         /* purpose */
+	      0,                         /* trust */
+	      100,                       /* depth */
+	EOF
+
   # Install to a subdirectory since we don't want shims for bin/openssl.
   OPENSSL_PREFIX_PATH="${PREFIX_PATH}/openssl"
 


hsbt:
I'm not sure what impact for security risk with this patch. So, I'm negative to apply this patch instead of upgrading OS and other middleware like OpenSSL.

entretechno-jeremiah:
@hsbt Thanks for looking at this.

Unfortunately upgrading the OS is not currently a solution, as the most current version of macOS still has the expired DST Root CA X3 certificate in [`/System/Library/Keychains/SystemRootCertificates.keychain`](https://github.com/rbenv/ruby-build/blob/fa04a02b68332b0d3663bb0a7fa4d85dd2a6a39f/bin/ruby-build#L1141) with no option to remove it.

And upgrading OpenSSL is not an option either, because the most recent publicly available version of OpenSSL 1.0 (1.0.2u) does not fix this, and [OpenSSL 1.1 is not an option](https://github.com/rbenv/ruby-build/issues/1353).

If it alleviates security concerns, this is the same change as [one that Ubuntu made to address this same issue](https://git.launchpad.net/ubuntu/+source/openssl/commit/?h=applied/ubuntu/xenial-security&id=5dd6415dd3a71d37a8a440ebbe5aeaa5f56645d4) (I meant to link this in the description, but I was having trouble finding the original source).

dentarg:
@hsbt https://bugs.chromium.org/p/boringssl/issues/detail?id=439 has more details and background about this, if it can be of any help

mislav:
@dentarg Thank you for the information and for the proposed fix! Like @hsbt, I'm unconvinced that we should apply the patch to our project. My reasons are:
- OpenSSL 1.0 is only used by Ruby versions < 2.4, which are all EOL.
- I'm assuming that the certificate problem only affects a small number of people: those who are on mac OS **and** linked their Ruby to OpenSSL1 1.0 **and** are trying to install from affected gem sources.
- If a Ruby user has some problem with OpenSSL, they should compile OpenSSL by themselves using their own configuration (and/or patches) and link Ruby to that using `RUBY_CONFIGURE_OPTS=--with-openssl-dir=/path/to/my/openssl`.

In my opinion, the fact that we have and maintain the `build_package_mac_openssl` hack within our project in the first place has already proven problematic in the past. We originally provided this to ease Ruby installation on mac OS which ships a "broken" version of OpenSSL, but us compiling OpenSSL on behalf of our users has sometimes caused more trouble than it's worth and presented a significant maintenance burden. We will refrain from adding any more "magic" functionality to this hack and will instead return the responsibility on to the user to maintain their own installation of OpenSSL on their system should they want to configure anything.

PR #1877, opened 2021-12-27
Update bump-homebrew-formula action

mislav:
The v1 branch is now fixed. Ref. https://github.com/rbenv/ruby-build/pull/1839

Thanks @Intrepidd


hsbt:
👍

PR #1889, opened 2022-01-03
Add issue templates

mislav:
This adds issue templates for this repository. It defines 3 categories of reports:
- "bug"
- "enhancement" (feature requests) 
- "BUILD FAILED" - instructs people to post to the Discussions section instead of filing Issues.

The goal is to redirect people to Discussions instead of Issues when seeking support for various build failures which, historically, are rarely caused by bugs in ruby-build code.


hsbt (approved):
👍  Thanks! It's the great to update for the users.

mislav:
Here is how it looks like:

PR #1929, opened 2022-02-03
Automatically detect and link to Homebrew's libyaml

### LINKED ISSUE #1928: Use yaml from homebrew
Homebrew is no longer located in `/usr/local` on Apple Silicon. It is now in `/opt/homebrew`. This leads to brew-originated yaml not being picked up during the build

We should probably manage yaml the same way we do it with readline. Looking at the code there's already a function that should do what is needed but it's not being called anywhere

https://github.com/rbenv/ruby-build/blob/cbdbc6e44be4756df3fc526ee97c1a5ce0632b74/bin/ruby-build#L1029-L1037

should we add it [here](https://github.com/rbenv/ruby-build/blob/cbdbc6e44be4756df3fc526ee97c1a5ce0632b74/bin/ruby-build#L582) ?

dreyks:
closes #1928 

turns out there already was a test for this behaviour but it was falsely passing due to `needs_yaml` (that was run only from tests) was calling `use_homebrew_yaml` (which was otherwise not called in production)

Bash is not my native language and I've had very limited experience with BATS framework so there might be things I've done wrong :)

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -558,8 +558,21 @@ build_package_standard_build() {
   local PACKAGE_CFLAGS="${package_var_name}_CFLAGS"
 
   if [ "$package_var_name" = "RUBY" ]; then
-      use_homebrew_readline || use_freebsd_pkg ||true
+    if [[ "$RUBY_CONFIGURE_OPTS ${RUBY_CONFIGURE_OPTS_ARRAY[*]}" != *--with-readline-dir=* ]]; then
+      use_homebrew_readline || use_freebsd_readline || true
+    fi
+    if [[ "$RUBY_CONFIGURE_OPTS ${RUBY_CONFIGURE_OPTS_ARRAY[*]}" != *--with-libyaml-dir=* ]]; then
+      use_homebrew_yaml || true
+    fi
+    if [[ "$RUBY_CONFIGURE_OPTS ${RUBY_CONFIGURE_OPTS_ARRAY[*]}" != *--with-gmp-dir=* ]]; then
       use_homebrew_gmp || true
+    fi
+    if [[ "$RUBY_CONFIGURE_OPTS ${RUBY_CONFIGURE_OPTS_ARRAY[*]}" != *--with-openssl-dir=* ]]; then
+      if [ "FreeBSD" = "$(uname -s)" ] && [ -f /usr/local/include/openssl/ssl.h ]; then
+        # use openssl installed from Ports Collection
+        package_option ruby configure --with-openssl-dir="/usr/local"
+      fi
+    fi
   fi
 
   ( if [ "${CFLAGS+defined}" ] || [ "${!PACKAGE_CFLAGS+defined}" ]; then
@@ -1037,18 +1050,10 @@ use_homebrew_gmp() {
   fi
 }
 
-use_freebsd_pkg() {
-  # check if FreeBSD
+use_freebsd_readline() {
   if [ "FreeBSD" = "$(uname -s)" ]; then
-    # use openssl if installed from Ports Collection
-    if [ -f /usr/local/include/openssl/ssl.h ]; then
-      package_option ruby configure --with-openssl-dir="/usr/local"
-    fi
-
-    # check if 11-R or later
-    release="$(uname -r)"
+    local release="$(uname -r)"
     if [ "${release%%.*}" -ge 11 ]; then
-      # prefers readline to compile most of ruby versions
       if pkg info -e readline > /dev/null; then
         # use readline from Ports Collection
         package_option ruby configure --with-readline-dir="/usr/local"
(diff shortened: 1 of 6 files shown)


bin/ruby-build
```diff
@@ -1022,20 +1023,28 @@ locate_llvm() {
...
 }
 
+homebrew_yaml_dir() {
+  if [[ "$RUBY_CONFIGURE_OPTS" != *--with-libyaml-dir=* ]]; then
+    local libdir="$(brew --prefix libyaml 2>/dev/null || true)"
+    if [[ -d "$libdir" ]]; then
+      echo "$libdir"
+    fi
+  fi
+}
```
dreyks, line 1046: I've moved common parts to a separate function so that it's possible to check whether yaml is needed without modifying package options

test/build.bats
```diff
@@ -186,7 +186,10 @@ OUT
   mkdir -p "$brew_libdir"
 
   stub uname '-s : echo Linux'
-  stub brew "--prefix libyaml : echo '$brew_libdir'" false
+  stub brew \
+    "--prefix libyaml : echo '$brew_libdir'" \
+    false \
+    "--prefix libyaml : echo '$brew_libdir'"
```
dreyks, line 192: 1. `needs_yaml` call from the definition file
2. readline check from code
3. libyaml check from code

bin/ruby-build
```diff
@@ -1022,20 +1023,28 @@ locate_llvm() {
...
-  [[ "$RUBY_CONFIGURE_OPTS" != *--with-libyaml-dir=* ]] &&
-  ! use_homebrew_yaml
+  [ -z "$(homebrew_yaml_dir)" ]
 }
 
 use_homebrew_yaml() {
-  local libdir="$(brew --prefix libyaml 2>/dev/null || true)"
-  if [ -d "$libdir" ]; then
+  local libdir="$(homebrew_yaml_dir)"
+  if [ ! -z "$libdir" ]; then
```
mislav, line 1031: Please use `-n` instead of `! -z`

bin/ruby-build
```diff
@@ -1022,20 +1023,28 @@ locate_llvm() {
 }
 
 needs_yaml() {
-  [[ "$RUBY_CONFIGURE_OPTS" != *--with-libyaml-dir=* ]] &&
-  ! use_homebrew_yaml
+  [ -z "$(homebrew_yaml_dir)" ]
```
mislav, line 1026: For backwards compatibility, please restore to the old implementation of `needs_yaml`. There was no need to change it, and the new implementation doesn't print the line `ruby-build: using libyaml from homebrew` anymore.

bin/ruby-build
```diff
@@ -1022,20 +1023,28 @@ locate_llvm() {
...
     package_option ruby configure --with-libyaml-dir="$libdir"
   else
     return 1
   fi
 }
 
+homebrew_yaml_dir() {
+  if [[ "$RUBY_CONFIGURE_OPTS" != *--with-libyaml-dir=* ]]; then
+    local libdir="$(brew --prefix libyaml 2>/dev/null || true)"
+    if [[ -d "$libdir" ]]; then
```
mislav, line 1042: Style: double brackets are not needed here, please use single ones

bin/ruby-build
```diff
@@ -1022,20 +1023,28 @@ locate_llvm() {
...
+  if [ ! -z "$libdir" ]; then
     echo "ruby-build: using libyaml from homebrew"
     package_option ruby configure --with-libyaml-dir="$libdir"
   else
     return 1
   fi
 }
 
+homebrew_yaml_dir() {
+  if [[ "$RUBY_CONFIGURE_OPTS" != *--with-libyaml-dir=* ]]; then
```
mislav, line 1040: It could be surprising that a function called `homebrew_yaml_dir` performs this check first. I would rather that the caller of `homebrew_yaml_dir` performs the check instead, even if that's going to lead to some repetition. Explicitness is better than burying some magic behavior deep down a call chain

mislav (changes requested):
Thanks for taking this on! Left some style suggestions

dreyks:
thanks for the review, @mislav!

I ended up reverting most of my changes and instead dropped the fixture usage in favor of an inline definition to avoid future false negatives in this test

dreyks:
anything else I need to do to move this forward?

KevDog:
I think I have issues this PR could fix. Trying to install on Ventura is bailing on not finding libyaml.

mislav:
@KevDog In the meantime, you can explicitly set `RUBY_CONFIGURE_OPTS=--with-libyaml-dir=/path/to/libyaml` to work around this in your setup.

bai:
Friendly bump 🙏🏼 Another use case is that package managers that depend on `ruby-build`, e.g. `asdf` are failing to install ruby and need more magic trickery to get this fixed.

mislav (approved):
Finally fixed the tests. Thanks for the patience!

KevDog:
Thanks!

Issue #1955, opened 2022-04-08
Builds of TruffleRuby by asdf are broken


pinealservo:
This is actually a `ruby-build` issue.  It was introduced here: https://github.com/rbenv/ruby-build/pull/1784

It's not inherent to the core of `ruby-build` but is due to code meant to prevent `ruby-build` from deleting `/usr/local` or other sorts of FHS-compliant install locations, which TruffleRuby doesn't support: https://github.com/oracle/truffleruby/issues/1389

The problem with the existing method of detecting these installs is that it's too general and conflicts with the way `asdf` install works. [This line](https://github.com/asdf-vm/asdf/blob/de6e22f909946f7d17047f4aeab41e581546b674/lib/commands/command-install.bash#L220) is run to make the directory just before the plugin (which in this case is `asdf-ruby` which then calls `ruby-build`) gets to run. These install directories are a core part of how asdf (a set of bash scripts) tracks what's installed and what versions are available.

It seems to me that the check should explicitly be for FHS paths rather than just existing directories, or at least they should explicitly allow existing paths with `.asdf` as a component in the path.

_Originally posted by @pinealservo in https://github.com/rbenv/ruby-build/discussions/1911#discussioncomment-2533456_

PR #1956, opened 2022-04-09
Allow truffleruby to install in an existing empty directory

pinealservo:
Fix for #1955 #1911 

The check to prevent deleting user files during clean_prefix_path by disallowing installs into existing directories that weren't earlier truffleruby installs broke asdf installs, as asdf always creates an empty directory before installing.

This fix adds a second exception to the existing directory case when that directory is empty. The check is done by testing the string length of ls output. The logic was manually tested against non-existing directories, existing empty directories, existing non-empty directories in general, and existing non-empty directories that contained a file named "bin/truffleruby".

I also copied the file over my existing asdf version of ruby-build and made sure that it successfully installed truffleruby.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -794,10 +794,12 @@ remove_windows_files() {
 }
 
 clean_prefix_path_truffleruby() {
-  if [ -d "$PREFIX_PATH" ] && [ ! -e "$PREFIX_PATH/bin/truffleruby" ]; then
+  if [ -d "$PREFIX_PATH" ] &&
+     [ ! -e "$PREFIX_PATH/bin/truffleruby" ] &&
+     [ ! -z "$(ls -A $PREFIX_PATH)" ]; then
     echo
     echo "ERROR: cannot install TruffleRuby to $PREFIX_PATH, which does not look like a valid TruffleRuby prefix" >&2
-    echo "TruffleRuby only supports being installed to a not existing directory or replacing an existing TruffleRuby installation"
+    echo "TruffleRuby only supports being installed to a not existing directory, an empty directory, or replacing an existing TruffleRuby installation"
     echo "See https://github.com/oracle/truffleruby/issues/1389 for details" >&2
     return 1
   fi


bin/ruby-build
```diff
@@ -793,11 +793,21 @@ remove_windows_files() {
   rm -f bin/*.exe bin/*.dll bin/*.bat bin/jruby.sh
 }
 
+check_path_is_empty() {
+  shopt -s nullglob dotglob
+  local files_array=(${1}/*)
+  local is_empty=${#files_array[*]}
+  shopt -u nullglob dotglob
+  return $is_empty
```
eregon, line 801: I think this solution feels rather cryptic. Googling a bit for "bash check directory empty" I get https://superuser.com/questions/352289/bash-scripting-test-for-empty-directory.
I think the `if [ -z "$(ls -A /path/to/dir)" ]; then` solution is much simpler, easier to read, and likely also more portable.

Could you switch to that?

eregon (commented):
Thank you for the PR!

bin/ruby-build
pinealservo: I agree it's not the most straightforward one; I guess I misinterpreted your request for "in bash" to mean using bash built-ins rather than just a standard POSIX shell environment. 

The array functionality involved has been in bash for a long time and is apparently portable to most ksh-like shells; it just expands a glob into an array and counts the members using the standard indexed-array functionality. But arrays are definitely not a well-known or POSIX-specified part of bash and I had to look them up to understand how it worked.

I'll happily switch to the ls version though; will update the PR in a bit.

eregon (approved):
Thanks!

mislav:
@eregon Should I finally address this TODO? 😸 https://github.com/rbenv/ruby-build/blob/5c814edb0236f5a104bde46f5f0e6ad915a8c928/script/release#L10

eregon:
I think it's not necessary :) I was thinking I could just add `-2` or add some suffix to the tag in the release script, but it seems safest to just wait a day in case something depends on the tag format.

eregon:
Released as v20220415.

PR #1967, opened 2022-05-01
Skip building docs when building openssl

eregon:
* See https://github.com/openssl/openssl/issues/8170
* The install_sw target exists since 1.0.0 or earlier:
  https://github.com/openssl/openssl/blob/OpenSSL_1_0_0-stable/Makefile.org


parkr:
Hooray, thank you @eregon!

PR #1973, opened 2022-05-09
Re-enable ssldir creation target for openssl install

jpr5:
Prior commit attempts to shorten install by omitting documentation install,
however an additional target is needed to get the conf dirs for capturing the
PEMs et al.

Solves issue mentioned in [discussion 1972](https://github.com/rbenv/ruby-build/discussions/1972).

Related to https://github.com/rbenv/ruby-build/pull/1967

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1120,7 +1120,7 @@ build_package_mac_openssl() {
   package_option openssl make -j 1
 
   # Use install_sw instead of install to skip building docs which is slow
-  MAKE_INSTALL_TARGET=install_sw build_package_standard "$@"
+  MAKE_INSTALL_TARGET="install_sw install_ssldirs" build_package_standard "$@"
 
   # Extract root certs from the system keychain in .pem format and rehash.
   local pem_file="$OPENSSLDIR/cert.pem"
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1119,8 +1119,13 @@ build_package_mac_openssl() {
   # gives precedence to the last -j option, so we can override that.
   package_option openssl make -j 1
 
-  # Use install_sw instead of install to skip building docs which is slow
-  MAKE_INSTALL_TARGET="install_sw install_ssldirs" build_package_standard "$@"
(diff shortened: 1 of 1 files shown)


mislav (approved):

mislav:
@jpr5 Thank you! The fix looks good to me.

eregon:
In openssl 1.0.2 (the oldest version used in ruby-build), there appears to be no `install_ssldirs` target though:
https://raw.githubusercontent.com/openssl/openssl/OpenSSL_1_0_2-stable/Makefile.org

So that seems an issue, as we'd break those (e.g., 2.3.8).
Could you try installing 2.3.8 with your change and see if that breaks (as expected)?
And then tweak your patch so it'd work both for 2.3.8 and newer versions?

eregon:
I should have started with: Thank you for the fix and sorry for breaking that (at least it's not on a ruby-build release).

mislav:
@eregon If we cannot reliably support all OpenSSL versions that we use, then we should revert https://github.com/rbenv/ruby-build/pull/1967.

eregon:
@mislav I'll make my own PR to try to fix this ASAP.

> If compilation speed is an issue, the ultimate solution would be that ruby-build doesn't compile openssl in the first place

The cases I let ruby-build compile openssl is the cases I would need to somehow compile openssl anyway. But maybe on macOS we could default to use Homebrew's openssl, or the system openssl if it's usable nowadays.

eregon:
I confirmed:
1.0.2 doesn't have `install_ssldirs`: https://github.com/openssl/openssl/blob/OpenSSL_1_0_2-stable/Makefile.org
1.1.0 has `install_ssldirs`: https://github.com/openssl/openssl/blob/OpenSSL_1_1_0-stable/Configurations/unix-Makefile.tmpl
3.0 has `install_ssldirs`: https://github.com/openssl/openssl/blob/openssl-3.0/Configurations/unix-Makefile.tmpl
master has `install_ssldirs`: https://github.com/openssl/openssl/blob/master/Configurations/unix-Makefile.tmpl

So we just need a check for openssl < 1.1

eregon (approved):

PR #1974, opened 2022-05-11
Install openssl 1 when openssl is 3

znz:
Ubuntu 22.04 includes openssl 3.0.2, and libssl-dev is also 3.0.2 only.
But Ruby < 3.1 does not support openssl 3.
So I want to install openssl 1 automatically.

I just tested `rbenv install 3.0.4` and `rbenv install 2.7.6` with this on Ubuntu 22.04.
Sorry if this makes trouble on other platforms.


bin/ruby-build
```diff
@@ -1119,13 +1133,27 @@ build_package_mac_openssl() {
...
+    security find-certificate -a -p /Library/Keychains/System.keychain > "$pem_file"
+    security find-certificate -a -p /System/Library/Keychains/SystemRootCertificates.keychain >> "$pem_file"
+  else
+    # symlink to the system openssl certs
+    local SYSTEM_OPENSSLDIR
+    SYSTEM_OPENSSLDIR=$(openssl version -d | sed -e 's/.*: "//' -e 's/"$//' || true)
+    if [ -n "$SYSTEM_OPENSSLDIR" ]; then
+      ln -sf "$SYSTEM_OPENSSLDIR/cert.pem" "$OPENSSLDIR/cert.pem"
+      ln -snf "$SYSTEM_OPENSSLDIR/certs" "$OPENSSLDIR/certs"
+    fi
```
eregon, line 1155: Could you compare/unify with the logic at https://github.com/rbenv/ruby-build/compare/master...eregon:openssl-linux ?
IIRC there is no cert.pem on Debian-based distributions (at the time I wrote that).

eregon:
Thank you for the PR, I agree this is useful and needed, given ruby-core seems unwilling to add openssl 3 support to Ruby 3.0 and 2.7: https://bugs.ruby-lang.org/issues/18658#note-8
I will also need this to build Ruby <= 3.0 on the upcoming ubuntu 22.04 github action runners.

mislav:
@znz I appreciate the effort that went into this, and I do agree that the current situation on Ubuntu is unfriendly for our users. However, I'd be hard to convince that this is the way to go.

Our current `has_mac_broken_openssl` approach was initially made because Ruby was usually absolutely uninstallable on macOS due to its patched system OpenSSL version. It was a way to ease Ruby installation for macOS users, but it opened a whole can of worms:

- It drastically slowed down installation of every Ruby version because every `rbenv install` call would compile its own contained openssl instance;
- It made the ruby-build project responsible for debugging when that compilation failed, which we (ruby-build maintainers) didn't always have capacity nor knowledge to do;
- It wouldn't result in a great long-term experience for our users because, since every Ruby version would have its own openssl instance, a new allow-listed certificate added to the system's certificate bundle wouldn't propagate to or be known by Ruby versions;
- Since the user wasn't responsible or even aware of openssl having been built on their behalf for each Ruby version, they typically also wouldn't apply openssl security upgrades.

Basically, all of this would also happens on other platforms too if we started to extend this approach outside of just macOS. However, what I would ideally like to see is that ruby-build stops building or maintaining its own openssl versions. Instead, a ruby-build process should ideally just quickly error out if a compatible openssl version wasn't found on the system, with instructions on how the user should take initiative to fix it.

@eregon @hsbt Thoughts?

eregon:
> It drastically slowed down installation of every Ruby version

Certainly better than failing. Maybe we could install openssl in a well-known place which could be reused if the same openssl version is used by multiple Ruby versions.

> It made the ruby-build project responsible for debugging when that compilation failed

This definitely sucks, but I think we'd get almost as much reports even without that. Failing to compile ruby feeks like 90% from openssl issues, and telling people to install manually wouldn't improve that, I don't see it likely people would report to upstream openssl, and they likely wouldn't be able to help with e.g. how ruby & ruby-build use that openssl.
IMHO many such issues should be filed to https://bugs.ruby-lang.org/ mostly, because errors are too obscure and it's too easy to pick a mix of openssl headers and libs from different versions (e.g., https://github.com/postmodern/ruby-install/issues/412#issuecomment-1118178152), but that's not what happens.

> It wouldn't result in a great long-term experience for our users because, since every Ruby version would have its own openssl instance, a new allow-listed certificate added to the system's certificate bundle wouldn't propagate to or be known by Ruby versions;

That's a non-issue with this PR, because it symlinks the system certificates.

> with instructions on how the user should take initiative to fix it.

The problem is this is very annoying to do on Ubuntu 22.04 and other recent distros which default to openssl 3.
AFAIK it is not possible to install an openssl 1.1 with the package manager (`apt`), the package manager only knows openssl 3 and no other openssl version. It's always been like that, there can only be one system openssl, i.e., one version in `/usr/include/openssl/opensslv.h` (in some old distros one could choose between 1.0 and 1.1, now it's just 3 and that's it).
So the only solution would be to compile manually from source, which seems worse to ask users to do manually vs ruby-build doing it.
Maybe ruby-build could fail by default in a system-openssl3 but need 1.1 situation, but have a flag to opt-in to make this more explicit?

The situation is already a pain and already existed for Ruby <= 2.3 (which doesn't compile with openssl 1.1 and needs 1.0), but those are old so it not so commonly a problem. But here it's simply not possible to build Ruby 3.0 on Ubuntu 22.04 with ruby-build, unless you manually download & compile openssl and link the certificates.

There is another issue to not use the system openssl not mentioned above:
* When a Ruby is not compiled against the system openssl (i.e. installed by the package manager), it means that Ruby does not work with other packages installed by that package manager if those packages depend on openssl. As an example database packages typically depend on openssl and are typically installed by the packager manager so mysql/postgres/etc typically don't work in such cases, because one ends up with two libssl loaded in the same process and that either breaks or segfaults. The workaround would be to manually compile these databases against the openssl Ruby was built against, which seems very difficult.

That reason is IMHO why Ruby 3.0 should support openssl 3, otherwise it's not possible/very difficult to use Ruby 3.0 with database drivers on Ubuntu 22.04.

BTW on macOS I think ruby-build should use Homebrew's openssl by default to avoid this issue, because that's the "system libssl" of macOS for most users (i.e., they'd install mysql/postgres via brew and I think those link against the Homebrew openssl). But I don't know the details of openssl on macOS anymore, it seems it ships with LibreSSL nowadays?

mislav:
> Maybe we could install openssl in a well-known place which could be reused

That would be great! If ruby-build is invoked in the context of rbenv, maybe we could have a shared openssl installation under `~/.rbenv/openssl` or something like that?

> That's a non-issue with this PR, because it symlinks the system certificates.

Ah, true; thanks for pointing that out. I guess this wasn't possible on macOS because there were no physical files to link to.

> That reason is IMHO why Ruby 3.0 should support openssl 3

Agreed—in the ideal world. I don't have the expertise to know what would actually be involved with that.

@eregon In the interest on making this changeset minimal, how would you feel about building this cross-platform openssl functionality into the old functions `mac_openssl` and `has_mac_broken_openssl`, so that we don't have to do rename all across all current build formulae? Sure, the functions would then become a slight misnomer (since they would apply to all systems, not just macOS), but we could address that in the future.

eregon:
> That would be great! If ruby-build is invoked in the context of rbenv, maybe we could have a shared openssl installation under `~/.rbenv/openssl` or something like that?

Maybe simply `~/.ruby-build/openssl`?
I also thought of `~/.rubies/openssl` too (IMHO the best place to install Rubies in a Ruby-switcher-agnostic way).
It could also be configurable.

mislav:
No, I meant to ping you as a maintainer.

> I think we should give it a fully general name like:
> 
> ```
> install_package "openssl-1.0.2u" ..." openssl --if system_openssl_version_does_not_match
> ```

Sure, that's descriptive, but I find it verbose and scary-looking. How about something generic and simpler, since this is going to be repeated across hundreds of build definitions:

```
openssl --if needs_openssl
```

mislav:
> Maybe simply `~/.ruby-build/openssl`?
> I also thought of `~/.rubies/openssl` too

I'd be open to that, but it would break the unspoken contract of rbenv never writing to locations outside of `~/.rbenv`, and `ruby-build` never writing anything to locations outside of the build target directory.

> It could also be configurable.

Like most features of ruby-build, I definitely agree it should be configurable via an environment variable.

eregon:
`openssl --if needs_openssl` sounds fine to me, it's a bit less clear (Ruby always "needs" openssl otherwise it's barely usable) but that probably doesn't matter too much, the behavior regarding openssl should anyway be documented in the README or so.

> I'd be open to that, but it would break the unspoken contract of [...] ruby-build never writing anything to locations outside of the build target directory.

Right, but that is unavoidable if we want to share the openssl installation between different Ruby versions.
Maybe the default should remain "inside the ruby installation prefix" then to honor that contract, although it would mean non-rbenv users would miss on the build time gains.

For rbenv we could just have rbenv pass the path for where to put openssl (`SOME_ENV_VAR=~/.rbenv`).

BTW, I was thinking we'd install openssl e.g. in `~/.rbenv/openssl-1.0.2u/`, so include the version in it, always and append `/openssl-$VERSION` to the given path.

mislav:
> Maybe the default should remain "inside the ruby installation prefix" then to honor that contract, although it would mean non-rbenv users would miss on the build time gains.

Non-rbenv users could set something like RUBY_BUILD_OPENSSL_DIR to keep openssl in a shared location. We do the same with speeding up downloads: rbenv users have ruby-build caching downloads to `~/.rbenv/cache` (if it exists) and non-rbenv users have to explicitly specify RUBY_BUILD_CACHE_PATH.

bin/ruby-build
```diff
@@ -1081,6 +1081,16 @@ has_broken_mac_openssl() {
   [[ "$RUBY_CONFIGURE_OPTS" != *--with-openssl-dir=* ]]
 }
 
+has_openssl_3() {
+  local openssl_version="$(/usr/bin/openssl version 2>/dev/null || true)"
```
eregon, line 1085: What should we do if openssl does not exist/there is system openssl?
I think we should fail in that case and instruct the user to follow https://github.com/rbenv/ruby-build/wiki#suggested-build-environment.
Otherwise we might install an openssl version which would be redundant with the system one, which is bad.

(Regarding detection, in the future, not for this PR, `/usr/bin/openssl` doesn't seem ideal e.g. on macOS where we should probably consider Homebrew and maybe macports. I think we should just pick `openssl` from PATH, so just `openssl version`. But anyway, not for this PR)

share/ruby-build/1.9.3-dev
```diff
@@ -1,3 +1,3 @@
-install_package "openssl-1.0.2u" "https://www.openssl.org/source/openssl-1.0.2u.tar.gz#ecd0c6ffb493dd06707d38b14bb4d8c2288bb7033735606569d8f90f89669d16" mac_openssl --if has_broken_mac_openssl
+install_package "openssl-1.0.2u" "https://www.openssl.org/source/openssl-1.0.2u.tar.gz#ecd0c6ffb493dd06707d38b14bb4d8c2288bb7033735606569d8f90f89669d16" openssl --if has_openssl_3_or_broken_mac_openssl
```
eregon, line 1: Could you use `--if needs_openssl` as discussed earlier?

bin/ruby-build
```diff
@@ -1119,18 +1133,45 @@ build_package_mac_openssl() {
...
+    # Use install_sw install_ssldirs instead of install to skip building docs which is slow.
+    # OpenSSL 1.1+ also needs install_ssldirs, 1.0 does not have that target.
+    if [[ "$1" == openssl-1.0.* ]]; then
+      MAKE_INSTALL_TARGET="install_sw" build_package_standard "$@"
+    else
+      MAKE_INSTALL_TARGET="install_sw install_ssldirs" build_package_standard "$@"
+    fi
   else
-    MAKE_INSTALL_TARGET="install_sw install_ssldirs" build_package_standard "$@"
+    build_package_standard "$@"
```
eregon, line 1145: We should also do the optimization to not build docs on non-macOS, so I think this section of the code shouldn't change.

dentarg:
> However, what I would ideally like to see is that ruby-build stops building or maintaining its own openssl versions.

I like that `ruby-build` builds OpenSSL for each Ruby version I install. 

I used to use `ruby-install` to build/install Ruby on my Mac. Often I discovered issues with Ruby versions previously built after a `brew upgrade` upgraded OpenSSL. I think I had to re-build the older Rubies to get them usable again. I've never run into that since switching to `ruby-build`.

mislav:
> Often I discovered issues with Ruby versions previously built after a `brew upgrade` upgraded OpenSSL. I think I had to re-build the older Rubies to get them usable again.

That happened to me as well. I think this was because Homebrew by default does aggressive pruning of older versions of software whenever it installs a newer version. Once I've configured `export HOMEBREW_NO_INSTALL_CLEANUP=true` globally on my system, my rubies were never broken even though they were linked to `brew --prefix openssl@1.1`

bin/ruby-build
```diff
@@ -1119,18 +1133,45 @@ build_package_mac_openssl() {
...
+    security find-certificate -a -p /Library/Keychains/System.keychain > "$pem_file"
+    security find-certificate -a -p /System/Library/Keychains/SystemRootCertificates.keychain >> "$pem_file"
+  elif [ -e /etc/pki/tls/cert.pem ]; then # RedHat
+    # See https://github.com/rubygems/rubygems/issues/2415#issuecomment-509806259
+    rm -rf "$OPENSSLDIR/certs" "$pem_file"
+    ln -s /etc/pki/tls/certs "$OPENSSLDIR"
+    ln -s /etc/pki/tls/cert.pem "$pem_file"
+  elif [ -e /etc/ssl/certs/ca-certificates.crt ]; then # Debian
+    # See https://github.com/rubygems/rubygems/issues/2415#issuecomment-509806259
+    rm -rf "$OPENSSLDIR/certs" "$pem_file"
```
gfx, line 1160: Looks like `$pem_file` is only defined in the `if is_mac` clause, and thus it's empty here, causing the following `ln -s /etc/ssl/certs/ca-certificates.crt "$pem_file"` to fail.

bin/ruby-build
znz: I misunderstand error: `ln: failed to create symbolic link '': No such file or directory`.
I thought it happened by `MAKE_INSTALL_TARGET`.
But it caused by bug mentioned at <https://github.com/rbenv/ruby-build/pull/1974#discussion_r874422761>.

amg-web:
I do not like any of the offered solution.
due to this line in error: OpenSSL >= 1.0.1, **< 3.0.0** or LibreSSL >= 2.5.0
I know from here https://discourse.ubuntu.com/t/openssl-3-0-transition-plans/24453
ruby2.7 **can be compiled** with OpenSSL 3.0
so we need to try change the requirements line to something like this:
OpenSSL >= 1.0.1, < 3.0.3 or LibreSSL >= 2.5.0

bin/ruby-build
```diff
@@ -1127,10 +1141,33 @@ build_package_mac_openssl() {
...
+    ln -s /etc/ssl/certs "$OPENSSLDIR"
+    ln -s /etc/ssl/certs/ca-certificates.crt "$pem_file"
+  elif type openssl &>/dev/null; then
```
eregon, line 1159: As a note it seems relatively common that openssl headers are installed yet there is no `openssl` executable, for instance this is the case on Fedora when installing `openssl-dev` (and not `openssl`).
I think it doesn't matter too much since we only get here for non-Debian non-RedHat systems, when a special openssl version is needed.

bin/ruby-build
```diff
@@ -1081,6 +1081,16 @@ has_broken_mac_openssl() {
   [[ "$RUBY_CONFIGURE_OPTS" != *--with-openssl-dir=* ]]
 }
 
+has_openssl_3() {
+  local openssl_version="$(openssl version 2>/dev/null || true)"
```
eregon, line 1085: I think we need a better check here because there might be openssl headers yet no openssl executable (https://github.com/rbenv/ruby-build/pull/1974/files#r885043692).

One idea is looking in /usr/include/openssl/opensslv.h, another is to compile a small C program to print `OPENSSL_VERSION_TEXT`.

share/ruby-build/3.1.0
```diff
@@ -1,2 +1,2 @@
-install_package "openssl-1.1.1n" "https://www.openssl.org/source/openssl-1.1.1n.tar.gz#40dceb51a4f6a5275bde0e6bf20ef4b91bfc32ed57c0552e2e8e15463372b17a" mac_openssl --if has_broken_mac_openssl
+install_package "openssl-1.1.1n" "https://www.openssl.org/source/openssl-1.1.1n.tar.gz#40dceb51a4f6a5275bde0e6bf20ef4b91bfc32ed57c0552e2e8e15463372b17a" openssl --if needs_openssl
```
eregon, line 1: For CRuby 3.1+, OpenSSL 3 is supported, so there we should not install a custom openssl but use the system openssl.
So this could remain `--if has_broken_mac_openssl`.

I'm thinking we should have `needs_openssl3` (3.0 or older, so not 4+), `needs_openssl11` (1.1 or older, so not 3+), `needs_openssl10` (1.0 or older, so not 1.1+).
For CRuby 3.1+ we'd use `needs_openssl3`, for other CRuby versions which `install_package "openssl-1.1` it'd be `needs_openssl11` and for `install_package "openssl-1.0` it'd be `needs_openssl10`.
That could be done in this PR or later, but we should not install openssl for CRuby 3.1+ on non-macOS.

eregon:
@mislav Could you do another review?
For me this looks good, modulo the comments above.

I think we also need to choose, either:
* This approach (install openssl for the user when system openssl is not compatible, this PR handles system openssl 3, but I think we should generalize it soon after). Pros is it's convenient and known to work relatively well. Cons is the Ruby built this way won't work with system packages depending on the system openssl (e.g., some databases do).
* Avoid installing openssl as much as possible, and rather automatically use workarounds like https://bugs.ruby-lang.org/issues/18658#note-4 (so build without openssl, then install a proper openssl gem). I'm not sure if workarounds like that exist for e.g. system openssl 1.1 and Ruby needs openssl 1.0 like in CRuby < 2.4 though. Pros is using the "system openssl", always.

share/ruby-build/3.2.0-dev
```diff
@@ -1,2 +1,2 @@
-install_package "openssl-3.0.2" "https://www.openssl.org/source/openssl-3.0.2.tar.gz#98e91ccead4d4756ae3c9cde5e09191a8e586d9f4d50838e7ec09d6411dfdb63" mac_openssl --if has_broken_mac_openssl
+install_package "openssl-3.0.2" "https://www.openssl.org/source/openssl-3.0.2.tar.gz#98e91ccead4d4756ae3c9cde5e09191a8e586d9f4d50838e7ec09d6411dfdb63" openssl --if needs_openssl
```
mislav, line 1: Ruby 3.2 seems to want OpenSSL 3, and `needs_openssl` will evaluate to true if system OpenSSL is v3. Is this desired? This will cause an additional OpenSSL to be compiled even though system OpenSSL is likely sufficient

bin/ruby-build
```diff
@@ -1081,6 +1081,16 @@ has_broken_mac_openssl() {
   [[ "$RUBY_CONFIGURE_OPTS" != *--with-openssl-dir=* ]]
 }
 
+has_openssl_3() {
```
mislav, line 1084: I don't think there is strong need to have `has_openssl_3` a separate functions, so let's inline this into `needs_openssl`

bin/ruby-build
```diff
@@ -1081,6 +1081,16 @@ has_broken_mac_openssl() {
...
+  local openssl_version="$(openssl version 2>/dev/null || printf '#include <openssl/opensslv.h>\nOPENSSL_VERSION_TEXT\n' | cc -xc -E - 2>/dev/null | tail -n1 | tr -d '"' || true)"
+  [[ $openssl_version = "OpenSSL 3"?* ]] &&
+  [[ "$RUBY_CONFIGURE_OPTS" != *--with-openssl-dir=* ]]
```
mislav, line 1087: This line should be the first condition so the rest of the logic short-circuits if `--with-openssl-dir` is set (that way we don't have to invoke `cc`, for example)

bin/ruby-build
```diff
@@ -1081,6 +1081,16 @@ has_broken_mac_openssl() {
   [[ "$RUBY_CONFIGURE_OPTS" != *--with-openssl-dir=* ]]
 }
 
+has_openssl_3() {
+  local openssl_version="$(openssl version 2>/dev/null || printf '#include <openssl/opensslv.h>\nOPENSSL_VERSION_TEXT\n' | cc -xc -E - 2>/dev/null | tail -n1 | tr -d '"' || true)"
```
mislav, line 1085: I don't think it's necessary to do so much output sanitization with `tail | tr`. Just pipe the version output to `grep -q "OpenSSL 3"` and see if it matches.

bin/ruby-build
```diff
@@ -1127,10 +1141,33 @@ build_package_mac_openssl() {
     MAKE_INSTALL_TARGET="install_sw install_ssldirs" build_package_standard "$@"
   fi
 
-  # Extract root certs from the system keychain in .pem format and rehash.
   local pem_file="$OPENSSLDIR/cert.pem"
-  security find-certificate -a -p /Library/Keychains/System.keychain > "$pem_file"
-  security find-certificate -a -p /System/Library/Keychains/SystemRootCertificates.keychain >> "$pem_file"
+  if is_mac; then
+    # Extract root certs from the system keychain in .pem format and rehash.
```
mislav, line 1146: I know a previous comment used to say "rehash", but where is the rehash step? If an explicit rehash step doesn't exist anymore, you can take the word out of the comment

bin/ruby-build
```diff
@@ -1127,10 +1141,33 @@ build_package_mac_openssl() {
...
+    security find-certificate -a -p /Library/Keychains/System.keychain > "$pem_file"
+    security find-certificate -a -p /System/Library/Keychains/SystemRootCertificates.keychain >> "$pem_file"
+  elif [ -e /etc/pki/tls/cert.pem ]; then # RedHat
+    # See https://github.com/rubygems/rubygems/issues/2415#issuecomment-509806259
+    rm -rf "$OPENSSLDIR/certs" "$pem_file"
+    ln -s /etc/pki/tls/certs "$OPENSSLDIR"
```
mislav, line 1152: Let's explicitly use the full path to the symlink file that will be created
```suggestion
    ln -s /etc/pki/tls/certs "$OPENSSLDIR/certs"
```

bin/ruby-build
```diff
@@ -1127,10 +1141,33 @@ build_package_mac_openssl() {
...
+    ln -s /etc/pki/tls/cert.pem "$pem_file"
+  elif [ -e /etc/ssl/certs/ca-certificates.crt ]; then # Debian
+    # See https://github.com/rubygems/rubygems/issues/2415#issuecomment-509806259
+    rm -rf "$OPENSSLDIR/certs" "$pem_file"
+    ln -s /etc/ssl/certs "$OPENSSLDIR"
```
mislav, line 1157:
```suggestion
    ln -s /etc/ssl/certs "$OPENSSLDIR/certs"
```

eregon:
@mislav Any thought on the above?

bin/ruby-build
```diff
@@ -1127,10 +1141,33 @@ build_package_mac_openssl() {
...
+    ln -s /etc/ssl/certs/ca-certificates.crt "$pem_file"
+  elif type openssl &>/dev/null; then
+    # symlink to the system openssl certs
+    local SYSTEM_OPENSSLDIR
+    SYSTEM_OPENSSLDIR=$(openssl version -d | sed -e 's/.*: "//' -e 's/"$//' || true)
```
mislav, line 1162: This post-processing with sed is hard to parse for a human like me. Let's use use `cut` for simplicity:
```suggestion
    SYSTEM_OPENSSLDIR=$(openssl version -d 2>/dev/null | cut -d'"' -f2)
```

mislav (changes requested):
Sorry for the delay! This is looking good. Some style suggestions and clarifying questions about Ruby 3.2

mislav (approved):
This looks great; thank you for your work and patience!

As far as I'm concerned, this can ship. One final thought I have about all this is forwards compatibility: it seems that with this change, we've kicked the can down the street (by redefining the semantics that were previously only around OpenSSL v1 to be OpenSSL v3-aware), but if anything significantly changes with how OSs package OpenSSL in the future (e.g. they start shipping OpenSSL v4), we'll have a set of problems again because we've hardcoded that `needs_openssl` in ruby-build means that it specifically needs OpenSSL v1 (and that it can't be Apple-patched OpenSSL/LibreSSL).

Ideally, in the future, `needs_openssl` would understand the precise version that a Ruby version needs. To a human, that version requirement is betrayed in the formula itself:
```sh
install_package "openssl-<ver>" "https://www.openssl.org/source/openssl-<ver>.tar.gz#<sha>" openssl --if <condition>
```
except that when the condition is evaluated, it doesn't have access to the `openssl-<ver>` bit. We could figure this out in subsequent changes.

yob:
For future travelers that find this and need to get ruby 3.0 installed on a system with openssl 3.0, here's one possible workaround:

https://gist.github.com/yob/08d53a003181aa0fcce9812b1b533870

tl;dr trick ruby-build into install ruby without openssl, manually download the openssl 3.0 gem using wget and install it.

Rather than install openssl 1.1 to compile ruby, is it possible to teach ruby-build an opt-in version of the above?

eregon:
> Rather than install openssl 1.1 to compile ruby, is it possible to teach ruby-build an opt-in version of the above?

Discussed in https://github.com/rbenv/ruby-build/pull/1974#issuecomment-1141436606 and https://bugs.ruby-lang.org/issues/18658#note-11.
So not really, because that technique does not work for Ruby < 2.6.
And also IMHO it feels quite hacky and not the right way to build Ruby (CRuby should really backport openssl 3 support to 2.7/3.0 IMHO: https://github.com/ruby/openssl/issues/517, TruffleRuby did that).

eregon:
I will try to rebase this and finish reviewing this PR. I'll squash all commits because otherwise it seems impossible to rebase with the recent openssl update.

eregon:
I've pushed a commit so the logic is intuitive and general:
> Install openssl whenever the system openssl version is not supported by that Ruby / the system openssl version cannot be used for that Ruby.

And for that I also made all supported versions explicit with `needs_openssl_101_111`, etc.
I tested locally and the version tests work well.

@mislav Could you review my Bash code added in ruby-build?
Then I think this should be ready to be merged and released.

bin/ruby-build
eregon: I decided to warn in such case that there is no system openssl.

eregon (approved):

eregon:
Actually, let me merge this and release it now, to make sure we don't get conflicts again.

mislav:
Thanks @znz for your hard work and @eregon for review, your patience, and additional fixes 🙇

PR #1980, opened 2022-06-03
Allow to download truffleruby-dev builds for Darwin-arm64 and Linux-aarch64

deepj:
@eregon there are native builds of TruffleRuby available in Graal build repo. So let's download them from there for dev at least


eregon:
Good idea, could you add Linux-aarch64 too?
And add both for https://github.com/rbenv/ruby-build/blob/master/share/ruby-build/truffleruby%2Bgraalvm-dev ?

There are some trade-offs, mentioned at https://github.com/oracle/truffleruby/issues/2181#issuecomment-1140946570:
> those builds lag a few days up to 1-2 weeks behind due to using the truffleruby import in GraalVM.

Maybe we should print a message on these architectures to warn about that?
Another way to deal with that would be to show the `RUBY_RELEASE_DATE` as part of `ruby -v` (or make it part of the engine version maybe), to make it easy to find how old a build is.
I think the second is probably best, I'll take care of that.

deepj:
@eregon done

eregon (approved):
Thanks!

PR #1987, opened 2022-06-28
Add picoruby-3.0.0

hasumikin:
This is the first time to send a PR of adding PicoRuby to ruby-build, so I show you how it's working on my local computer:

image

In addition to the above executables (`picoruby` and `picoirb`), `picorbc` which is the PicoRuby compiler executable will be installed, too.

This version number `3.0.0` corresponds to the mruby's version which means a VM code generated by the picorbc 3.0.0 runs on mruby virtual machine 3.0.0 like this:

image

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -688,6 +688,15 @@ build_package_mruby() {
   } >&4 2>&1
 }
 
+build_package_picoruby() {
+  { ./minirake
+    mkdir -p "$PREFIX_PATH"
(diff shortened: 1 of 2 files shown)


hsbt:
@hasumikin Thanks! I'm not sure why we add only `picoruby-3.0.0`. Is it the first stable version?

hasumikin:
@hsbt Yes! It is the only stable version. The reason that the number should be 3.0.0 is described in my first comment

Issue #1998, opened 2022-07-11
OpenSSL 1.1.1q failing to build


stack:
This is a fresh install of macOS 12.4 on a M1 Pro MacBook Pro. Attempting to install Ruby 3.1.2.

* Xcode 13.4.1
* Homebrew installed as of today (2022-07-11).
* ruby-build 20220710

The OpenSSL 1.1.1q fails to build in the tests:

```
Downloading openssl-1.1.1q.tar.gz...
-> https://dqw8nmjcqpjn7.cloudfront.net/d7939ce614029cdff0b6c20f0e2e5703158a489a72b2507b8bd51bf8c8fd10ca
Installing openssl-1.1.1q...

BUILD FAILED (macOS 12.4 using ruby-build 20220710)

Inspect or clean up the working tree at /var/folders/4y/m2qk7qcx099b6g_zt98mp3jh0000gn/T/ruby-build.20220711120126.51360.08iZ3t
Results logged to /var/folders/4y/m2qk7qcx099b6g_zt98mp3jh0000gn/T/ruby-build.20220711120126.51360.log

Last 10 log lines:
clang  -Iinclude -arch arm64 -O3 -Wall -D_REENTRANT -DZLIB -DZLIB_SHARED -DNDEBUG -I/Users/stephen/.rbenv/versions/3.1.2/include  -MMD -MF test/v3ext.d.tmp -MT test/v3ext.o -c -o test/v3ext.o test/v3ext.c
clang  -Iinclude -arch arm64 -O3 -Wall -D_REENTRANT -DZLIB -DZLIB_SHARED -DNDEBUG -I/Users/stephen/.rbenv/versions/3.1.2/include  -MMD -MF test/v3nametest.d.tmp -MT test/v3nametest.o -c -o test/v3nametest.o test/v3nametest.c
test/v3ext.c:201:24: error: implicitly declaring library function 'memcmp' with type 'int (const void *, const void *, unsigned long)' [-Werror,-Wimplicit-function-declaration]
        if (!TEST_true(memcmp(ip1->data, ip2->data, ip1->length) <= 0))
                       ^
test/v3ext.c:201:24: note: include the header <string.h> or explicitly provide a declaration for 'memcmp'
1 error generated.
make[1]: *** [test/v3ext.o] Error 1
make[1]: *** Waiting for unfinished jobs....
make: *** [all] Error 2
```

eregon:
[...]

PR #2000, opened 2022-07-12
Use OpenSSL 3.x with Ruby 3.1.x

hsbt:
Related with https://github.com/rbenv/ruby-build/issues/1998#issuecomment-1180698666

Ruby 3.1.x already support OpenSSL 3.0. There is no reason why Ruby 3.1 use OpenSSL 1.1.

### DIFF
--- a/share/ruby-build/3.1.0
+++ b/share/ruby-build/3.1.0
@@ -1,2 +1,2 @@
-install_package "openssl-1.1.1q" "https://www.openssl.org/source/openssl-1.1.1q.tar.gz#d7939ce614029cdff0b6c20f0e2e5703158a489a72b2507b8bd51bf8c8fd10ca" openssl --if needs_openssl_102_300
+install_package "openssl-3.0.5" "https://www.openssl.org/source/openssl-3.0.5.tar.gz#aa7d8d9bef71ad6525c55ba11e5f4397889ce49c2c9349dcea6d3e4f0b024a7a" openssl --if needs_openssl_102_300
 install_package "ruby-3.1.0" "https://cache.ruby-lang.org/pub/ruby/3.1/ruby-3.1.0.tar.gz#50a0504c6edcb4d61ce6b8cfdbddaa95707195fab0ecd7b5e92654b2a9412854" ldflags_dirs enable_shared standard verify_openssl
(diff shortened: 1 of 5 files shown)


eregon:
Agreed, thanks.

beporter:
One reason to use OpenSSL v1.1.x instead of 3.x is that they [deprecated](https://www.openssl.org/docs/man3.0/man7/migration_guide.html#Legacy-Algorithms) a [number of methods in 3.0.x](https://www.openssl.org/docs/man3.0/man7/OSSL_PROVIDER-legacy.html). `OpenSSL::Digest::MD4.hexdigest`, for example. If you have code that happens to use any of those now-deprecated methods, that code will start failing because of this change.

Summarized here: https://dev.to/wetterkrank/openssldigestdigesterror-when-using-md4-3o0i

Switching to OpenSSL 3.x, you should probably enable the [legacy providers](https://github.com/ruby/openssl/issues/500#issuecomment-1100660910) as well, if you want to maintain full compatibility with 1.1.x.

The workaround is to [manually install and compile OpenSSL 1.1.x](https://github.com/openssl/openssl/releases) first, then use `RUBY_CONFIGURE_OPTS="--with-openssl-dir=/INSTALL/PATH/HERE" rbenv install` to keep using 1.1.x.

eregon:
There was some discussion about this in https://github.com/rbenv/ruby-build/pull/2111
The bottom line is longer-term openssl 1.1.x won't be available, and is already not available on recent distros.
So anyway such code needs to be adapted to work on openssl 3.

Similarly about legacy providers, that won't work with a system-installed openssl 3, again Ruby code cannot rely on that if it wants to work on modern distros/Ruby/etc.

eregon:
> The workaround is to [manually install and compile OpenSSL 1.1.x](https://github.com/openssl/openssl/releases) first, then use `RUBY_CONFIGURE_OPTS="--with-openssl-dir=/INSTALL/PATH/HERE" rbenv install` to keep using 1.1.x.

And indeed that's the intended workaround when the calling code cannot be changed yet.

mislav:
@beporter Sorry that you've been bitten by a breaking change, and thanks for reporting here and blogging about your findings. 🙇

Issue #1970, opened 2022-05-06
Rust YJIT Support in `ruby-build`


maximecb:
First of all thank you for creating the incredibly useful tool that is ruby-build 🙏 

As you may know, we have been working on porting YJIT to Rust, and this has now been upstreamed: https://github.com/ruby/ruby/pull/5826

Rust YJIT requires `rustc` 1.60 to be installed to build in release mode, and `cargo` to build in dev/debug mode. As such, by default, CRuby will build without YJIT support. In order to build YJIT, `--enable-yjit` has to be specified when running `./configure`. More info on how to build YJIT can be found in the [YJIT readme](https://github.com/ruby/ruby/blob/master/doc/yjit/yjit.md#installation).

Given that `ruby-build` is the primary way that many people deploy Ruby and a central part of the Ruby ecosystem, we would really like to have the tool support building Rust YJIT. This would really help lower the bar for people to get YJIT running and could truly help YJIT see more use in the wild.

My colleague @byroot has told me that the `ruby-build` maintainers don't feel comfortable automatically installing the Rust toolchain. I understand your concerns with respect to security (and maybe the size of the download required). I was wondering if maybe we could explore other options.

Some ideas that come to mind:
- Possibly `ruby-build` could check if `rustc` >= 1.60 is already installed, and set `--enable-yjit` if present
- Maybe `ruby-build` could report to the user whether `ruby` is being built with or without YJIT support. If built without YJIT support, `ruby-build` could print a message explaining that `rustc` >= 1.60 is required to build YJIT.

I think we're likely going to freeze the `rustc` dependency at version 1.60 for some time so this requirement is going to be pretty stable. Happy to discuss pros and cons of various options and how we can help the `ruby-build` maintainers make this happen if you are open to it.

Tagging my colleagues @byroot @xrxr @noahgibbs.

koic:
JFYI, I have Rust YJIT installed by the following workaround command.

```sh
% rustc --version
rustc 1.60.0 (7737e0b5c 2022-04-04)
% RUBY_CONFIGURE_OPTS=--enable-yjit rbenv install 3.2.0-dev
(snip)

% ruby --yjit -ve 'p RubyVM::YJIT.enabled?'
ruby 3.2.0dev (2022-05-06T09:13:22Z master 67950a4c0a) +YJIT [x86_64-darwin19]
true
```

[...]

PR #2005, opened 2022-07-13
For Ruby 3.2.0-dev, install YJIT by default if rustc 1.60+ is available

noahgibbs:
This PR is a followup to https://github.com/rbenv/ruby-build/issues/1970. It adds build_package_enable_yjit.

If YJIT is explicitly enabled or disabled in RUBY_CONFIGURE_OPTS, enable_yjit should do nothing.
If YJIT is not specified and rustc is 1.6 or higher, it should print a message and enable YJIT.
If YJIT is not specified and rustc is not found or too old, it should print a "yjit disabled by default" message and do nothing else.

I've added the enable_yjit package to 3.2.0-dev only. In theory this could cause problems for 3.2.0 dev versions earlier than Rust-YJIT integration (19th April, CRuby SHA f90549). But it looks like 3.2.0-dev will only install latest master, not earlier random CRuby SHAs.

Rust-YJIT technically supports rustc 1.58+. But we said 1.60+ in the issue, so that's what I used here.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1256,6 +1256,29 @@ build_package_ldflags_dirs() {
   done
 }
 
+build_package_enable_yjit() {
+  local rustc_installed rustc_ver
+
+  rustc_installed=""
+  if type -p rustc &>/dev/null; then
+    rustc_ver=`rustc --version`
+    # Some kind of built-in bash comparison for dotted version strings would be awesome.
+    if [[ "${rustc_ver}" == *"rustc 1."[6789]* ]]; then
+      rustc_installed="1"
+    fi
+  fi
+
+  # If YJIT is explicitly specified, don't change anything
+  if [[ " ${RUBY_CONFIGURE_OPTS} " != *" --enable-yjit"* ]] && [[ " ${RUBY_CONFIGURE_OPTS} " != *" --disable-yjit"* ]]; then
+    if [[ "${rustc_installed}" == "1" ]]; then
+      echo "Building with YJIT by default because ${rustc_ver} is installed; add RUBY_CONFIGURE_OPTS='--disable-yjit' to disable explicitly"
+      package_option ruby configure --enable-yjit
+    else
+      echo "rustc 1.60+ is not installed, YJIT will not be built; if you want to use YJIT, install rustc 1.60+ and rerun this command."
+    fi
+  fi
+}
+
 build_package_enable_shared() {
   if [[ " ${RUBY_CONFIGURE_OPTS} " != *" --disable-shared"* ]]; then
     package_option ruby configure --enable-shared
--- a/share/ruby-build/3.2.0-dev
+++ b/share/ruby-build/3.2.0-dev
@@ -1,2 +1,2 @@
 install_package "openssl-3.0.5" "https://www.openssl.org/source/openssl-3.0.5.tar.gz#aa7d8d9bef71ad6525c55ba11e5f4397889ce49c2c9349dcea6d3e4f0b024a7a" openssl --if needs_openssl_102_300
-install_git "ruby-master" "https://github.com/ruby/ruby.git" "master" ldflags_dirs autoconf standard_build standard_install_with_bundled_gems verify_openssl
+install_git "ruby-master" "https://github.com/ruby/ruby.git" "master" ldflags_dirs autoconf enable_yjit standard_build standard_install_with_bundled_gems verify_openssl


bin/ruby-build
```diff
@@ -1256,6 +1256,29 @@ build_package_ldflags_dirs() {
...
+    # Some kind of built-in bash comparison for dotted version strings would be awesome.
+    if [[ " ${rustc_ver} " == "1."[6789]* ]]; then
+      rustc_installed="1"
+    fi
+  fi
+
+  # If YJIT is explicitly specified, don't change anything
+  if [[ " ${RUBY_CONFIGURE_OPTS} " != *" --enable-yjit"* ]]; then
+    if [[ -z "${rustc_installed}" ]]; then
+      echo "rustc ${rustc_ver} is installed: enabling yjit by default"
```
maximecb, line 1274: I think I would prefer that we say "building CRuby with YJIT" or "building CRuby without YJIT" instead of enabling/disabling, but open to arguments either way.

maximecb:
Looks pretty good Noah. Thanks for putting this together.

bin/ruby-build
```diff
@@ -1256,6 +1256,29 @@ build_package_ldflags_dirs() {
...
 }
 
+build_package_enable_yjit() {
+  local rustc_installed rustc_ver
+
+  rustc_installed=""
+  if which rustc; then
+    rustc_ver=`rustc --version | cut -c7-10`
+    # Some kind of built-in bash comparison for dotted version strings would be awesome.
+    if [[ " ${rustc_ver} " == "1."[6789]* ]]; then
```
eregon, line 1266:
```suggestion
    if [[ "${rustc_ver}" == "1."[6789]* ]]; then
```
Otherwise the extra spaces probably prevent matching, no?

bin/ruby-build
```diff
@@ -1256,6 +1256,29 @@ build_package_ldflags_dirs() {
   done
 }
 
+build_package_enable_yjit() {
+  local rustc_installed rustc_ver
+
+  rustc_installed=""
+  if which rustc; then
+    rustc_ver=`rustc --version | cut -c7-10`
```
eregon, line 1264: Likely needs a `2>/dev/null || true` or so, like other checks in this file, so there is no extra output if missing.
Could you test what happens if rustc is not installed? One easy way is to change to e.g. `rustcNOTEXIST` here.

Re the `cut -c7-10` there is a builtin bash way to do that, or it could be done by the matching below which seems safer than hardcoded character offsets.

bin/ruby-build
eregon: Sounds good to me.

I think we should also mention how to disable in this message, since e.g. compilation might fail due to a badly setup/broken rust toolchain, and I think we wouldn't want users to open issues/discussions just to ask that.

bin/ruby-build
```diff
@@ -1256,6 +1256,29 @@ build_package_ldflags_dirs() {
...
+    fi
+  fi
+
+  # If YJIT is explicitly specified, don't change anything
+  if [[ " ${RUBY_CONFIGURE_OPTS} " != *" --enable-yjit"* ]]; then
+    if [[ -z "${rustc_installed}" ]]; then
+      echo "rustc ${rustc_ver} is installed: enabling yjit by default"
+      package_option ruby configure --enable-yjit
+    else
+      echo "rustc 1.60+ is not installed; disabling yjit by default"
```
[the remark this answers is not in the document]
eregon: I think that may be too much, notably because that page seems very opinionated (e.g., it does not even mention using a package manager, which has advantages, especially if one does not plan to develop in rust but just build rust software).
And in general if people want yjit I'm confident they can find how to install rustc, OTOH for users which do not mind about YJIT/do not well about installing extra things then it seems better to not lead to yet another step of installing packages which could fail, especially a custom one like this with curl.

I'd phrase it like:
```
rustc 1.60+ is not installed, YJIT will not be built; if you want to use YJIT, install rustc 1.60+ and rerun this command.
```

eregon (commented):
> If YJIT is explicitly enabled or disabled in RUBY_CONFIGURE_OPTS, enable_yjit should do nothing.

I don't see logic to do nothing if disabled, or does that work because it's disabled via `--enable-yit=false` and not `--disable-yjit`?

bin/ruby-build
noahgibbs: Ah, I had pushed a less-than-latest version without realising. D'oh! Fixed now.

bin/ruby-build
noahgibbs: Okay. Updated the language.

bin/ruby-build
noahgibbs: I don't think there's an easy built-in bash way to do it, nor a built-in bash comparison for dotted version strings. There's "sort -V" but it's not as broadly available as we'd want. I'll update the matching to do it that way.

eregon: https://www.baeldung.com/linux/bash-substring#3-using-bashs-substring-expansion is what I meant for the builtin way to substring with offsets.
But yes I think extending the matching to be like `"${rustc_ver}" == "rustc 1."[6789]*` is the way to go.

bin/ruby-build
```diff
@@ -1256,6 +1256,29 @@ build_package_ldflags_dirs() {
   done
 }
 
+build_package_enable_yjit() {
+  local rustc_installed rustc_ver
+
+  rustc_installed=""
+  if type rustc &>/dev/null; then
```
eregon, line 1263: Could you use `type -p rustc`? That's what's used in other places.

noahgibbs: Sure. I had actually copied a different location that did *not* use -p (search for "type shasum"). But I'm not opposed to that. I couldn't find documentation on type easily since it's a bash builtin and the man page is horrible for searching. But it's easy enough to add.

eregon (approved):
Thank you, this looks good to me.

Could you do a test locally with rustc 1.60+ installed and without rustc (fake by checking another command)?
To ensure there is no extra output and it works as expected.

eregon:
I tested as well:
```
(with older rustc)
ruby-build 3.2.0-dev ~/.rubies/ruby-3.2.0-dev
Cloning https://github.com/ruby/ruby.git...
Installing ruby-master...
rustc 1.60+ is not installed, YJIT will not be built; if you want to use YJIT, install rustc 1.60+ and rerun this command.
Installed ruby-master to /home/eregon/.rubies/ruby-3.2.0-dev


ruby-build 3.2.0-dev ~/.rubies/ruby-3.2.0-dev
Cloning https://github.com/ruby/ruby.git...
Installing ruby-master...
Building with YJIT by default because rustc 1.62.0 (a8314ef7d 2022-06-27) is installed; add RUBY_CONFIGURE_OPTS='--disable-yjit' to disable explicitly
Installed ruby-master to /home/eregon/.rubies/ruby-3.2.0-dev


RUBY_CONFIGURE_OPTS='--disable-yjit' ruby-build 3.2.0-dev ~/.rubies/ruby-3.2.0-dev
Cloning https://github.com/ruby/ruby.git...
Installing ruby-master...
Installed ruby-master to /home/eregon/.rubies/ruby-3.2.0-dev
```
And `ruby --yjit -e0` works as expected based on the enabling.

It's not fully ideal that the message is shown after `Installing ruby-master...` because the long pause is then with a message not ending in `...` or "wait it might take a while", but I think it's good enough.

hsbt:
@noahgibbs @maximecb Does YJIT support the native build of Apple Silicon? I failed to build with the current master of ruby-build(and also failed with ruby master branch) on arm64 arch.

Should We disabled this when `arch` returned `arm64` if YJIT didn't support Apple Silicon yet?

noahgibbs:
@hsbt That's a really good point! While we're working on it actively, we do *not* support it yet. I'll fix that and submit a PR for it today.

hsbt:
FYI: I update requirements of ruby-build on our wiki about rustc.

https://github.com/rbenv/ruby-build/wiki/Home/67bb3b2746b8de0426c8e7ee00919a1f974df02d

maximecb:
Thank you @hsbt 🙏

PR #2010, opened 2022-07-14
Don't auto-enable YJIT except on x86_64

noahgibbs:
As @hsbt points out on my previous PR (https://github.com/rbenv/ruby-build/pull/2005#issuecomment-1184023153), we should only enable YJIT on x86_64, not on arm64 like Mac M1s.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1268,7 +1268,7 @@ build_package_enable_yjit() {
   local rustc_ver="$(rustc --version 2>/dev/null)"
   [ -n "$rustc_ver" ] || return 0
   # Some kind of built-in bash comparison for dotted version strings would be awesome.
-  if [[ "${rustc_ver})" == *"rustc 1."[6789]* ]]; then
+  if [[ "${rustc_ver}" == *"rustc 1."[6789]* ]]; then
     echo "Building with YJIT by default because ${rustc_ver} is installed; add RUBY_CONFIGURE_OPTS='--disable-yjit' to disable explicitly" >&3
     package_option ruby configure --enable-yjit
   else


hsbt (approved):
👍

bin/ruby-build
```diff
@@ -1268,8 +1268,13 @@ build_package_enable_yjit() {
     fi
   fi
 
-  # If YJIT is explicitly specified, don't change anything
-  if [[ " ${RUBY_CONFIGURE_OPTS} " != *" --enable-yjit"* ]] && [[ " ${RUBY_CONFIGURE_OPTS} " != *" --disable-yjit"* ]]; then
+  x86_arch=""
+  if [[ "$(uname -m)" == "x86_64" ]]; then
+    x86_arch="1"
+  fi
```
eregon, line 1274: Isn't it much simpler to just `return`?
Like `[[ "$(uname -m)" != "x86_64" ]] && return 0` (or with a `if`)

eregon: Might also want to check + return for `RUBY_CONFIGURE_OPTS`, so we don't have nested `if` at all.

BTW, those checks should be done before we check `rustc`, so we don't take extra time if anyway we're not going to use the result of `rustc --version`

mislav (changes requested):
+1 to @eregon's suggestions about short-circuiting the `build_package_enable_yjit` function.

While we're making changes to `build_package_enable_yjit`, could we also change the `echo` lines to print to stderr in the terminal instead of stdout (since the latter will be redirected to a log file)? You can do that with `echo ... >&3` (file descriptor 3 is where the preserve the original stderr)

bin/ruby-build
noahgibbs: Okay. I've restructured this.

noahgibbs:
Okay. I've restructured build_package_enable_yjit to short-circuit and it should now be echoing to fd 3.

bin/ruby-build
```diff
@@ -1257,30 +1257,22 @@ build_package_ldflags_dirs() {
...
-  rustc_installed=""
-  if type -p rustc &>/dev/null; then
-    rustc_ver=`rustc --version`
-    # Some kind of built-in bash comparison for dotted version strings would be awesome.
-    if [[ "${rustc_ver}" == *"rustc 1."[6789]* ]]; then
-      rustc_installed="1"
-    fi
-  fi
+  # If YJIT is explicitly enabled or disabled, don't change the config
+  ([[ " ${RUBY_CONFIGURE_OPTS} " == *" --enable-yjit"* ]] || [[ " ${RUBY_CONFIGURE_OPTS} " == *" --disable-yjit"* ]]) && return 0;
```
mislav, line 1261: The `(...)` expression in bash spawns a subshell; let's avoid that since it's not necessary.
```suggestion
  if [[ " ${RUBY_CONFIGURE_OPTS} " == *" --enable-yjit"* || " ${RUBY_CONFIGURE_OPTS} " == *" --disable-yjit"* ]]; then
    return 0
  fi
```

bin/ruby-build
```diff
@@ -1257,30 +1257,22 @@ build_package_ldflags_dirs() {
...
-    if [[ "${rustc_installed}" == "1" ]]; then
-      echo "Building with YJIT by default because ${rustc_ver} is installed; add RUBY_CONFIGURE_OPTS='--disable-yjit' to disable explicitly"
-      package_option ruby configure --enable-yjit
-    else
-      echo "rustc 1.60+ is not installed, YJIT will not be built; if you want to use YJIT, install rustc 1.60+ and rerun this command."
-    fi
+  local rustc_ver
+
+  (type -p rustc &>/dev/null) || return 0;
+  rustc_ver=`rustc --version`
```
mislav, line 1269: This bit could be simplified to again avoid `(...)`
```suggestion
  local rustc_ver="$(rustc --version 2>/dev/null)"
  [ -n "$rustc_ver" ] || return 0
```

bin/ruby-build
```diff
@@ -1257,30 +1257,22 @@ build_package_ldflags_dirs() {
...
-  fi
+  # If YJIT is explicitly enabled or disabled, don't change the config
+  ([[ " ${RUBY_CONFIGURE_OPTS} " == *" --enable-yjit"* ]] || [[ " ${RUBY_CONFIGURE_OPTS} " == *" --disable-yjit"* ]]) && return 0;
 
-  x86_arch=""
-  if [[ "$(uname -m)" == "x86_64" ]]; then
-    x86_arch="1"
-  fi
+  # If we aren't on x86_64, don't enable YJIT
+  [[ "$(uname -m)" != "x86_64" ]] && return 0;
```
mislav, line 1264: Might read better:
```suggestion
  [ "$(uname -m)" = "x86_64" ] || return 0
```

mislav (changes requested):
Thanks!

bin/ruby-build
```diff
@@ -1257,25 +1257,22 @@ build_package_ldflags_dirs() {
...
-    else
-      echo "rustc 1.60+ is not installed, YJIT will not be built; if you want to use YJIT, install rustc 1.60+ and rerun this command."
-    fi
+  # If we aren't on x86_64, don't enable YJIT
+  [ "$(uname -m)" = "x86_64" ] || return 0
+
+  local rustc_ver="$(rustc --version 2>/dev/null)"
+  [ -n "$rustc_ver" ] || return 0
+  # Some kind of built-in bash comparison for dotted version strings would be awesome.
+  if [[ "${rustc_ver})" == *"rustc 1."[6789]* ]]; then
```
eregon, line 1271:
```suggestion
  if [[ "${rustc_ver}" == *"rustc 1."[6789]* ]]; then
```
There is an extra paren here

PR #2011, opened 2022-07-15
Apply patch for implicitly declaration error of OpenSSL 1.1.1q

hsbt:
Resolve https://github.com/rbenv/ruby-build/discussions/2009#discussioncomment-3146585

@mislav @eregon Can you review this? I'm not sure which location is the best for this workaround.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1163,7 +1163,21 @@ build_package_openssl() {
 
   if [[ "$1" == openssl-1.1.1q ]]; then
     # https://github.com/rbenv/ruby-build/discussions/2009#discussioncomment-3146585
-    curl -s https://github.com/openssl/openssl/commit/f9e578e720bb35228948564192adbe3bc503d5fb.diff | patch -p1
+    # https://github.com/openssl/openssl/commit/f9e578e720bb35228948564192adbe3bc503d5fb.diff
+    patch -p1 <<EOF
+diff --git a/test/v3ext.c b/test/v3ext.c
+index 926f3884b13..a8ab64b2714 100644
+--- a/test/v3ext.c
++++ b/test/v3ext.c
+@@ -8,6 +8,7 @@
+  */
+
+ #include <stdio.h>
++#include <string.h>
+ #include <openssl/x509.h>
+ #include <openssl/x509v3.h>
+ #include <openssl/pem.h>
+EOF
   fi
 
   # Compile a shared lib with zlib dynamically linked.
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1161,6 +1161,11 @@ build_package_openssl() {
   local nokerberos
   [[ "$1" != openssl-1.0.* ]] || nokerberos=1
 
+  if [[ "$1" == openssl-1.1.1q ]]; then
+    # https://github.com/rbenv/ruby-build/discussions/2009#discussioncomment-3146585
+    curl -s https://github.com/openssl/openssl/commit/f9e578e720bb35228948564192adbe3bc503d5fb.diff | patch -p1
+  fi
+
   # Compile a shared lib with zlib dynamically linked.
   package_option openssl configure --openssldir="$OPENSSLDIR" zlib-dynamic no-ssl3 shared ${nokerberos:+no-ssl2 no-krb5}
 


bin/ruby-build
```diff
@@ -1161,6 +1161,11 @@ build_package_openssl() {
   local nokerberos
   [[ "$1" != openssl-1.0.* ]] || nokerberos=1
 
+  if [[ "$1" == openssl-1.1.1q ]]; then
+    # https://github.com/rbenv/ruby-build/discussions/2009#discussioncomment-3146585
+    curl -s https://github.com/openssl/openssl/commit/f9e578e720bb35228948564192adbe3bc503d5fb.diff | patch -p1
```
eregon, line 1166: I don't think we can assume `curl` is installed, especially on macOS.
Given the patch is so small, could you just copy it inline here in a bash heredoc? (and add a comment with the link to the commit above it).

If you're on macOS please also test it to ensure it works as expected (I'm on Linux and my system openssl is 1.1 so I can't test easily).

eregon (commented):
Thank you for creating a PR for this :)

bin/ruby-build
VxJasonxV: `curl` has been available on all MacOS systems by default since at least 2007, most of if not the whole OS X lineage. I was adept with `wget` having come from a Linux desktop before then, and had to learn `curl` equivalents for everything I used `wget` for at a previous job.

Hand-patching this change worked on my machine, for the record;

```
➜  ~ rbenv install 2.7.6
[…]
test/v3ext.c:201:24: error: implicitly declaring library function 'memcmp' with type 'int (const void *, const void *, unsigned long)' [-Werror,-Wimplicit-function-declaration]
        if (!TEST_true(memcmp(ip1->data, ip2->data, ip1->length) <= 0))
                       ^
test/v3ext.c:201:24: note: include the header <string.h> or explicitly provide a declaration for 'memcmp'
1 error generated.
make[1]: *** [test/v3ext.o] Error 1
…

➜  ~ head -n 1169 /usr/local/Cellar/ruby-build/20220713/bin/ruby-build |tail -n 8

  if [[ "$1" == openssl-1.1.1q ]]; then
    # https://github.com/rbenv/ruby-build/discussions/2009#discussioncomment-3146585
    curl -s https://github.com/openssl/openssl/commit/f9e578e720bb35228948564192adbe3bc503d5fb.diff | patch -p1
  fi

➜  ~ rbenv install 2.7.6
Downloading openssl-1.1.1q.tar.gz...
-> https://dqw8nmjcqpjn7.cloudfront.net/d7939ce614029cdff0b6c20f0e2e5703158a489a72b2507b8bd51bf8c8fd10ca
Installing openssl-1.1.1q...
patching file test/v3ext.c
Installed openssl-1.1.1q to /Users/jsalaz/.rbenv/versions/2.7.6

Downloading ruby-2.7.6.tar.bz2...
-> https://cache.ruby-lang.org/pub/ruby/2.7/ruby-2.7.6.tar.bz2
Installing ruby-2.7.6...
Installed ruby-2.7.6 to /Users/jsalaz/.rbenv/versions/2.7.6

Fetching bundler-2.3.18.gem
…
```

Using macOS 12.4

```
➜  ~ curl --version
curl 7.79.1 (x86_64-apple-darwin21.0) libcurl/7.79.1 (SecureTransport) LibreSSL/3.3.6 zlib/1.2.11 nghttp2/1.45.1
Release-Date: 2021-09-22
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp 
Features: alt-svc AsynchDNS GSS-API HSTS HTTP2 HTTPS-proxy IPv6 Kerberos Largefile libz MultiSSL NTLM NTLM_WB SPNEGO SSL UnixSockets
```

eregon: Ah, maybe I mixed with `wget` not being available on macOS. Either way since ruby-build supports having either curl or wget available, it's not safe to rely on `curl` being always available.
We don't need to do any network request here, that's the simplest solution to simply include the patch in the code as a herdoc.

hsbt: Thanks for your comments. I'll try to use heredoc instead of curl download.

eregon:
I confirm this works nicely and it fails without:
```
$ ruby-build 2.7.1 ~/.rubies/ruby-2.7.1        
Downloading openssl-1.1.1q.tar.gz...
-> https://dqw8nmjcqpjn7.cloudfront.net/d7939ce614029cdff0b6c20f0e2e5703158a489a72b2507b8bd51bf8c8fd10ca
Installing openssl-1.1.1q...
patching file test/v3ext.c
Installed openssl-1.1.1q to /home/eregon/.rubies/ruby-2.7.1
...
```

(Since I'm on Linux I needed `OPENSSL_CFLAGS=-Werror=implicit-function-declaration` to repro the original error,
and since I have a 1.1 system openssl I changed the definition of 2.7.1 locally to remove the `--if`).

eregon (approved):

hsbt:
Thanks a lot!

eregon:
Released as v20220721

PR #2028, opened 2022-08-29
Re-enable YJIT with arm64

hsbt:
https://github.com/ruby/ruby/pull/6289 support arm64 arch with YJIT now.

/cc @noahgibbs

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1279,9 +1279,6 @@ build_package_enable_yjit() {
     return 0
   fi
 
-  # If we aren't on x86_64, don't enable YJIT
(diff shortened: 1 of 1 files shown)


noahgibbs:
I'm a little worried by just deleting the check. Does ruby-build support architectures other than x86_64 and arm64? Because YJIT only supports those two.

k0kubun:
Me too. i386 and arm (32bit) might be still common enough to have ruby-build users.

hsbt:
Maked sense. How about https://github.com/rbenv/ruby-build/pull/2029 ?

PR #2029, opened 2022-08-29
Enable YJIT only x86_64 and arm64

hsbt:
Followed up https://github.com/rbenv/ruby-build/pull/2028

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1280,7 +1280,7 @@ build_package_enable_yjit() {
   fi
 
   # If we aren't on x86_64 and arm64, don't enable YJIT
-  [ "$(uname -m)" = "x86_64" ] || [ "$(uname -m)" = "arm64" ] || return 0
+  [ "$(uname -m)" = "x86_64" -o "$(uname -m)" = "arm64" ] || return 0
 
   local rustc_ver="$(rustc --version 2>/dev/null)"
   [ -n "$rustc_ver" ] || return 0
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1279,6 +1279,9 @@ build_package_enable_yjit() {
     return 0
   fi
 
+  # If we aren't on x86_64 and arm64, don't enable YJIT
+  [ "$(uname -m)" = "x86_64" ] || [ "$(uname -m)" = "arm64" ] || return 0
+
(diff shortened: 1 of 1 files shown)


bin/ruby-build
```diff
@@ -1279,6 +1279,9 @@ build_package_enable_yjit() {
     return 0
   fi
 
+  # If we aren't on x86_64 and arm64, don't enable YJIT
+  [ "$(uname -m)" = "x86_64" ] || [ "$(uname -m)" = "arm64" ] || return 0
```
k0kubun, line 1283: nitpick:

```suggestion
  [ "$(uname -m)" = "x86_64" -o "$(uname -m)" = "arm64" ] || return 0
```

k0kubun (approved):
Thanks :+1:

noahgibbs:
Looks good. Thank you!

bin/ruby-build
```diff
@@ -1279,6 +1279,9 @@ build_package_enable_yjit() {
     return 0
   fi
 
+  # If we aren't on x86_64 and arm64, don't enable YJIT
+  [ "$(uname -m)" = "x86_64" -o "$(uname -m)" = "arm64" ] || return 0
```
eregon, line 1283: `uname -m` is only arm64 on darwin, on linux it's aarch64 (the proper name for it :p)

eregon: Fixed in https://github.com/rbenv/ruby-build/commit/1f6b94cc9392ada220be3cd4d104b9e9ee883517

noahgibbs: YJIT should work with at least Mac M1 and Graviton Linux. We test on both regularly.

eregon: Yes: https://github.com/ruby/ruby/pull/6289/commits/b216672cfd35131d0f3890280a55b8c9c9375e36
And it's even tested on linux-aarch64: https://github.com/ruby/ruby/pull/6289/files#diff-62587956f943bb2503db7bc6dd27d0d888074a1c0ecaab3f570ad611aff0f7bb

hsbt: Thanks. I also confirm to work YJIT with AWS g series.

```
$ ruby -v --jit
ruby 3.2.0dev (2022-08-30T09:46:06Z master b98653f766) +YJIT [aarch64-linux]
```

PR #2030, opened 2022-09-01
Relax rustc version to 1.58.1

hsbt:
@noahgibbs /cc @k0kubun 

YJIT supported rustc-1.58.1+. I suggest relaxing its version on ruby-build because apt package manager of Ubuntu 20.04 and 22.04 provides rustc-1.59.0.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1285,8 +1285,8 @@ build_package_enable_yjit() {
 
   local rustc_ver="$(rustc --version 2>/dev/null)"
   [ -n "$rustc_ver" ] || return 0
-  # Some kind of built-in bash comparison for dotted version strings would be awesome.
-  if [[ "${rustc_ver}" == *"rustc 1."[6789]* ]]; then
+  # YJIT supports Rust 1.58.1 and later
+  if [[ "${rustc_ver}" =~ ^(rustc 1\.)(58\.1|59*|[6789]*) ]]; then
     echo "Building with YJIT by default because ${rustc_ver} is installed; add RUBY_CONFIGURE_OPTS='--disable-yjit' to disable explicitly" >&3
     package_option ruby configure --enable-yjit
   else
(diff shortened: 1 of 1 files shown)


bin/ruby-build
```diff
@@ -1285,8 +1285,8 @@ build_package_enable_yjit() {
 
   local rustc_ver="$(rustc --version 2>/dev/null)"
   [ -n "$rustc_ver" ] || return 0
-  # Some kind of built-in bash comparison for dotted version strings would be awesome.
-  if [[ "${rustc_ver}" == *"rustc 1."[6789]* ]]; then
+  # YJIT supports Rust 1.58.1 and later
+  if [[ "${rustc_ver}" =~ ^(rustc 1\.)(58\.1|59*|[6789]*) ]]; then
```
k0kubun, line 1289: It seems like `*` is not glob but a normal `*` of regex.
```
$ [[ "rustc 1.58.0 (123abcdef 2022-08-31)" =~ ^(rustc 1\.)(58\.1|59*|[6789]*) ]]; echo $?
0
$ [[ "rustc 1.58.1 (123abcdef 2022-08-31)" =~ ^(rustc 1\.)(58\.1|59*|[6789]*) ]]; echo $?
0
```

It should probably be something like:

```suggestion
  if [[ "${rustc_ver}" =~ ^rustc\ 1\.(58\.1|59\.|[6789]\d\.) ]]; then
```

(Ignoring >= 2 and >= 1.100 for now)

bin/ruby-build
```diff
@@ -1285,8 +1285,8 @@ build_package_enable_yjit() {
 
   local rustc_ver="$(rustc --version 2>/dev/null)"
   [ -n "$rustc_ver" ] || return 0
-  # Some kind of built-in bash comparison for dotted version strings would be awesome.
-  if [[ "${rustc_ver}" == *"rustc 1."[6789]* ]]; then
+  # YJIT supports Rust 1.58.1 and later
+  if [[ "${rustc_ver}" =~ ^rustc\ 1\.(58\.1|59\.|[6789]\d\.) ]]; then
```
eregon, line 1289: Should it accept 1.58.2 till 1.58.9?

noahgibbs: I don't think there are releases with those numbers: https://github.com/rust-lang/rust/blob/master/RELEASES.md

PR #2042, opened 2022-09-11
ci(lint): add shell linter - Differential ShellCheck

jamacku:
Differential ShellCheck is a GitHub action that performs **differential ShellCheck scans** on shell scripts changed via PR and reports results directly in PR. The beauty of differential scans is that you will get **reports only about newly added defects**. 

Since this repository has few shell scripts, I think you might find some value in having a shell linter. Differential ShellCheck action is able to produce reports in SARIF format. GitHub understands this format and is able to display it nicely as a PR comment, and on the `Files Changed` tab, please see below.

image

image

Documentation is available at [@redhat-plumbers-in-action/differential-shellcheck](https://github.com/redhat-plumbers-in-action/differential-shellcheck#readme). Let me know If you are missing some feature or setting. I'm always happy to extend functionality.

### DIFF
--- a/.github/workflows/differential-shellcheck.yml
+++ b/.github/workflows/differential-shellcheck.yml
@@ -1,0 +1,26 @@
+# Doc: https://github.com/redhat-plumbers-in-action/differential-shellcheck#usage
+---
+
+name: Differential ShellCheck
+on:
+  pull_request:
+    branches: [master]
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Repository checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Differential ShellCheck
+        uses: redhat-plumbers-in-action/differential-shellcheck@v3
+        with:
+          severity: warning
+          token: ${{ secrets.GITHUB_TOKEN }}


.github/workflows/differential-shellcheck.yml
```diff
@@ -0,0 +1,30 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    permissions:
+      security-events: write
```
eregon, line 17: Why is this added? It seems unnecessary, this is not a private repository

.github/workflows/differential-shellcheck.yml
```diff
@@ -0,0 +1,30 @@
...
+      pull-requests: write
```
eregon, line 18: Is this needed?

eregon:
This sounds quite nice :)

.github/workflows/differential-shellcheck.yml
jamacku: No, it's not. It is used for comments pointing to defects.

image

It can be removed, and defects will still be marked in the Files changed tab. Also, comments aren't created when there are many defects (to keep the conversation "clean").

.github/workflows/differential-shellcheck.yml
jamacku: I usually use `permissions` to set the minimum required permissions for the job. And to be honest, I'm not sure what will happen when you leave out `security-events: write`, I don't know what the default permissions are for the forks. I'm going to test that.

jamacku: Yes, you are right. It doesn't have to be set.
- https://github.com/jamacku/ruby-build/pull/1

Would you prefer the removal of `permissions`?

eregon: Let's remove `security-events: write`, but keep `pull-requests: write`

.github/workflows/differential-shellcheck.yml
eregon: The comments seem useful, let's keep `pull-requests: write`

eregon: Although, if the feedback is sufficiently visible without `pull-requests: write` then that seems worth removing so we only have `contents: read` which is safe.

.github/workflows/differential-shellcheck.yml
jamacku: Removed.

.github/workflows/differential-shellcheck.yml
jamacku: Removed, @eregon PTAL

.github/workflows/differential-shellcheck.yml
```diff
@@ -0,0 +1,26 @@
...
+      - name: Repository checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Differential ShellCheck
+        uses: redhat-plumbers-in-action/differential-shellcheck@v3
+        with:
+          severity: warning
+          token: ${{ secrets.GITHUB_TOKEN }}
```
mislav, line 26: Since we're not uploading findings as comments anymore, can we drop the token argument? Is it used for anything in read-only mode?

mislav (approved):
Let's give this a shot! Thanks for the recommendation

.github/workflows/differential-shellcheck.yml
jamacku: My [previous comment was not correct](https://github.com/rbenv/ruby-build/pull/2042#discussion_r969460120). Permissions don't have any effect on Pull-Request review comments. I have tested different configurations of permissions. Sorry about that, It was a new thing for me.

Commenting and uploading of sarif is handled via GitHub bot, and I don't have any control over it. Also, I couldn't find sufficient documentation about how it works or how to configure the bot. The only thing I was able to find was:
- blog post - https://github.blog/changelog/2022-06-02-users-can-view-and-comment-on-code-scanning-alerts-on-the-conversation-tab-in-a-pull-request/
- general documentation of feature - https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/triaging-code-scanning-alerts-in-pull-requests#viewing-an-alert-on-your-pull-request

I have already raised a question on GitHub/community, but so far, no answer:
- https://github.com/community/community/discussions/33035

It is possible to run `differential-shellcheck` without a token, but you lose all SARIF-related features (comments and warnings on the Files changed tab). However, results would still be accessible upon inspecting console logs of the failed job.

jamacku:
Thank you for merging. If you have any issues, feel free to contact me.

Would you be interested in shell linter in other repositories with shell scripts?

mislav:
Feel free to propose adding this to `rbenv/rbenv` too, but it's not necessary for other repos in this org; thank you.

PR #2044, opened 2022-09-13
macOS: Use GMP from homebrew

casperisfine:
MRI's BigInt use GMP under the hood if available, and if not fallbacks to an homemade implementation with much worse performance.

Without GMP:
```
>> Benchmark.realtime { Integer('1' * (10 ** 7)) }
=> 13.80743500002427
```

With GMP
```
>> Benchmark.realtime { Integer('1' * (10 ** 7)) }
=> 0.4098639999865554
```

macOS not being a common production platform, it's not a huge deal but it would still preferable to compile with `gmp.h` if present.

I'd also suggest to print a warning if compiling on macOS and gmp isn't installed, but I don't know if it's desirable.

cc @eregon @hsbt

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -580,6 +580,7 @@ build_package_standard_build() {
 
   if [ "$package_var_name" = "RUBY" ]; then
       use_homebrew_readline || use_freebsd_pkg ||true
+      use_homebrew_gmp || true
   fi
 
   ( if [ "${CFLAGS+defined}" ] || [ "${!PACKAGE_CFLAGS+defined}" ]; then
@@ -1042,6 +1043,16 @@ use_homebrew_yaml() {
   if [ -d "$libdir" ]; then
     echo "ruby-build: using libyaml from homebrew"
     package_option ruby configure --with-libyaml-dir="$libdir"
+  else
+    return 1
+  fi
+}
+
+use_homebrew_gmp() {
+  local libdir="$(brew --prefix gmp 2>/dev/null || true)"
+  if [ -d "$libdir" ]; then
+    echo "ruby-build: using gmp from homebrew"
+    package_option ruby configure --with-gmp-dir="$libdir"
   else
     return 1
   fi
--- a/script/test
+++ b/script/test
@@ -1,4 +1,4 @@
 #!/bin/bash
 set -e
 
-bats ${CI:+--tap} test
+bats ${CI:+--tap} test "$@"
--- a/test/build.bats
+++ b/test/build.bats
@@ -186,7 +186,7 @@ OUT
   mkdir -p "$brew_libdir"
 
   stub uname '-s : echo Linux'
-  stub brew "--prefix libyaml : echo '$brew_libdir'" false
+  stub brew "--prefix libyaml : echo '$brew_libdir'" false false
   stub_make_install
 
   install_fixture definitions/needs-yaml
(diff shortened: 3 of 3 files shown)


hsbt:
👍 I discussed this feature with @mame. I'd like to add this too. 

Can you look `test/build.bats` ?

bin/ruby-build
```diff
@@ -1047,6 +1048,16 @@ use_homebrew_yaml() {
   fi
 }
 
+use_homebrew_gmp() {
+  local libdir="$(brew --prefix gmp 2>/dev/null || true)"
+  if [ -d "$libdir" ]; then
+    echo "ruby-build: using gmp from homebrew"
+    package_option ruby configure --with-opt-dir="$libdir"
```
mrkn, line 1055: You should specify the value of $(brew --prefix) to `--with-opt-dir` instead of libdir.

hsbt: Only `$(brew --prefix)` has side effect for other dependencies. We should keep `--with-opt-dir` values before this function and add gmp path to them.

eregon: Is there no `--with-gmp-dir`?

eregon:
> I'd also suggest to print a warning if compiling on macOS and gmp isn't installed, but I don't know if it's desirable.

I would say in general we want to have the same behavior on all OS, so I'd say warn on all or warn on none.
I think unnecessary to warn if the default behavior picks gmp anyway.
We should add gmp in https://github.com/rbenv/ruby-build/wiki#suggested-build-environment though.

Unfortunately on macOS Homebrew is no longer in the system library search path (used to be /usr/local and that just worked) so it needs to be added manually in many places :/

bin/ruby-build
hsbt: > Is there no `--with-gmp-dir`?

Yes, It's not yet. We need to use `--with-opt-dir` now.

hsbt:
> We should add gmp in https://github.com/rbenv/ruby-build/wiki#suggested-build-environment though.

👍  I also agree to add it to wiki.

casperisfine:
> Can you look `test/build.bats` ?

Yes sorry, I always forget ruby-build has a test suite. Should be good now, and I added a test case for that new feature.

> I would say in general we want to have the same behavior on all OS, so I'd say warn on all or warn on none.

Make sense. If it should warn on all, then it's probably more for Ruby's make / configure. I know this will be discussed soon upstream, so I guess we can wait for the conclusions upstream.

Either way I believe it's dangerous to run a Ruby in production without GMP. AFAIK most if not all linux distributions out there have GMP installed by default, but maybe this might be a problem with people trying to compile ruby in striped down containers? Not too sure.

I don't have an opinion on what the best course of action / implementation is, but I do strongly believe that if we end up compiling Ruby without GMP, we should warn the user that this Ruby shouldn't be used in production.

hsbt:
FYI: https://github.com/ruby/ruby/pull/6366 @mrkn try to add `--with-gmp-dir`. After that, we will use it without side effect.

mrkn:
I added `--with-gmp-dir` in the ruby master branch.

bin/ruby-build
mislav: @hsbt Should we add a guard that automatically using `--with-opt-dir` to point to GMP shouldn't happen if the user supplied their own `--with-opt-dir`? Just in case

hsbt:
We have a plan to backport `--with-gmp-dir` to Ruby 2.7-3.1 at https://bugs.ruby-lang.org/issues/19014

Should we use `--with-gmp-dir` directly? After Ruby 2.7-3.1 released, we can use libgmp automatically.

@casperisfine Do you want to enable libgmp with the old versions too? If it's yes, we handle `--with-opt-dir` and backing up user's configuration of `--with-opt-dir` outside of ruby-build.

casperisfine:
> Do you want to enable libgmp with the old versions too?

Since this is for macOS, which is pretty much never used as a production environment, I think I can leave with this only applying on future versions.

casperisfine:
I updated the PR to use `--with-gmp-dir`.

casperisfine:
What I wonder is wether we should print a warning or something if Homebrew's `gmp` isn't installed?

hsbt:
We discuss it at today as DevMeeting of Ruby core team. We should investigate `libgmp` license that is LGPL before always recommend to use `libgmp`.

We will add an additional message after clarifying the above concern.

Issue #2054, opened 2022-09-28
Change `-l` output


AlexWayfer:
### Steps to reproduce the behavior

```sh
rbenv install -l | grep '^[[:digit:]]\+\.[[:digit:]]\+\.[[:digit:]]\+$'
```

### Expected 

Output like this:

```
2.6.10
2.7.6
3.0.4
3.1.2
```

### Actual

```

Only latest stable releases for each Ruby implementation are shown.
Use 'rbenv install --list-all / -L' to show all local versions.
2.6.10
2.7.6
3.0.4
3.1.2
```

Is it from `stderr` output? Why is it here? I think this hint is OK in a regular `stdout`.

eregon:
This is the current output of ruby-build:
```
$ ruby-build -l
2.6.10
2.7.6
3.0.4
3.1.2
jruby-9.3.8.0
mruby-3.1.0
picoruby-3.0.0
rbx-5.0
truffleruby-22.2.0
truffleruby+graalvm-22.2.0
```

So anyway you need to filter if you only want CRuby releases, e.g. with the grep above.

It's likely on stderr to make it easy to process all versions, so I think you should use:
```
ruby-build -l | grep '^[[:digit:]]\+\.[[:digit:]]\+\.[[:digit:]]\+$'
or
rbenv install -l |& grep '^[[:digit:]]\+\.[[:digit:]]\+\.[[:digit:]]\+$'
```

mislav:
Thanks @eregon.

@AlexWayfer You could silence stderr to get rid of the notice in your script, but I agree that the notice is unnecessary noise if you're redirecting the output to a filter command. I've opened a PR to avoid printing the notice if the stdout was redirected.

PR #2055, opened 2022-09-29
List output: only print stderr notice if printing to a terminal

mislav:
This makes it easier to redirect `rbenv install -l` output without having to silence stderr to get rid of the extra notice.

Fixes https://github.com/rbenv/ruby-build/issues/2054


eregon (approved):
Looks good to me

PR #2056, opened 2022-09-30
Release with gh instead of with hub

mislav:
- Uses `gh` (GitHub CLI) instead of `hub` to make the release
- Generates release notes on the server instead of locally

/cc @eregon @hsbt


hsbt:
I see. Thanks ❤️

PR #2064, opened 2022-10-10
Show how to use multiple patches using HTTP

jaygooby:
Uses [Bash command grouping](https://www.gnu.org/software/bash/manual/html_node/Command-Grouping.html) to apply multiple remote patches


mislav (commented):
Hi, thank you, but we already have an example on how to apply multiple patches (from files). I don't think we need to show off advanced features of our users' shells, especially because this syntax won't work in every shell.

PR #2077, opened 2022-11-01
Mark Ruby 2.6 as EOL and 2.7 as "unsupported"

mislav:
Note: I wish we could somehow embed the support/EOL date into the build definition itself so that these warnings could automatically start printing when past the cutoff date instead of us having to ship updates to ruby-build.

Ref. #942


eregon:
Indeed. One tricky part is it's not always clear upfront when a release becomes EOL, although end of March seems pretty close: https://en.wikipedia.org/wiki/Ruby_(programming_language)#Table_of_versions

eregon (approved):

jasonkarns:
@mislav we could maintain a static file in ruby-build that is consulted by the build defs. It would still require updates to ruby-build itself, but wouldn't require retroactively modifying all the build definition files.

relatedly, nodejs has its supported version schedule in a [json file](https://github.com/nodejs/Release/blob/main/schedule.json). node-build uses this file within our scrapers to automatically add the warning "functions" to the build definitions. it's still retroactively updating build defs the same way as ruby-build here, but at least the changes are automated :D

mislav:
@jasonkarns I'd be up for anything that's semi-automated and less error-prone than the current, by-hand approach 😅

PR #2091, opened 2022-11-16
Do not automatically enable YJIT nor attempt to detect rustc

mislav:
Latest Ruby 3.2 development version now detects `rustc` on the system and automatically assumes `--enable-yjit` on susupported platforms unless `--disable-yjit` was explicitly passed. https://github.com/ruby/ruby/pull/6662

I propose that ruby-build does not attempt to duplicate this logic, but that it leaves it completely to Ruby's own build system.

Followup to https://github.com/rbenv/ruby-build/issues/1970 /cc @byroot @maximecb

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1284,27 +1284,6 @@ build_package_ldflags_dirs() {
   done
 }
 
-build_package_enable_yjit() {
-  # If YJIT is explicitly enabled or disabled, don't change the config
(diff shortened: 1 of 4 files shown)


eregon (approved):
I think this makes a lot of sense with how ruby-build in general handles dependencies, and I've already been bittne in the past by dependency checks getting outdated.

There is a mention to install `rust` on https://github.com/rbenv/ruby-build/wiki, that should be enough and is consistent with where other dependencies are specified.

byroot (approved):

PR #2111, opened 2022-12-10
Don't use OpenSSL-3, `openssl` gem doesn't support it yet

ojab:
While `openssl` gem could be build with openssl-3, it would work correctly in many cases, see https://github.com/ruby/openssl/issues/369.

Use openssl-1.1 instead until compatibility is fixed.

### DIFF
--- a/share/ruby-build/3.1.0
+++ b/share/ruby-build/3.1.0
@@ -1,2 +1,2 @@
-install_package "openssl-3.0.7" "https://www.openssl.org/source/openssl-3.0.7.tar.gz#83049d042a260e696f62406ac5c08bf706fd84383f945cf21bd61e9ed95c396e" openssl --if needs_openssl_102_300
+install_package "openssl-1.1.1s" "https://www.openssl.org/source/openssl-1.1.1s.tar.gz#c5ac01e760ee6ff0dab61d6b2bbd30146724d063eb322180c6f18a6f74e4b6aa" openssl --if needs_openssl_102_300
 install_package "ruby-3.1.0" "https://cache.ruby-lang.org/pub/ruby/3.1/ruby-3.1.0.tar.gz#50a0504c6edcb4d61ce6b8cfdbddaa95707195fab0ecd7b5e92654b2a9412854" ldflags_dirs enable_shared standard verify_openssl
--- a/share/ruby-build/3.1.0-dev
+++ b/share/ruby-build/3.1.0-dev
@@ -1,2 +1,2 @@
-install_package "openssl-3.0.7" "https://www.openssl.org/source/openssl-3.0.7.tar.gz#83049d042a260e696f62406ac5c08bf706fd84383f945cf21bd61e9ed95c396e" openssl --if needs_openssl_102_300
+install_package "openssl-1.1.1s" "https://www.openssl.org/source/openssl-1.1.1s.tar.gz#c5ac01e760ee6ff0dab61d6b2bbd30146724d063eb322180c6f18a6f74e4b6aa" openssl --if needs_openssl_102_300
 install_git "ruby-3.1.0-dev" "https://github.com/ruby/ruby.git" "ruby_3_1" ldflags_dirs autoconf standard_build standard_install_with_bundled_gems verify_openssl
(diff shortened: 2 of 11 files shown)


eregon:
What specific incompatibilities are there? Are they likely to impact most Rubyists?
My understanding so far is it only affects code using very specific parts of OpenSSL directly, i.e., it affects very few Ruby workloads but maybe I'm wrong?
And I think it's up to gems/apps using parts of the OpenSSL which is no longer supported with libssl 3 to migrate, not ruby-build to go down to a version which will be EOL way sooner than v3.

eregon:
Also to react to the title
> Don't use OpenSSL-3, openssl gem doesn't support it yet

My understanding is the openssl gem does support libssl 3 well, there are still a few edge cases, mostly unlikely to affect many users.

eregon:
I think a simple workaround if you do want to use 1.1 specifically is to use the system openssl and make sure that's 1.1.

n1xn:
I am using `ruby-setup` for my github actions and am getting this build error. Does this may have a connection to this issue?

```
[quality.yml/Linters & Style]   ❓  ::group::Installing Bundler
| Using Bundler 2.3.7 from Gemfile.lock BUNDLED WITH 2.3.7
| [command]/opt/hostedtoolcache/Ruby/3.1.2/x64/bin/gem install bundler -v 2.3.7
| ERROR:  While executing gem ... (Gem::Exception)
|     OpenSSL is not available. Install OpenSSL and rebuild Ruby (preferred) or use non-HTTPS sources
| Took   0.38 seconds
[quality.yml/Linters & Style]   ❓  ::endgroup::
[quality.yml/Linters & Style]   ❗  ::error::Error: The process '/opt/hostedtoolcache/Ruby/3.1.2/x64/bin/gem' failed with exit code 1%0A    at ExecState._setResult (/run/act/actions/ruby-setup-ruby@v1/dist/index.js:5340:25)%0A    at ExecState.CheckComplete (/run/act/actions/ruby-setup-ruby@v1/dist/index.js:5323:18)%0A    at ChildProcess.<anonymous> (/run/act/actions/ruby-setup-ruby@v1/dist/index.js:5217:27)%0A    at ChildProcess.emit (node:events:513:28)%0A    at maybeClose (node:internal/child_process:1100:16)%0A    at Process.ChildProcess._handle.onexit (node:internal/child_process:304:5)
[quality.yml/Linters & Style]   ❌  Failure - Main Setup Ruby
[quality.yml/Linters & Style] exitcode '1': failure
[quality.yml/Linters & Style] 🏁  Job failed
```

hsbt:
I agreed @eregon 's opinion. Ubuntu 22.04 or other modern Linux distribution only provide OpenSSL 3. We should use OpenSSL and improve it in the future.

ojab:
@eregon any in the https://github.com/ruby/openssl/issues/369, personally I encountered https://github.com/ruby/openssl/issues/369#issuecomment-1300929021.  
Not sure about affected %, but this issue is quite hard to debug: usually I work on another machine with openssl-1.1 and when I needed to setup working env and do something on the machine with openssl-3 — I spent an hour or soundesrstanding that failing testsuite is not a setup issue, but openssl issue and rebuilding ruby after `dnf install openssl1.1-devel` would help.
I agree that the most common usecases is already working, but personally I don't understand why ruby-build chooses subtly-broken openssl version over non-broken one.

@hsbt this PR doesn't prevent anyone from using system openssl-3 and ruby-build-could easily switch to providing openssl-3 when it's ready.

eregon:
The error in https://github.com/ruby/openssl/issues/369#issuecomment-1300929021 seems pretty clear it's an openssl-related issue. So either an issue in the openssl gem or in libssl. And that's worth fixing anyway because e.g. the latest Linux distributions already use libssl 3.

I think to change this we'd need a popular gem that doesn't work with libssl 3 or something like that, and also a PR which doesn't imply libssl3 is not supported because that's not the case.

mislav:
@ojab Thanks for bringing this to our attention. Like my teammates, I also don't believe that this edge case is worth a downgrade.

If you are bitten by this, just use `RUBY_CONFIGURE_OPTS=--with-openssl-dir=/path/to/openssl` to link all your Ruby versions across different machines to the same OpenSSL version.

PR #2140, opened 2023-02-07
Bump up latest stable versions of OpenSSL

hsbt:

PR #2145, opened 2023-02-07
Update old versions of JRuby URLs

hsbt:
https://github.com/rbenv/ruby-build/pull/2140#issuecomment-1421513364


hsbt:
@headius I removed old versions because these URLs return 404 statuses.

headius:
Thank you. We lost the official builds of these versions when our cloud provider mistakenly wiped out all our files some years ago. After that, we started pushing the tarballs to Maven so they would live forever. All of these releases are very much EOL anyway.

PR #2151, opened 2023-02-20
Remove brew update instruction

ohbarye:
Same motivation as https://github.com/cli/cli/pull/6087.
I just found that `rbenv` suggests `brew update` but it's basically unnecessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ brew install ruby-build
 
 Upgrade with:
 ```sh
-brew update && brew upgrade ruby-build
+brew upgrade ruby-build
 ```
 
 ### Clone as rbenv plugin using git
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -48,7 +48,7 @@ $ ruby-build --definitions             # lists all available versions of Ruby
 $ ruby-build 2.2.0 ~/local/ruby-2.2.0  # installs Ruby 2.2.0 to ~/local/ruby-2.2.0
 ```
 
-> **Warning**
```
mislav, line 51: The whitespace here is meaningful. Please don't commit unrelated changes that your editor makes. Thanks!

mislav (changes requested):
Thanks! In the past there used to be people setting HOMEBREW_NO_AUTO_UPDATE and forgetting about it, but now that Homebrew 4 is out and API-based installs are the default, I guess being cautious about this is not relevant anymore.

README.md
ohbarye: Oops, I was so careless. Retrieved those spaces. [ad71af3](https://github.com/rbenv/ruby-build/pull/2151/commits/ad71af30d0ab8b61d7a13da28af42873de497674)

ohbarye:
Thanks for your review!

PR #2152, opened 2023-02-22
Update TruffleRuby dev release URLs

nirvdrum:
The current _truffleruby-dev_ builds are based on Ubuntu 18.04. Since Ubuntu 18.04 will be EOL in two months, we should switch over to Ubuntu 20.04 as our newest base.

Likewise, the _truffleruby+graalvm-dev_ installation URLs broke because Java 19 was not an LTS release and there are no longer snapshot builds based on Java 19. While Java 20 builds are available, we should just use Java 17 as that's the most recent Java LTS. As such, those URLs will be more stable; they won't change until the next Java LTS release.


eregon (approved):
Thank you!

PR #2173, opened 2023-03-30
Add ruby 3.2.2

blaedj:
https://www.ruby-lang.org/en/news/2023/03/30/ruby-3-2-2-released/

I'm not sure if I should've used `scripts/update-openssl` to update openssl or not, I just copied the install_package from the 3.2.1 release as mentioned in the `scripts/update-cruby` script's `TODO` comment


share/ruby-build/3.2.2
```diff
@@ -0,0 +1,2 @@
+install_package "openssl-3.0.8" "https://www.openssl.org/source/openssl-3.0.8.tar.gz#6c13d2bf38fdf31eac3ce2a347073673f5d63263398f1f69d0df4a41253e4b3e" openssl --if needs_openssl_102_300
```
poudelmadhav, line 1: @blaedj You could use OpenSSL 3.1.0
```suggestion
install_package "openssl-3.1.0" "https://www.openssl.org/source/openssl-3.1.0.tar.gz#aaa925ad9828745c4cad9d9efeb273deca820f2cdcf2c3ac7d7c1212b7c497b4" openssl --if needs_openssl_102_300
```

poudelmadhav (changes requested):

blaedj:
probably makes sense to close this in favor of https://github.com/rbenv/ruby-build/pull/2174, which actually includes the updates for the 3.1, 3.0 and 2.7 series

PR #2177, opened 2023-03-30
Fall back on shasum if sha256sum is unavailable

jas14:
`sha256sum` is not by default installed on macOS, but `shasum` is. They function equivalently, as far as I can tell, so this PR introduces a seamless fallback onto `shasum` so macOS users can run `scripts/update-cruby`. 🙂

If desired, I can update the other scripts as well.


hsbt:
Thanks!

PR #2187, opened 2023-04-14
Fix compilation of Ruby 3.2.x on FreeBSD

jarmo:


bin/ruby-build
```diff
@@ -1040,6 +1043,18 @@ use_homebrew_yaml() {
   fi
 }
 
+use_freebsd_yaml() {
+  if [ "FreeBSD" = "$(uname -s)" ]; then
+    local release="$(uname -r)"
+    if [ "${release%%.*}" -ge 11 ]; then
+      if pkg info -e libyaml > /dev/null; then
+        package_option ruby configure --with-libyaml-dir="$(pkg info --prefix libyaml | cut -wf2)"
```
mislav, line 1051: Could we just get away with calling `pkg info --prefix libyaml` once instead of `pkg info -e libyaml && pkg info --prefix`?

And what is the full output of `pkg info --prefix` before it gets cut?

mislav (changes requested):
Thanks for working on this!

How come these checks are scoped only to FreeBSD 11 or later?

jarmo:
> How come these checks are scoped only to FreeBSD 11 or later?

Not sure. I used the same style as was used in the `use_freebsd_readline` function. Since I didn't understand the reason behind that check then I went with the safer choice of doing the same and not deviate away from it. Maybe it's time to remove these version checks everywhere since FreeBSD 10 is already EOL-ed?

> Could we just get away with calling pkg info --prefix libyaml once instead of pkg info -e libyaml && pkg info --prefix?

It's possible. However I think that then we need to check for empty string after command `pkg info -p libyaml 2>/dev/null | cut -wf2`. Currently `pkg info -e` returns exit code 0 or 1 depending on the existance of the package:
```
root@freebsd:~ pkg info -e libyaml
root@freebsd:~ echo $?
0
root@freebsd:~ pkg info -e foo
root@freebsd:~ echo $?
1
root@freebsd:~ pkg info -p libyaml
libyaml-0.2.5                  /usr/local
root@freebsd:~ pkg info -p foo
pkg: No package(s) matching foo
```

Not sure what `stdout` is been redirected to `/dev/null` in `use_freebsd_readline` function when doing `pkg info -e PKG`, but as I wrote already then I followed the same style.

> And what is the full output of pkg info --prefix before it gets cut?

Output is seen above.

jarmo:
@mislav made all the relevant changes.

bin/ruby-build
```diff
@@ -106,6 +106,15 @@ is_mac() {
   [ $# -eq 0 ] || [ "$(osx_version)" "$@" ]
 }
 
+is_freebsd() {
+  [ "$(uname -s)" = "FreeBSD" ]
+}
+
+freebsd_package_prefix() {
+  local package="$1"
+  echo "$(pkg info --prefix "$package" 2>/dev/null | cut -wf2)"
```
mislav, line 115: How about just unwrapping this instead of capturing the output and echo-ing it?
```suggestion
  pkg info --prefix "$package" 2>/dev/null | cut -wf2
```

bin/ruby-build
```diff
@@ -1040,6 +1052,15 @@ use_homebrew_yaml() {
   fi
 }
 
+use_freebsd_yaml() {
+  if is_freebsd; then
+    local libyaml_prefix="$(freebsd_package_prefix libyaml)"
+    if [ "$libyaml_prefix" ]; then
```
mislav, line 1058: Nit: although it might be functionally equivalent, I do like the explicitness of testing strings for not being blank, i.e.
```suggestion
    if [ -n "$libyaml_prefix" ]; then
```

mislav (changes requested):
Thanks; looking good! Only style suggestions remain

bin/ruby-build
jarmo: Thanks for noticing this silly code :) Fixed it.

bin/ruby-build
jarmo: Explicit checks added.

mislav (approved):

PR #2207, opened 2023-06-13
Add TruffleRuby and TruffleRuby+GraalVM 23.0.0

eregon:
* Use the new Oracle GraalVM distribution which uses the GFTC license as it is significantly faster than GraalVM CE and is free for development and production use.
* See https://medium.com/graalvm/whats-new-in-graalvm-languages-161527df3d76

### DIFF
--- a/share/ruby-build/truffleruby-23.0.0
+++ b/share/ruby-build/truffleruby-23.0.0
@@ -1,0 +1,21 @@
+platform="$(uname -s)-$(uname -m)"
+case $platform in
+Linux-x86_64)
+  install_package "truffleruby-23.0.0" "https://gds.oracle.com/api/20220101/artifacts/FD4AB182EA4CEDFDE0531518000AF13E/content#57f770e7ba754327c8c160571978c7da0e10a086f8b5dc542e5c8ff5b2c83190" truffleruby
+  ;;
+Linux-aarch64)
+  install_package "truffleruby-23.0.0" "https://gds.oracle.com/api/20220101/artifacts/FD40BA2367C226B6E0531518000AE71A/content#b75b6ddab76bda5d9cbff2fb53f5c5559fa6c5e11b845986b02c3cd4d3b98b3a" truffleruby
+  ;;
+Darwin-x86_64)
+  use_homebrew_openssl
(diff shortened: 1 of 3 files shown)


share/ruby-build/truffleruby-23.0.0
```diff
@@ -0,0 +1,21 @@
+platform="$(uname -s)-$(uname -m)"
+case $platform in
+Linux-x86_64)
+  install_package "truffleruby-23.0.0" "https://gds.oracle.com/api/20220101/artifacts/FD4AB182EA4CEDFDE0531518000AF13E/content#57f770e7ba754327c8c160571978c7da0e10a086f8b5dc542e5c8ff5b2c83190" truffleruby
```
eregon, line 4: These URLs are not pretty, but they are stable and should keep working forever.
We're working on getting a stable URL format for them, similar to `truffleruby+graalvm-23.0.0` in the other file.

hsbt:
👍  I have questions about new GraalVM.

* Is the new GraalVM only for TruffleRuby-23.0 or later?
* Should we switch new GraalVM from GraalVM CE for TruffleRuby-23.0 or later? I'm curious to see what happens to GraalVM CE in the future.

eregon:
> * Is the new GraalVM only for TruffleRuby-23.0 or later?

Yes.

> * Should we switch new GraalVM from GraalVM CE for TruffleRuby-23.0 or later? I'm curious to see what happens to GraalVM CE in the future.

That's what this PR does.

There is still GraalVM CE and also for future releases, but given Oracle GraalVM is faster and free for development and production use, I think there is little value to use GraalVM CE.
I want to give the best experience to TruffleRuby users, and Oracle GraalVM provides that through being faster and more efficient (also in memory used).

To clarify, before 23.0 it was called GraalVM Enterprise Edition, which was not free for production use, now it's called Oracle GraalVM and is under the GFTC license and so it is free for development and production use.

The [FAQ](https://www.oracle.com/java/technologies/javase/jdk-faqs.html#GraalVM-licensing) linked in the blog post has more details (GraalVM for JDK 17/20 corresponds to the GraalVM 23.0.0 release, we always use the LTS JDK for TruffleRuby in ruby-build, which is 17 here).

hsbt (approved):

hsbt:
Thanks for your explanation.

PR #2223, opened 2023-07-30
Display the actual error when testing for Ruby's openssl extension

YellowApple:
I was encountering some build issues on MicroOS and needed the actual exception message in order to pinpoint a workaround.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1243,7 +1243,7 @@ build_package_verify_openssl() {
       begin
         require lib
       rescue LoadError => e
-        $stderr.puts "There was a problem loading the Ruby #{lib} extension; make sure it compiled successfully.  (#{e})"
+        $stderr.puts "Loading the Ruby #{lib} extension failed (#{e})"
       end
     end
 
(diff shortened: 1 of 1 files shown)


YellowApple:
(The specific error and workaround, in case anyone's curious: https://github.com/rvm/rvm/issues/3150#issuecomment-359323888)

mislav (commented):
Hi, thank you for the contribution, but I do not believe that including the LoadError error message will meaningfully improve output for scenarios in which extensions failed to compile. In order to debug a failed extension, it's best to view the extension compilation log instead of just observing the error message. In the future, we might surface the locations of those logs better in ruby-build output, but for now, I'm opting to leave the LoadError output as-is.

mislav:
> The problem is that the extension compiles "successfully", so there is no indication in the compilation log that something went wrong

That's a good point; thanks.

> Or could/should we reword `"The Ruby #{lib} extension was not compiled."` to be more accurate (e.g. `"There was a problem loading the Ruby #{lib} extension; make sure it compiled successfully."`)?

I like this idea! You are welcome to further amend your PR.

YellowApple:
> I like this idea! You are welcome to further amend your PR.

Done :+1:

mislav (approved):
Thanks! After some deliberation, I've shortened the error message further. I plan to soon make an overhaul of this whole check process to make the output even clearer, but for now, I hope that the addition of the actual error message helps.

YellowApple:
Looks good, thanks!

Issue #2025, opened 2022-08-13
Compiler invocation not printed with --verbose


varungandhi-src:
### Steps to reproduce the behavior

Apply an incorrect patch to `readline.c` that causes compilation to fail (to mimic the situation I'm in, I don't have a minimal reproducer yet). Try building with `--verbose`. You'll see something like:

```
readline.c: ... error: <compiler error message>

1 error generated.
make[2]: *** [readline.o] Error 1
make[1]: *** [ext/readline/all] Error 2

<other stuff>

Inspect or clean up the working tree at /tmp/ruby-build.20220813172251.84708.XSQCpi
Results logged to /tmp/ruby-build.20220813172251.84708.log
```

The failing invocation itself is not printed (I checked the mentioned `log` as well), perhaps because the verbose is not being passed through to the make invocation.

### Expected vs. actual behavior

With `--verbose`, I would expect all compiler invocations to be printed, so that one can narrow down the cause of the failure quickly, and potentially compare the invocation with a successful build in a different configuration.

### Logs

n/a -- this bug isn't about my specific failure (I'll investigate that), but about `--verbose`.

varungandhi-src:
It wasn't too hard to work around this by using a custom CC which logs the invocations, so I think it's OK to close this if you want `--verbose` to not be so verbose. IIRC the compiler invocations for OpenSSL are printed though, so this behavior seems inconsistent and potentially unintentional.

eregon:
It depends on what the Makefile prints, for openssl it seems to print command lines, for ruby the Makefile prints something like `compiling foo.c`. Not sure if there is a way to tweak that.

mislav:
Agreed that ruby-build could logging names of external commands and the arguments that it invokes them with. These would be generally useful in logs, not just in `--verbose` mode.

PR #2230, opened 2023-08-03
Restructure console output

mislav:
I've always imagined that ruby-build could print the external commands executed, such as `./configure` (with full arguments list), `make`, and `make install` in most cases. This experimental branch attempts to achieve just that, while at the same time cleaning up old ruby-build code that uses so much shell redirection (e.g. `>&4 2>&1`) that it's hard to figure out which output goes where at a given point in the code.

The goal of this work would be improved user experience: hopefully, the printed commands would make it more clear as to what ruby-build does, and more importantly, what it _doesn't_ do. Feedback welcome! 🙏 

Here is the current output (note the `===>` markers):
```
$ bin/ruby-build 3.2.2 /tmp/ruby-test

To follow progress, use 'tail -f /var/folders/6k/fwzzx67n163b0lfftyw3313h0000gn/T/ruby-build.20230803034426.4419.log' or pass --verbose
Downloading ruby-3.2.2.tar.gz...
===> wget -nv --show-progress -O ruby-3.2.2.tar.gz https://cache.ruby-lang.org/pub/ruby/3.2/ruby-3.2.2.tar.gz
ruby-3.2.2.tar.gz                      100%[===========================================================================>]  19,52M  31,1MB/s    in 0,6s    
2023-08-03 03:44:27 URL:https://cache.ruby-lang.org/pub/ruby/3.2/ruby-3.2.2.tar.gz [20467023/20467023] -> "ruby-3.2.2.tar.gz" [1]
===> tar xzf ruby-3.2.2.tar.gz
Installing ruby-3.2.2...
ruby-build: using libyaml from homebrew
ruby-build: using gmp from homebrew
===> ./configure --prefix=/tmp/ruby-test --enable-shared --with-readline-dir=/opt/homebrew/opt/readline --with-libyaml-dir=/opt/homebrew/opt/libyaml --with-gmp-dir=/opt/homebrew/opt/gmp --disable-install-doc --with-openssl-dir=/opt/homebrew/opt/openssl@1.1
===> make -j 8
===> make install
Installed ruby-3.2.2 to /tmp/ruby-test
```

Features in this PR:
- prints the full invocation of external commands as they are executed (common shell commands such as `cd`, `mkdir`, and `ln -s` are _not_ logged since I think that would add too much noise)
- connects downloaders such as curl/wget/aria2c to the original stderr of the terminal so that they can draw progress bars; closes #536
- `ruby-build --verbose` mode should be unchanged, for backwards compatibility

Overall goals:
- Better clarity around what ruby-build actually does. Often users report build problems to this project, when it fact they should be reported to the upstream project that they're trying to build, such as the Ruby codebase itself.
- Make it obvious which build step actually failed.
- Improve build failure output to give more power to users for debugging. For example, draw attention to environment variables (e.g. CFLAGS, LDFLAGS) that could have affected the build process.
- Reduce the number of BUILD FAILED reports to this project that are not indicative of any problem with the ruby-build codebase.

Closes https://github.com/rbenv/ruby-build/issues/2025

TODO:
- [x] Remove the output line that suggests to `tail -f` the log file. ~Instead, **have ruby-build handle SIGINFO** (<kb>Ctrl-T</kbd>) by entering `--verbose` mode. That way the user can interactively opt into seeing the verbose log without copy-pasting the log location.~
- [x] **Do not print last 10 lines of the log file** on build failures. In my experience, the last 10 lines rarely include the actual error message, but users keep pasting that output in their build failure reports anyway. The full log file is what matters most.
- [ ] Consider **printing relevant environment variables** on build failures.
- [x] **Polish the output format** with more clear section markers (I'm not in love with `===>`) and a splash of terminal colors for readability.
- [x] Log ruby-build progress to the process' original stdout instead of to stderr, since stderr should only be used for warnings and error messages.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -369,7 +369,7 @@ verify_checksum() {
       echo "unexpected checksum length: ${#expected_checksum} (${expected_checksum})"
       echo "expected 0 (no checksum), 32 (MD5), or 64 (SHA2-256)"
       echo
-    } >&4
+    } >&2
     return 1 ;;
   esac
 
@@ -386,7 +386,7 @@ verify_checksum() {
       echo "checksum mismatch: ${filename} (file is corrupt)"
       echo "expected $expected_checksum, got $computed_checksum"
       echo
-    } >&4
+    } >&2
     return 1
   fi
 }
@@ -396,7 +396,7 @@ http() {
   [ -n "$2" ] || return 1
   shift 1
 
-  RUBY_BUILD_HTTP_CLIENT="${RUBY_BUILD_HTTP_CLIENT:-$(detect_http_client 2>&3)}"
+  RUBY_BUILD_HTTP_CLIENT="${RUBY_BUILD_HTTP_CLIENT:-$(detect_http_client)}"
   [ -n "$RUBY_BUILD_HTTP_CLIENT" ] || return 1
 
   # http_get_curl, http_get_wget, etc.
@@ -474,7 +474,7 @@ fetch_tarball() {
 
   if [ "$package_url" != "${package_url%bz2}" ]; then
     if ! type -p bzip2 >/dev/null; then
-      echo "warning: bzip2 not found; consider installing \`bzip2\` package" >&4
+      echo "warning: bzip2 not found; consider installing the \`bzip2\` package" >&2
     fi
     package_filename="${package_filename%.gz}.bz2"
     tar_args="${tar_args/z/j}"
@@ -619,7 +619,7 @@ build_package_warn_eol() {
     echo "WARNING: $package_name is past its end of life and is now unsupported."
     echo "It no longer receives bug fixes or critical security updates."
     echo
-  } >&3
+  } >&2
 }
 
 build_package_warn_unsupported() {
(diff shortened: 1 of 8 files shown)


mislav:
I made some output tweaks; let me know what you think!

mislav:
I have cleaned up the implementation and commit history in this branch and plan to merge it soon.

This is the most significant recent change to ruby-build and, even though I tried my best to test it thoroughly, there is a larger-than-usual chance of bugs introduced due to the sheer number of code changes in this branch. If something breaks in the next release of ruby-build, it would be sensible to assume this PR was the culprit.

eregon (approved):
Nice!

I think it would be useful to have a way to show all commands executed, so one can really do the same as `ruby-build` manually to investigate/debug. Maybe `-v/--verbose` should imply that?

mislav:
> I think it would be useful to have a way to show all commands executed, so one can really do the same as `ruby-build` manually to investigate/debug (including cd, mkdir, ln, etc). Maybe `-v/--verbose` should imply that?

That's a good idea, and I briefly entertained the concept of adding a verbosity mode to print *all* external commands no matter how small, but I wasn't sure how to expose that. The `--verbose` mode might be suitable for this, since it will already have lots of detailed output. I can experiment with this in a followup PR.

PR #2258, opened 2023-09-21
Add RUBY_BUILD_TARBALL_OVERRIDE to override the ruby tarball URL

eregon:
* And add RUBY_BUILD_IGNORE_CHECKSUM when it is expected the alternative URL has different contents.

This is useful for instance when testing a truffleruby release which is not published yet.
Then I need to use a custom URL, like:
```
RUBY_BUILD_MIRROR_PACKAGE_URL=https://github.com/my/url.tar.gz ruby-build truffleruby-23.1.0 ~/.rubies/truffleruby-test
```
I also want this to try dev builds at a different URL, hence this change.
(and those might have a different checksum too, see below)

And finally I would also like a way to skip the checksum check, for instance when trying a variant of a release which therefore has a different URL and contents but yet I want to reuse the existing definition.
For example this can be useful if I want to try the CE build instead of GFTC build of truffleruby 23.1.0, like this:
```
RUBY_BUILD_IGNORE_CHECKSUM=1 RUBY_BUILD_MIRROR_PACKAGE_URL=https://github.com/oracle/truffleruby/releases/download/graal-23.1.0/truffleruby-community-23.1.0-linux-amd64.tar.gz ruby-build truffleruby-23.1.0 ~/.rubies/truffleruby-test
```

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -381,6 +381,10 @@ fetch_tarball() {
   local checksum
   local extracted_dir
 
+  if is_ruby_package "$1" && [ -n "$RUBY_BUILD_TARBALL_OVERRIDE" ]; then
+    package_url="$RUBY_BUILD_TARBALL_OVERRIDE"
+  fi
+
   if [ "$package_url" != "${package_url/\#}" ]; then
     checksum="${package_url#*#}"
     package_url="${package_url%%#*}"
@@ -1257,14 +1261,23 @@ isolated_gem_install() {
 
 apply_ruby_patch() {
   local patchfile
-  case "$1" in
-  ruby-* | jruby-* | rubinius-* | truffleruby-* )
+  if is_ruby_package "$1"; then
     patchfile="$(mktemp "${TMP}/ruby-patch.XXXXXX")"
     cat "${2:--}" >"$patchfile"
 
     local striplevel=0
     grep -q '^--- a/' "$patchfile" && striplevel=1
     patch -p$striplevel --force -i "$patchfile"
+  fi
+}
+
+is_ruby_package() {
+  case "$1" in
+  ruby-* | jruby-* | rubinius-* | truffleruby[+-]* | mruby-* | picoruby-* )
+    return 0
+    ;;
+  *)
+    return 1
     ;;
   esac
 }
--- a/README.md
+++ b/README.md
@@ -82,9 +82,10 @@ The build process may be configured through the following environment variables:
 | `RUBY_BUILD_CURL_OPTS`          | Additional options to pass to `curl` for downloading.                                            |
 | `RUBY_BUILD_WGET_OPTS`          | Additional options to pass to `wget` for downloading.                                            |
 | `RUBY_BUILD_MIRROR_URL`         | Custom mirror URL root.                                                                          |
-| `RUBY_BUILD_MIRROR_PACKAGE_URL` | Custom complete mirror URL (e.g. http://mirror.example.com/package-1.0.0.tar.gz).                  |
+| `RUBY_BUILD_MIRROR_PACKAGE_URL` | Custom complete mirror URL (e.g. http://mirror.example.com/package-1.0.0.tar.gz).                |
 | `RUBY_BUILD_SKIP_MIRROR`        | Bypass the download mirror and fetch all package files from their original URLs.                 |
 | `RUBY_BUILD_ROOT`               | Custom build definition directory. (Default: `share/ruby-build`)                                 |
+| `RUBY_BUILD_TARBALL_OVERRIDE`   | Override the URL to fetch the ruby tarball from, optionally followed by `#checksum`.             |
 | `RUBY_BUILD_DEFINITIONS`        | Additional paths to search for build definitions. (Colon-separated list)                         |
 | `CC`                            | Path to the C compiler.                                                                          |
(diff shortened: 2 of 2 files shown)


bin/ruby-build
```diff
@@ -1194,7 +1198,7 @@ isolated_gem_install() {
 apply_ruby_patch() {
   local patchfile
   case "$1" in
-  ruby-* | jruby-* | rubinius-* | truffleruby-* )
+  ruby-* | jruby-* | mruby-* | picoruby-* | truffleruby-* | truffleruby+graalvm-* )
```
mislav, line 1201: This feels like it belongs in a separate PR

mislav (commented):
I'm skeptical of this feature because I already wasn't a fan of the original RUBY_BUILD_MIRROR_PACKAGE_URL feature, and this adds even more features around it https://github.com/rbenv/ruby-build/pull/1793

> And finally I would also like a way to skip the checksum check, for instance when trying a variant of a release which therefore has a different URL and contents but yet I want to reuse the existing definition.

From everything you described so far, it sounds like what you're doing isn't at all mirroring existing releases, but dynamically modifying the target that a build definition downloads and installs. If you need to modify a truffleruby definition slightly, would it be an option for you to create a custom definition and provide `ruby-build` with a path to your custom definition?

bin/ruby-build
eregon: It's just a cleanup I noticed while working on this and already its own commit, but sure I can split it up.

eregon:
Right, this is indeed not mirroring.
@mislav How would you feel about adding another environment variable then?
I was thinking something like `RUBY_BUILD_OVERRIDE_RUBY_URL` or so originally (and this would only override the URL if it's "a ruby" package, which then would use the same check as [here](https://github.com/rbenv/ruby-build/pull/2258#discussion_r1334230471)).

(then I saw RUBY_BUILD_MIRROR_PACKAGE_URL already existed and it seemed nicer to extend it)

mislav:
I take this as custom build definitions not being a suitable workaround for you?

eregon:
Yeah they would be too annoying to be of any real use.
Also for the first 2 use cases, those are actually quite related to mirroring, and custom definitions there would not be testing what we want.

eregon:
@mislav What do you think of the above? Which way would be OK to merge?

bin/ruby-build
mislav: The problem with this change being bundled here, apart from being wholly unrelated, is that it won't get its own changelog entry automatically upon cutting a release with `script/release` (nor when creating a release using the web interface and choosing "Generate release notes"). I believe that all feature changes deserve a changelog entry.

mislav, line 1201:
```suggestion
  ruby-* | jruby-* | mruby-* | picoruby-* | truffleruby[+-]* )
```

bin/ruby-build
```diff
@@ -365,14 +365,18 @@ fetch_tarball() {
...
     package_url="${package_url%%#*}"
+  fi
 
-    if [ -n "$RUBY_BUILD_MIRROR_URL" ]; then
-      if [[ -z "$RUBY_BUILD_DEFAULT_MIRROR" || $package_url != */cache.ruby-lang.org/* ]]; then
-        mirror_url="${RUBY_BUILD_MIRROR_URL}/$checksum"
-      fi
-    elif [ -n "$RUBY_BUILD_MIRROR_PACKAGE_URL" ]; then
-      mirror_url="$RUBY_BUILD_MIRROR_PACKAGE_URL"
+  if [ -n "$RUBY_BUILD_IGNORE_CHECKSUM" ]; then
```
mislav, line 370: I do not believe we should expose this additional option, since we would not want anyone to actually use it.

Instead, in order to support the use-case that you are describing (testing Ruby releases that do not yet have their own definition), I suggest that we introduce a new environment variable completely orthogonal to the mirror feature (since this is technically not related to mirroring, either). The new environment variable could be called something like `RUBY_BUILD_TARBALL_OVERRIDE`, and the value read from it would replace both the URL and checksum values in `fetch_tarball`. If the environment variable only supplies a URL, then checksum checks would be off anyway, but a checksum could also be optionally supplied too. How does that sound?

mislav (changes requested):

bin/ruby-build
eregon: Yes that sounds great, I'll update the PR with it.

eregon:
@mislav Updated, could you review and merge if it's OK?

bin/ruby-build
```diff
@@ -371,6 +371,10 @@ fetch_tarball() {
   local checksum
   local extracted_dir
 
+  if is_ruby_package "$1" && [ -n "$RUBY_BUILD_TARBALL_OVERRIDE" ]; then
```
eregon, line 374: The reason we need the is_ruby_package is because fetch_tarball is also used for non-ruby packages like openssl and overriding both URLs with the same URL would make no sense.

bin/ruby-build
eregon: In the updated PR it's fully related and necessary.

eregon:
Rebased on top of master to include https://github.com/rbenv/ruby-build/pull/2268, and locally `shellcheck bin/ruby-build` is clean.

mislav (approved):
Love the brevity of the final solution! Thank you.

eregon:
Indeed it's beautiful how small it is, I was positively surprised as well :)

Issue #1791, opened 2021-08-06
Installing ruby fails in the openssl step when having set MAKE env variable to `make -j 16`


johannesluedke:
Steps to reproduce: Install ruby 3.0.2 on MacOS 10.15.7 (19H1217) 

Observed result:
```
Configuring OpenSSL version 1.1.1k (0x101010bfL) for darwin64-x86_64-cc
Using os-specific seed configuration
Creating configdata.pm
Creating Makefile

**********************************************************************
***                                                                ***
***   OpenSSL has been successfully configured                     ***
***                                                                ***
***   If you encounter a problem while building, please open an    ***
***   issue on GitHub <https://github.com/openssl/openssl/issues>  ***
***   and include the output from the following command:           ***
***                                                                ***
***       perl configdata.pm --dump                                ***
***                                                                ***
***   (If you are new to OpenSSL, you might want to consult the    ***
***   'Troubleshooting' section in the INSTALL file first)         ***
***                                                                ***
**********************************************************************
/Users/myusername/.asdf/plugins/ruby/ruby-build/bin/ruby-build: line 595: make -j16: command not found
```

Exptected result: it works

johannesluedke:
I think I found the reason -- somehow my MAKE env variable was set to `export MAKE="make -j$(nproc)"`. Could one adjust L959 so it also works in case flags have been configured?

eregon:
`$MAKE` is the standard env var to tell which `*make` *program* to use, not to pass flags.
There is nothing we can do if that's set incorrectly.
For gmake (and maybe others) use `export MAKEFLAGS=-j8` instead.

PR #2270, opened 2023-10-13
Cleanup in OpenSSL compilation step

mislav:
- The `make -j 1` workaround seems neither in effect nor necessary anymore
- Assume that KERNEL_BITS workaround isn't necessary anymore
- Declare more variables as local

Ref. 959e5cb22b19621b79c2236a62e636957dcac180

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1091,7 +1091,7 @@ use_homebrew_openssl() {
 
 build_package_openssl() {
   # Install to a subdirectory since we don't want shims for bin/openssl.
-  OPENSSL_PREFIX_PATH="${PREFIX_PATH}/openssl"
+  local OPENSSL_PREFIX_PATH="${PREFIX_PATH}/openssl"
 
   # Put openssl.conf, certs, etc in ~/.rbenv/versions/*/openssl/ssl
   OPENSSLDIR="${OPENSSLDIR:-$OPENSSL_PREFIX_PATH/ssl}"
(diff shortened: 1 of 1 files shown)


bin/ruby-build
```diff
@@ -1102,27 +1102,17 @@ build_package_openssl() {
...
   local nokerberos
   [[ "$1" != openssl-1.0.* ]] || nokerberos=1
 
   # Compile a shared lib with zlib dynamically linked.
   package_option openssl configure --openssldir="$OPENSSLDIR" zlib-dynamic no-ssl3 shared ${nokerberos:+no-ssl2 no-krb5}
 
-  # Default MAKE_OPTS are -j 2 which can confuse the build. Thankfully, make
-  # gives precedence to the last -j option, so we can override that.
-  package_option openssl make -j 1
+  # Skip building OpenSSL docs, which is slow.
```
eregon, line 1111: This is still needed, except for openssl 3+.
For example I found https://github.com/rbenv/ruby-build/issues/1791 where it's needed for 1.1

mislav: @eregon During my work on #2230 I've found that our [default MAKE_OPTS](https://github.com/rbenv/ruby-build/blob/1d9b4e268f13b71a09c59a29dc0de5217236d130/bin/ruby-build#L589) take precedence over the `-j 1` setting, resulting in this invocation:
```
make -j 1 -j 8
```
Since make most likely only respects the last `-j` option, it seems that some time in the last decade, our `-j 1` workaround stopped having effect, but our users largely haven't noticed.

Furthermore, I have just built openssl-1.1.1w successfully using `make -j 8` on macOS.

eregon: Mmh, so the comment above was wrong then?

`{ "$MAKE" "${!PACKAGE_MAKE_OPTS_ARRAY}" $MAKE_OPTS ${!PACKAGE_MAKE_OPTS}`
So it depends on whether `package_option openssl make -j 1` adds it to PACKAGE_MAKE_OPTS_ARRAY or PACKAGE_MAKE_OPTS. If PACKAGE_MAKE_OPTS it should be fine.
Seems it's to PACKAGE_MAKE_OPTS_ARRAY hence what you saw: https://github.com/rbenv/ruby-build/blob/1d9b4e268f13b71a09c59a29dc0de5217236d130/bin/ruby-build#L555

IIRC some older versions of openssl don't always work with parallel make. And it's transient, sometimes it just works and sometimes not, e.g. due to missing dependencies between make targets.
I have found https://github.com/openssl/openssl/issues/298 and https://github.com/openssl/openssl/issues/5762#issuecomment-376622684, so it seems a problem for OpenSSL < 1.1.0.

eregon: Also https://github.com/conan-io/conan-center-index/pull/3001#issuecomment-698794769 so maybe some older versions happen to work with parallel builds but in general it seems not guaranteed at all for OpenSSL < 1.1.0.
So I think we should still `-j1` on OpenSSL < 1.1.0.

eregon: Could we either handle it here and update these definitions < 2.4 to automatically set it?
I don't think it's any good if we keep definitions which are broken and need a manual workaround like this when we know the issue and it's somewhat-indirectly caused by ruby-build (which builds openssl).
For instance it would break https://github.com/ruby/ruby-builder for Ruby < 2.4 so I'm not keen on breaking this.

PR #2272, opened 2023-10-13
Cleanup in truffleruby+graalvm installation

mislav:
Reuse original `build_package_copy` instead of having to maintain an additional `build_package_copy_to` step.

As a bonus, this prevents a global variable `to` from leaking.


eregon:
I tested it locally, it seems fine.

PR #2274, opened 2023-10-13
Avoid compiling OpenSSL if the user supplied `--with-openssl-dir` on the command line

mislav:
Same with linking libyaml and checking for `--disable-shared`. This considers all user configuration inputs when checking for existing flags.

Fixes:
```sh
ruby-build <version> <destination> -- --disable-shared --with-openssl-dir="..." --with-libyaml-dir="..."
```

Followup to https://github.com/rbenv/ruby-build/pull/2267

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -956,7 +956,7 @@ require_llvm() {
 }
 
 needs_yaml() {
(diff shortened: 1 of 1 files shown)


eregon (approved):

mislav:
@eregon Thanks for the review! Since this is a bug fix, it would have been good too include it in the release as well https://github.com/rbenv/ruby-build/releases/tag/v20231014

Issue #2157, opened 2023-03-06
Automatically link to Homebrew openssl


mislav:
I think it would make sense to default to Homebrew's openssl@correctVersion if it's installed. It would be consistent with readline, gmp and libyaml. And also because Homebrew's openssl works for using any extension which uses a Homebrew formula which links against libssl, unlike the custom-built openssl (which would cause to load 2 libssl in the same process, which usually doesn't work well).

_Originally posted by @eregon in https://github.com/rbenv/ruby-build/issues/2023#issuecomment-1456819753_

* * *

In the past we avoided setting `--with-openssl-dir` by default to link to Homebrew's OpenSSL because we were not sure if this is what the user wanted or desired, and because ruby-build will automatically download and compile the correct version of openssl if system openssl is missing (or broken on macOS).

However, we find ourselves suggesting to our users to manually set RUBY_CONFIGURE_OPTS to link to Homebrew's openssl so often, that it could as well be a built-in feature. The benefits are:
- **Faster build times.** A separate version of openssl now doesn't have to be compiled per each Ruby version, and `brew install openssl` is fast due to it being “bottled”.
- **Not having to know** which `openssl@VERSION` to link to. Currently, we have to explain to people to manually link to `openssl@1.1` for older Rubies and to `openssl@3` for newer. As new OpenSSL versions keep being released in the future, if would be a relief to our users that they don't have to keep track of compatibility.
- **Keep openssl upgraded.** Doing `brew upgrade openssl` will keep it upgraded for any Ruby version linked to it.

In the past, we did try to automatically link to Homebrew's openssl, but we hardcoded that to `openssl@1.1` for all Ruby versions which naturally caused troubles, so it had to be reverted: https://github.com/rbenv/ruby-build/pull/1375. If we brought back the feature, we would have to dynamically choose `openssl@VERSION` depending on Ruby version.

dbackeus:
Just to point out an issue with relying on brews libraries: historically brew has had a tendency of bricking mine and my colleagues Ruby installations by automatically upgrading and removing older versions of libraries, which has then broken the dynamic linking.

Perhaps the best of all worlds would be to default to using locally pre-existing libraries, but make copies of them into `~/.rbenv` somewhere to ensure that Ruby installations are not subject to dynamic linking failures once installed?

dentarg:
On that topic, I think `export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1` have probably helped me in this regard.

[...]

PR #2275, opened 2023-10-13
Automatically link to Homebrew openssl

mislav:
If a build definition invokes an openssl requirement, but there is no satisfying system OpenSSL version—e.g. always a state of affairs on macOS—automatically query Homebrew-installed versions with `brew --prefix openssl@<version>` (with a range of versions between "1.1" and "3.1") and automatically link to the latest one found that Ruby is known to be compatible with.

The goal is to skip compiling a per-Ruby version of OpenSSL whenever possible, speeding up Ruby installation and reducing the number of compile dependencies for a typical user, hopefully resulting in fewer "build failed" reports in this repository.

We tried this before by always linking to `openssl@1.1`, but that was neither future-proof (today is more common to have `openssl@3` installed) nor backward compatible as it sabotaged installations of older Ruby versions. https://github.com/rbenv/ruby-build/pull/1375

Fixes https://github.com/rbenv/ruby-build/issues/2157
Followup to #1974 

* * *

Feedback wanted: this PR introduces this syntax:
```sh
install_package openssl-1.1.1 "https://..." --if "needs_openssl 1.0.2 3.x.x"
```
i.e. there is a single function `needs_openssl` that now takes arbitrary arguments. This replaces the previous set of functions named like `needs_openssl_102_300` which weren't future-proof. Now, every build definition explicitly and precisely specifies which openssl versions it wants.

I'm not in love with the new syntax, especially because quotes are necessary to keep the condition function and its arguments as a single argument in bash. But I do like the added functionality! What do you think?

Possible alternative syntaxes:
- `--if "needs_openssl 1.0.2 3.x.x"` - I don't like the quotes; it looks like the code in question will get `eval`'d from string (not true)
- `--if needs_openssl:1.0.2:3.x.x`
- `--if needs_openssl:1.0.2-3.x.x`
- `--if needs_openssl@1.0.2-3.x.x`

Keep in mind that the intent here is that this is a more generalized syntax for any `--if` condition, means that it could facilitate passing different arguments than just version numbers in the future.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1023,6 +1023,8 @@ use_freebsd_libffi() {
   fi
 }
 
+# macOS prevents linking to its system OpenSSL/LibreSSL installation, so
+# it's basically useless for Ruby purposes.
 has_broken_mac_openssl() {
   is_mac || return 1
   local openssl_version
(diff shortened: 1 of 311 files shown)


bin/ruby-build
```diff
@@ -1038,43 +1038,55 @@ system_openssl_version() {
...
-  local version
-  version="$(system_openssl_version)"
-  (( version < 96 || version >= 110 ))
+  local system_version
+  system_version="$(system_openssl_version)" || return 0
+  system_version="$(normalize_semver "$system_version")"
+
+  local lower_bound upper_bound
+  lower_bound="$(normalize_semver "$1")"
+  upper_bound="$(normalize_semver "$1")"
```
eregon, line 1072: `$2` right?

mislav: Good catch; thanks!

mislav:
@hsbt @eregon This is now ready for final review. The change here is that ruby-build will automatically link to a compatible Homebrew OpenSSL _on all platforms_ unless `--with-openssl-dir` was passed by the user or if the system OpenSSL is already compatible. I suspect that this will have the most positive effect on macOS where Homebrew is really common and system OpenSSL/LibreSSL is useless.

@eregon I noticed that Truffleruby definitions invoke `use_homebrew_openssl` on Darwin, which unconditionally links to `openssl@1.1` from Homebrew or errors out if not present. Is Truffleruby also compatible with later OpenSSL versions, e.g. OpenSSL v3.x? Could we in theory use some of the new code from this PR to indicate a range of supported OpenSSL versions in Truffleruby definitions? That way, if I have `openssl@3` installed with Homebrew but not `openssl@1.1`, Truffleruby installation would still succeed.

mislav:
It looks like TruffleRuby looks up Homebrew OpenSSL in `$(brew --prefix)/opt/openssl@1.1` and `$(brew --prefix)/opt/openssl` (the latter is a symlink to the latest-installed `openssl@*` version), and otherwise this location can be overridden by setting OPENSSL_PREFIX environment variable.

@eregon With all that in mind, it sounds to me that `use_homebrew_openssl` has very little effect in TruffleRuby build definitions, since its main side effects are that it either [sets RUBY_CONFIGURE_OPTS_ARRAY](https://github.com/rbenv/ruby-build/blob/1d9b4e268f13b71a09c59a29dc0de5217236d130/bin/ruby-build#L1086) or errors out if `openssl@1.1` is not installed. However, since the user might have installed `openssl@3` or has specified their OPENSSL_PREFIX manually, the behavior of erroring out from `use_homebrew_openssl` might not be beneficial.

I would propose to remove `use_homebrew_openssl` step from TruffleRuby definitions and to instead extend `build_package_truffleruby` with extra Darwin-specific logic relating to OpenSSL. Thoughts?

eregon:
> I would propose to remove `use_homebrew_openssl` step from TruffleRuby definitions and to instead extend `build_package_truffleruby` with extra Darwin-specific logic relating to OpenSSL. Thoughts?

Yes, let's remove it.
I don't think we need any logic for this in ruby-build, it's best to let truffleruby handle as it knows better which opnessl versions it's compatible with.

eregon:
> Should there be an opt-out mechanism?

I think it would be safer, yes.
But it could probably be added later, and if there is no issue with the new default then not needed.

eregon (approved):
Looks great!

mislav:
> Feedback wanted: this PR introduces this syntax:
> 
> ```shell
> install_package openssl-1.1.1 "https://..." --if "needs_openssl 1.0.2 3.x.x"
> ```

Update: I went with this syntax in the end, which I find more readable and, most importantly, avoids the use of shell quotes:
```shell
install_package openssl-1.1.1 "https://..." --if needs_openssl:1.0.2-3.x.x
```

mislav:
@brandonfriess-stripe Not yet, but if you open a new issue detailing how the current mechanism (automatically linking to Homebrew) doesn't work for you, we will consider adding an opt-out mechanism.

I prefer to first understand potential negative ramifications of the current approach before adding options around it. Thank you!

Issue #278, opened 2013-02-06
Support Topaz


ethnt:
[Topaz](http://docs.topazruby.com/) is a fairly new implementation of Ruby built with Python on top of RPython. The maintainers have a [status page](http://docs.topazruby.com/en/latest/current-status/) in their docs indicating what they are missing from their implementation of `1.9.3`.

It would be cool to add support for Topaz within `ruby-build`, it would need to add the ability to work with Python's `virtualenv` and check for PyPy. I'll work on a rough draft to see what I can do.

PR #2294, opened 2023-11-06
Remove Topaz, Maglev

mislav:
Neither project has seen releases in the last 10 years, so I'm making a call that there is no strong reason for ruby-build to support them.

Ref. #278 https://github.com/rbenv/ruby-build/pull/1420


sferik:
😢

hsbt (approved):

PR #2296, opened 2023-11-07
Instruct Ruby to fail the build if openssl or psych are missing

### LINKED ISSUE #1415: DESTDIR setting fails
Hi,
I am trying to build Ruby for packaging with fpm, with these two steps:

PREFIX=/usr/local/ruby ./install.sh
MAKE_INSTALL_OPTS="DESTDIR=/tmp/ruby" ruby-build "2.7.0" "/usr/local/ruby"

This installs Ruby to e.g. `/tmp/usr/local/ruby/bin/ruby`, so far so good. However, `RUBY_BIN` is set to `/usr/local/ruby/bin`, not `/tmp/usr/local/ruby/bin`, so anything that calls it, for example `verify_openssl`, also fails, because the binary is in the wrong place.

It would be nice to have a global DESTDIR setting that modifies RUBY_BIN (which appears to be unsettable from the outside) and also modifies MAKE_INSTALL_OPTS.

### LINKED ISSUE #2241: Feature Request: Ability/Docs to Build for Target Locations Inaccessible to the Builder
Attempting to use `ruby-build` as an alternative to the terminal Arch Linux Ruby packaging (2y out of date on a rolling distro); and denting my desk and my forehead pretty badly trying to get the options set up correctly. Targeting 3.2.2 - saw in a previous issue/PR relating to `DESTDIR=` that the Ruby version apparently matters here (at least 1.9/2.x behaved differently). There _was_ a comment about replacing shebang strings in there, but there's ABI linkage as well and i'm a bit concerned _that_ might get broken/left inconsistent if i start messing with paths post-process.

Arch PKGBUILDs generally `build` in a source tree and then `install` to a target subtree which is then compressed for decompression (installation) under `/`. Building Ruby using the package path target results in the full path of the build environment's package path target being written to the resulting files and breaking when run after installation under `/` (vs `/var/abs/packagename/pkg/packagename/`). Trying to build with the various `RUBY_*` env vars set results in a failure to build the gems because _that_ part already expects Ruby to be in the `/` prefix:
```sh
build() {
  #export RUBY_PREFIX_PATH="/usr/local"
  RUBY_BUILD_BUILD_PATH="${srcdir}/build" \
  RUBY_BUILD_CACHE_PATH="${srcdir}/cache" \
  RUBY_MAKE_INSTALL_OPTS="DESTDIR=${srcdir}/ruby-$_pkgver" \
  RUBY_PREFIX_PATH="${srcdir}/ruby-$_pkgver/usr" \
  ruby-build $_pkgver /usr
}

package() {
  mv ruby-$_pkgver/* "${pkgdir}/"
  chown -R root:root "${pkgdir}/usr"
}
```
produces the following when built in its own Arch chroot:
```sh
Inspect or clean up the working tree at /build/packagename/src/build
Results logged to /tmp/ruby-build.20230819230130.806.log

Last 10 log lines:
                                    net-imap 0.3.4
                                    net-pop 0.1.2
                                    net-smtp 0.3.3
                                    matrix 0.4.2
                                    prime 0.1.2
                                    rbs 2.8.2
                                    typeprof 0.21.3
                                    debug 1.7.1
installing bundled gem cache:       /build/packagename/src/ruby-3.2.2/usr/local/lib/ruby/gems/3.2.0/cache
/usr/bin/ruby-build: line 1230: /usr/local/bin/ruby: No such file or directory
==> ERROR: A failure occurred in build().
    Aborting...
==> Removing chroot copy [/var/abs/chroot/packagename]...done
==> ERROR: Build failed
```
What is the correct way to use `ruby-build` by unprivileged builder-users in CI's/containers/chroots/etc to create packages which are to be installed by privileged users on target systems (with the same ABI) later?

mislav:
Normally, Ruby `make` step will print a warning about any missing extensions, but will not abort the build and instead proceed as normal.

Since Ruby installations without openssl or psych are essentially broken, ruby-build used to have a `verify_openssl` build step to test if the newly built Ruby can load these extensions, and print helpful information and abort the build on errors:

    Loading the Ruby openssl extension failed

The `verify_opensl` implementation was necessary to provide a good experience for ruby-build users, but was hacky and I would prefer to eliminate it. Fixes #2241, fixes #1415

**It appears that passing `--with-ext=openssl,psych` to the Ruby configure step marks those extensions as mandatory and fails the `make` process if they failed to build.** That is exactly the behavior we want, so this change enables the configure option for all Ruby builds. https://github.com/ruby/ruby/commit/b58a30e1c14e971adba4096104274d5d692492e9

```
*** Following extensions are not compiled:
openssl:
	Could not be configured. It will not be installed.
	/private/var/folders/6k/fwzzx67n163b0lfftyw3313h0000gn/T/ruby-build.20231107113317.26428.4N23Wf/ruby-3.2.2/ext/openssl/extconf.rb:101: OpenSSL library could not be found. You might want to use --with-openssl-dir=<dir> option to specify the prefix where OpenSSL is installed.
	Check ext/openssl/mkmf.log for more details.
psych:
	Could not be configured. It will not be installed.
	Check ext/psych/mkmf.log for more details.
*** Fix the problems, then remove these directories and try again if you want.
make[1]: *** [note] Error 1
make: *** [build-ext] Error 2
```

/cc @nobu @hsbt: as people who are more familiar with Ruby build process than myself, do you foresee any potential problems with using `./configure --with-ext=openssl,psych` by default when building any Ruby version, even old ones? From what I can see, the `--with-ext` flag seems to have existed for a long time, but only started having the behavior that we want from Ruby 2.5.0. Thank you!

The new default ruby-build behavior can be avoided by explicitly passing another `--with-ext` value, e.g.:
```sh
ruby-build 3.2.2 /tmp/ruby-3.2.2 -- --with-ext=+
```

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -673,6 +673,11 @@ build_package_standard_build() {
         package_option ruby configure --with-openssl-dir="/usr/local"
       fi
     fi
+    if [[ "$RUBY_CONFIGURE_OPTS ${RUBY_CONFIGURE_OPTS_ARRAY[*]}" != *--with-ext* &&
+          "$RUBY_CONFIGURE_OPTS ${RUBY_CONFIGURE_OPTS_ARRAY[*]}" != *--without-ext* ]]; then
+      # Fail the `make` step if any of these extensions were not compiled.
+      package_option ruby configure --with-ext=openssl,psych,+
+    fi
   fi
 
   ( if [ "${CFLAGS+defined}" ] || [ "${!PACKAGE_CFLAGS+defined}" ]; then
@@ -1214,48 +1219,9 @@ build_package_openssl() {
   fi
 }
 
-# Post-install check that the openssl extension was built.
-# TODO: explore replacing this implementation with scanning the `make` log
-# for the "Following extensions are not compiled" block.
+# Kept for backward compatibility with 3rd-party definitions.
 build_package_verify_openssl() {
-  local msg
-  msg="->$(print_command "$RUBY_BIN" -e "<SCRIPT>")"
-
-  colorize 36 "$msg"
-  echo
-
-  # shellcheck disable=SC2016
-  "$RUBY_BIN" -e '
-    manager = ARGV[0]
-    packages = {
-      "apt-get" => Hash.new {|h,k| "lib#{k}-dev" }.update(
-        "openssl" => "libssl-dev",
-        "zlib" => "zlib1g-dev"
-      ),
-      "yum" => Hash.new {|h,k| "#{k}-devel" }.update(
-        "yaml" => "libyaml-devel"
-      )
-    }
-    ext_name = Hash.new {|h,k| k }.update("yaml" => "psych")
-
-    failed = %w[openssl readline zlib yaml].reject do |lib|
-      begin
-        require lib
-      rescue LoadError => e
-        $stderr.puts "Loading the Ruby #{lib} extension failed (#{e})"
-        $stderr.puts "See the extension log at #{Dir.pwd}/ext/#{ext_name[lib]}/mkmf.log"
-      end
-    end
-
-    if failed.size > 0
-      $stderr.puts "ERROR: Ruby install aborted due to missing extensions"
-      $stderr.print "Try running `%s install -y %s` to fetch missing dependencies.\n\n" % [
-        manager,
-        failed.map { |lib| packages.fetch(manager)[lib] }.join(" ")
-      ] unless manager.empty?
-      exit 1
-    end
-  ' "$(basename "$(type -p yum apt-get | head -1)")"
+  true
 }
 
 # Kept for backward compatibility with 3rd-party definitions.
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -692,9 +692,20 @@ build_package_standard_build() {
       "${!PACKAGE_CONFIGURE_OPTS_ARRAY}" $CONFIGURE_OPTS ${!PACKAGE_CONFIGURE_OPTS}
   ) || return $?
 
+  local status=0
   # make -j <num_cpu_cores>
   # shellcheck disable=SC2086
-  capture_command "$MAKE" "${!PACKAGE_MAKE_OPTS_ARRAY}" $MAKE_OPTS ${!PACKAGE_MAKE_OPTS}
+  capture_command "$MAKE" "${!PACKAGE_MAKE_OPTS_ARRAY}" $MAKE_OPTS ${!PACKAGE_MAKE_OPTS} || status=$?
+
+  if [[ $status -ne 0 && -z $VERBOSE ]]; then
+    # Surface any extension building problems from `make` log to stderr.
+    # https://github.com/ruby/ruby/blob/HEAD/ext/extmk.rb
+    sed -n '/Following extensions are not compiled/,/Fix the problems/p' "$LOG_PATH" | \
+      sed '/remove these directories and try again/d' | \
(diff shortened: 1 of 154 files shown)


eregon:
I think given this is a flag that basically no end user would use and so probably receives little testing I would suggest to avoid using it in ruby-build, it's not so unlikely to cause issues.

For instance what if `--with-ext=openssl,psych` changed to mean only build those 2 exts and no other exts?

Verifying if openssl/psych is there after building is trivial and portable, that incurs no risks.

mislav:
> Verifying if openssl/psych is there after building is trivial and portable, that incurs no risks.

Good point about there being no guarantee that `--with-ext` will work consistently. However, for reasons mentioned above, it's not like `verify_openssl` was trivial by comparison. It required maintaining an inline Ruby script, it ran too late in the build process (after `make` completed and `make install` was over), and it sabotaged installs for people compiling with the DESTDIR setting. I would rather rely more on Ruby's own build implementation than add workarounds for this to ruby-build.

eregon:
Could we check the logs for things like:
```
openssl:
	Could not be configured. It will not be installed.
```
It would likely work quite well on all Ruby versions.

Or maybe even simpler/safer, to automatically scan for these lines and print them highlighted in the ruby-build output? So users can easily see which extensions have been skipped.

mislav:
> Could we check the logs for things like:
> 
> ```
> openssl:
> 	Could not be configured. It will not be installed.
> ```

That was my initial approach to solving this, before I discovered `--with-ext`. There are some hurdles that I've had with scanning the log file:

* Ruby build system seems to output ANSI colors for these "Could not be configured" reports even when output goes to a pipe and when NO_COLOR is set. This seems to be fixed in Ruby 3.3.0-dev, but for older Ruby versions we would have to first strip ANSI sequences before pattern-matching.

* In `ruby-build --verbose` mode,  the stderr and stdout of `make` command (as well as any other external command) is connected directly to the user's terminal, as there is no log file. This makes it harder to scan for particular output.

* Scanning the output for text following the "Following extensions are not compiled" section is also its own kind of brittle, since any regular expressions matching this output would be made obsolete if someone tweaked this text a bit in Ruby core. I'm having more confidence that a `./configure` flag would be more stable in the longer run than the `make` output format.

mislav:
> What does the flag do though on < 2.5?

I'm not sure! That is why I cc'ed some Ruby core folks above to ask for advice. 🙇

mislav:
I have updated this changeset into something shippable:

- I am now using the `--with-ext=openssl,psych,+` syntax (notice the plus character). This ensures that all the default extensions are still getting built even though `--with-ext` was passed. The "openssl" and "psych" extensions are still marked as "mandatory".

- "Following extensions are not compiled" section from the `make` log is now forwarded to stderr so that the user can spot the extremely common type of OpenSSL problems more directly:

mislav:
Adding another comment to bust GitHub's frontend cache. I am positive there has been timeline items on this thread newer than 2 days ago, but GitHub isn't showing it even after refreshes.

mislav:
@eregon Thank you for the in-depth reflection on this PR. I went back and ascertained that using `--with-ext` should be safe for all Ruby versions. I have also added some rudimentary log file scanning, but that is only done to surface errors better and not to fail the build based on the contents of the log, which I feel would be brittle.

I am very pleased that we now git rid of the long inline Ruby script that did the job of a single Ruby configuration flag.

anevsevra:
It breaks build for at least Ruby 2.4.10. Switching back to [v20231114](https://github.com/rbenv/ruby-build/tree/v20231114) allows to build 2.4.10 again.

PR #2311, opened 2023-11-15
Add ruby-dev definition

eregon:
* Currently the same as 3.3.0-dev but that name changes every year, so this makes it easy to install ruby dev/master.

I always wondered why this is missing, well, let's add it.
cc @nirvdrum who mentioned there is some non-trivial workaround at Shopify due to the lack of this.

### DIFF
--- a/share/ruby-build/ruby-dev
+++ b/share/ruby-build/ruby-dev
@@ -1,0 +1,2 @@
+install_package "openssl-3.1.4" "https://www.openssl.org/source/openssl-3.1.4.tar.gz#840af5366ab9b522bde525826be3ef0fb0af81c6a9ebd84caa600fea1731eee3" openssl --if needs_openssl:1.0.2-3.x.x
+install_git "ruby-master" "https://github.com/ruby/ruby.git" "master" autoconf standard_build standard_install_with_bundled_gems verify_openssl


mislav (commented):
I like that there is a definition that always means "Ruby master", but it also might be slightly confusing that we now have to identical definitions: `ruby-dev` and `3.3.0-dev` (or whichever the latest MAJOR.MINOR version Ruby is at).

Could we set up `ruby-dev` as an alias that resolves to the latest `MAJOR.MINOR.0-dev` definition? I've also been meaning to set up `ruby` as an alias for the latest Ruby stable version, so it can be used as an easy way to install latest Ruby.

eregon:
I think it's best to keep them separate, because e.g. once Ruby 3.3.0 is released the 3.3.0-dev file should be updated to use the `ruby_3_3` branch, and maybe we forget to add 3.4.0-dev immediately or so. Or maybe even before the release, if CRuby uses a release branch or so (not the case so far though), then 3.3.0-dev would already change.

Specifically,  ruby-dev should never need to change and always use master, while n.n.n-dev changes at least once.

There is also the question of whether it makes sense to even have e.g 3.4.0-dev before 3.4 is released, but that can be considered separately.

eregon:
Aside: the naming 3.2.0-dev is pretty strange given there is 3.2.2 now, but 3.2.0-dev is newer than 3.2.2.
Maybe it should be 3.2-dev? (not something I want to change in this PR though)

mislav:
> Maybe it should be 3.2-dev?

Agreed. Maybe that's a naming scheme we could go for `3.4-dev` onward.

hsbt:
>Agreed. Maybe that's a naming scheme we could go for 3.4-dev onward.

+1

How about this transition?

1. We add `ruby-dev` definition same as `3.3.0-dev` with this PR.
2. We will add `3.4-dev` after bumping ruby version to 3.4 on `ruby/ruby` repo.
3. We will rename all of `X.Y.0-dev` to `X.Y-dev`.

mislav:
That all sounds great, thank you.

> 3\. We will rename all of `X.Y.0-dev` to `X.Y-dev`.

Should the legacy names `X.Y.0-dev` still work with ruby-build going forward (for backwards compatibility), or is it fine if they error out? Since these are "dev" builds, I wouldn't mind breaking a few things in this transition. But yet again, I do not use these definitions regularly myself.

eregon:
I think it's very rare to use X.Y-dev after X.Y.0 is out, so I think it'd be fine to break that, and if it proves to affect too many users then we could do some kind of aliasing.

PR #1098, opened 2017-06-09
invoke make update-gems before make install with ruby-trunk

hsbt:
Fixes #1093 

Added `make update-gems` command before `make install` when installing to 2.5.0-dev named "ruby-trunk"

/cc @k0kubun

I'm not a familiar with shell script. Can I handle ruby-trunk, ruby-2.4.0-dev and ruby-2.3.0-dev smartly?

PR #2318, opened 2023-11-30
Introduce "with_bundled_gems" build step for dev rubies and unify "standard"

mislav:
This merges "standard_build" and "standard_install" build steps into one again and defines a "standard_install_with_bundled_gems" build step that just invokes "standard" with the addition of `update-gems` & `extract-gems`.

The goal is to have a single `standard` step again instead of maintaining it as two distinct `build_package_standard_build` and `build_package_standard_install` functions.

A runtime difference that will be a result of this change is that three separate invocations `make update-gems`, `make extract-gems`, and `make install` are now executed as one: `make update-gems extract-gems install`.

/cc @hsbt as this reverts parts of https://github.com/rbenv/ruby-build/pull/1098

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -634,7 +634,7 @@ build_package_warn_unsupported() {
   } >&2
 }
 
-build_package_standard_build() {
+build_package_standard() {
   local package_name="$1"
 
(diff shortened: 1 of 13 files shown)


hsbt:
👍

Issue #2300, opened 2023-11-08
Old Ruby, mismatched OpenSSL


pointlessone:
### Steps to reproduce the behavior

`rbenv install 2.6.10`

### Details

The build fails with mismatched OpenSSL error.

- I have both openssl@1.1 and openssl@3 installed in homebrew
- I run macOS Sonoma 14.2
- I have ruby-build installed through homebrew, version 20231107

ruby-build seem to correctly detect openssl@1.1:

```
-> ./configure "--prefix=$HOME/.rbenv/versions/2.6.10" --enable-shared --with-readline-dir=/usr/local/opt/readline --with-libyaml-dir=/usr/local/opt/libyaml --with-gmp-dir=/usr/local/opt/gmp --with-openssl-dir=/usr/local/opt/openssl@1.1
```

But this somehow gets lost during compilation phase:

```
compiling ossl.c
ossl.c:311:14: error: 'ERR_get_error_line_data' is deprecated [-Werror,-Wdeprecated-declarations]
        while ((e = ERR_get_error_line_data(&file, &line, &data, &flags))) {
                    ^
/usr/local/Cellar/openssl@3/3.1.4/include/openssl/err.h:424:1: note: 'ERR_get_error_line_data' has been explicitly marked deprecated here
OSSL_DEPRECATEDIN_3_0
^
/usr/local/Cellar/openssl@3/3.1.4/include/openssl/macros.h:193:49: note: expanded from macro 'OSSL_DEPRECATEDIN_3_0'
#   define OSSL_DEPRECATEDIN_3_0                OSSL_DEPRECATED(3.0)
                                                ^
/usr/local/Cellar/openssl@3/3.1.4/include/openssl/macros.h:62:52: note: expanded from macro 'OSSL_DEPRECATED'
#     define OSSL_DEPRECATED(since) __attribute__((deprecated))
                                                   ^
1 error generated.
```

### Logs

[Full log](https://gist.github.com/pointlessone/77756e5058ce8a183506c57c366f10a5)

mislav:
Thank you for reporting. It looks like ruby-build does its job correctly, but Ruby 2.6 ignores the `--with-openssl-dir` passed to it. If you temporarily `brew unlink openssl@3`, does the problem go away?

Was this a problem for you before the latest release of ruby-build?

mislav:
Some more information here: https://github.com/rbenv/ruby-build/discussions/2125#discussioncomment-7516993

[...]

PR #2322, opened 2023-12-07
Fix linking to older OpenSSL for Ruby < 2.7.7

mislav:
Explicitly set PKG_CONFIG_PATH when `--with-openssl-dir` is used for older Rubies. Otherwise, Ruby will attempt to link to the latest OpenSSL found by pkg-config, which could be incompatible.

* https://github.com/rbenv/ruby-build/discussions/2320#discussioncomment-7787227
* https://github.com/rbenv/ruby-build/issues/2300#issuecomment-1803762801

Even though these Ruby versions are EOL, people are still inexplicably using them, and this problem comes up over and over again in user support questions.

```sh
# openssl gem version that fixes this is 2.1.4
$ curl -fsSL https://cache.ruby-lang.org/pub/ruby/2.7/ruby-2.7.6.tar.bz2 | \
    tar jxO -- \*/ext/openssl/openssl.gemspec | grep -w version
  spec.version       = "2.1.3"
```


eregon:
Nice, PKG_CONFIG_PATH is what truffleruby internally set as well before it was fixed in the openssl gem.

PR #2323, opened 2023-12-11
Add Ruby 3.3.0-rc1

casperisfine:
https://www.ruby-lang.org/en/news/2023/12/11/ruby-3-3-0-rc1-released/


casperisfine:
The CI failed because the tarballs weren't uploaded yet. I pinged Naruse and it's now OK, but I don't have the permission to restart the build.

PR #2328, opened 2023-12-25
Add Ruby 3.4.0-dev

casperisfine:


hsbt:
@casperisfine Thanks. But we use `3.4-dev` instead of `3.4.0-dev` since Ruby 3.4. see https://github.com/rbenv/ruby-build/pull/2311#issuecomment-1820206672

I will merge https://github.com/rbenv/ruby-build/pull/2329.

casperisfine:
Ah, thank you, I wasn't aware.

Issue #2330, opened 2023-12-25
Can we skip installing readline when Ruby version is greater than 3.3.0?


okuramasafumi:
https://www.ruby-lang.org/en/news/2023/12/25/ruby-3-3-0-released/

> We no longer need to install libraries like libreadline or libedit.

I installed Ruby 3.3.0 with ruby-build and noticed that it uses readline.

```
==> Installing ruby-3.3.0...
ruby-build: using readline from homebrew
ruby-build: using libyaml from homebrew
ruby-build: using gmp from homebrew
-> ./configure "--prefix=$HOME/.rbenv/versions/3.3.0" --with-openssl-dir=/opt/homebrew/opt/openssl@3 --enable-shared --with-readline-dir=/opt/homebrew/opt/readline --with-libyaml-dir=/opt/homebrew/opt/libyaml --with-gmp-dir=/opt/homebrew/opt/gmp --with-ext=openssl,psych,+
-> make -j 10
-> make install
==> Installed ruby-3.3.0 to /Users/okuramasafumi/.rbenv/versions/3.3.0
```

This `--with-readline-dir=/opt/homebrew/opt/readline` can be skipped for Ruby 3.3.0 and greater as far as I understand.

mislav:
Thank you for raising the question. I just want to first clarify that ruby-build never installs any dependencies; it just potentially discovers them elsewhere on the system and enables the Ruby build process to link to them by passing `--with-readline-dir=` to the configure process.

Now the question is, since Ruby 3.3 apparently doesn't need external libreadline, does `--with-readline-dir=` actually have any effect or if it's simply ignored. If it is ignored, then I suggest that ruby-build keeps passing it because it still helps older Ruby versions.

hsbt:
>If it is ignored, then I suggest that ruby-build keeps passing it because it still helps older Ruby versions.

That is. It's better to remove `--with-readline-dir` option from Ruby 3.3. But It's not effect with build process on Ruby 3.3.

PR #2333, opened 2024-01-07
Skip passing `--with-readline-dir` for Ruby 3.3+

mislav:
Apparently, `--with-readline-dir` is no longer needed starting from Ruby 3.3.0. It doesn't seem harmful to pass it, but since it might cause confusion to see the implicit value being passed as part of ruby-build output, it might be a good idea to only conditionally pass it for older rubies.

Fixes https://github.com/rbenv/ruby-build/issues/2330


hsbt (approved):
Thanks!

PR #2356, opened 2024-03-18
Switch to EA builds from graalvm/graal-languages-ea-builds for truffleruby+graalvm-dev

eregon:
* See https://github.com/graalvm/graal-languages-ea-builds
* These are based on Oracle GraalVM.
* cc @nirvdrum

### DIFF
--- a/share/ruby-build/truffleruby+graalvm-dev
+++ b/share/ruby-build/truffleruby+graalvm-dev
@@ -1,19 +1,26 @@
 platform="$(uname -s)-$(uname -m)"
 case $platform in
 Linux-x86_64)
-  install_package "truffleruby+graalvm-dev" "https://github.com/graalvm/graalvm-ce-dev-builds/releases/latest/download/truffleruby-community-jvm-dev-linux-amd64.tar.gz" truffleruby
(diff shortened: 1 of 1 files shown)


share/ruby-build/truffleruby+graalvm-dev
```diff
@@ -1,19 +1,26 @@
...
 *)
   colorize 1 "Unsupported platform: $platform"
   return 1
   ;;
 esac
+
+pushd "$BUILD_PATH" >/dev/null
+http get $url url.txt
+url=$(<url.txt)
+popd
```
eregon, line 24: The `pushd` is needed because `aria2c -o /absolute/path` doesn't work, it actually writes in `/cwd//absolute/path` which is the same as `/cwd/absolute/path`.
It would be tempting to just use `curl` to avoid the extra file, but that doesn't work on systems without `curl`.

rwstauner: Yeah, the ruby-build script only ever calls `http` with a path-less filename so it just downloads to the CWD and eventually removes the file.
I worked around this once by just using a tempfile: https://github.com/Shopify/ruby-definitions/pull/36

PR #2361, opened 2024-03-22
Simplify `truffleruby+graalvm-dev` build definition re: download redirect

mislav:
The `http get <url> <destfile>` utility had a bug with aria2c downloader where it couldn't properly save to destfile if it was an absolute path.

I have tried having `http get <url> -` consistently print the downloaded file to stdout, but this conflicted with the output of `log_command` (which is also to stdout) so for now let's keep using the temporary file to resolve manual URL redirects.

Followup to https://github.com/rbenv/ruby-build/pull/2356 /cc @rwstauner

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -423,8 +423,15 @@ http_head_aria2c() {
 }
 
 http_get_aria2c() {
-  # shellcheck disable=SC2086
-  log_command aria2c --allow-overwrite=true --no-conf=true --console-log-level=warn --stderr $ARIA2_OPTS -o "$2" "$1" 2>&3
+  local destfile="$2"
+  if [[ $destfile == /* ]]; then
+    # the "-o" option to aria2c cannot be an absolute path, but we can achieve that with "--dir"
+    # shellcheck disable=SC2086
+    log_command aria2c --allow-overwrite=true --no-conf=true --console-log-level=warn --stderr $ARIA2_OPTS --dir "${destfile%/*}" -o "${destfile##*/}" "$1" 2>&3
+  else
+    # shellcheck disable=SC2086
+    log_command aria2c --allow-overwrite=true --no-conf=true --console-log-level=warn --stderr $ARIA2_OPTS -o "$destfile" "$1" 2>&3
+  fi
 }
 
 http_head_curl() {
--- a/share/ruby-build/truffleruby+graalvm-dev
+++ b/share/ruby-build/truffleruby+graalvm-dev
@@ -1,16 +1,16 @@
 platform="$(uname -s)-$(uname -m)"
 case $platform in
 Linux-x86_64)
-  url="https://raw.githubusercontent.com/graalvm/graal-languages-ea-builds/main/truffleruby/versions/latest-jvm-linux-amd64.url"
+  url="https://github.com/graalvm/graal-languages-ea-builds/raw/HEAD/truffleruby/versions/latest-jvm-linux-amd64.url"
   ;;
 Linux-aarch64)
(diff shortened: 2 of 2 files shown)


share/ruby-build/truffleruby+graalvm-dev
```diff
@@ -1,26 +1,24 @@
 platform="$(uname -s)-$(uname -m)"
 case $platform in
 Linux-x86_64)
-  url="https://raw.githubusercontent.com/graalvm/graal-languages-ea-builds/main/truffleruby/versions/latest-jvm-linux-amd64.url"
+  url="https://github.com/graalvm/graal-languages-ea-builds/raw/HEAD/truffleruby/versions/latest-jvm-linux-amd64.url"
```
mislav, line 4: Let's stick to using `github.com` URLs instead of `raw.githubusercontent.com` since the latter is an implementation detail of GitHub which might change in the future.

bin/ruby-build
```diff
@@ -423,8 +423,15 @@ http_head_aria2c() {
...
-  log_command aria2c --allow-overwrite=true --no-conf=true --console-log-level=warn --stderr $ARIA2_OPTS -o "$2" "$1" 2>&3
+  local destfile="$2"
+  if [[ $destfile == /* ]]; then
+    # the "-o" option to aria2c cannot be an absolute path, but we can achieve that with "--dir"
+    # shellcheck disable=SC2086
+    log_command aria2c --allow-overwrite=true --no-conf=true --console-log-level=warn --stderr $ARIA2_OPTS --dir "${destfile%/*}" -o "${destfile##*/}" "$1" 2>&3
+  else
+    # shellcheck disable=SC2086
+    log_command aria2c --allow-overwrite=true --no-conf=true --console-log-level=warn --stderr $ARIA2_OPTS -o "$destfile" "$1" 2>&3
+  fi
```
rwstauner, line 434: Makes sense to me, and confirmed it works with "file", "subdir/file" and "/abs/path".
You could reduce a bit of duplication:
```suggestion
  local destfile="$2"
  local args=()

  if [[ $destfile == /* ]]; then
    # the "-o" option to aria2c cannot be an absolute path, but we can achieve that with "--dir"
    args+=(--dir "${destfile%/*}" -o "${destfile##*/}")
  else
    args+=(-o "$destfile")
  fi

  # shellcheck disable=SC2086
  log_command aria2c --allow-overwrite=true --no-conf=true --console-log-level=warn --stderr $ARIA2_OPTS "${args[@]}" "$1" 2>&3
```

eregon (approved):
LGTM, I checked locally and it works with `aria2c`.

I don't know well temp files in Bash so I was happy with my solution in https://github.com/rbenv/ruby-build/pull/2356.
This looks fine too. It might not delete the tempfile until repeat if the installation fails but that seems OK.

FWIW I also tried to pass `-` as the file but aria2c creates a file named `-`, unlike wget/curl which behave as expected and print to stdout, oh well.

bin/ruby-build
mislav: @rwstauner Thanks for the deduplication suggestion, but I find that I prefer seeing the full `aria2c` invocation line in both instances even if there is a little duplication, mostly for the ease of reading.

mislav:
> FWIW I also tried to pass `-` as the file but aria2c creates a file named `-`, unlike wget/curl which behave as expected and print to stdout, oh well.

Yeah, that inconsistency worries me. I think a future change to our `http get <url> <dest>` helper should either prohibit the `-` value or make it work consistently across all downloaders. But for now, I'm happy with at least having aria2c download respect the destination file argument consistently.

PR #2370, opened 2024-04-23
Add CRuby security-fix releases 3.0.7, 3.1.5, 3.2.4, and 3.3.1

kpfleming:
https://www.ruby-lang.org/en/news/2024/04/23/ruby-3-0-7-released/

https://www.ruby-lang.org/en/news/2024/04/23/ruby-3-1-5-released/

https://www.ruby-lang.org/en/news/2024/04/23/ruby-3-2-4-released/

https://www.ruby-lang.org/en/news/2024/04/23/ruby-3-3-1-released/

### DIFF
--- a/share/ruby-build/3.0.7
+++ b/share/ruby-build/3.0.7
@@ -1,0 +1,2 @@
+install_package "openssl-1.1.1w" "https://www.openssl.org/source/openssl-1.1.1w.tar.gz#cf3098950cb4d853ad95c0841f1f9c6d3dc102dccfcacd521d93925208b76ac8" openssl --if needs_openssl:1.0.1-1.x.x
+install_package "ruby-3.0.7" "https://cache.ruby-lang.org/pub/ruby/3.0/ruby-3.0.7.tar.gz#2a3411977f2850431136b0fab8ad53af09fb74df2ee2f4fb7f11b378fe034388" enable_shared standard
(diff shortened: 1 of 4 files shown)


colszowka:
Thanks @kpfleming @hsbt!

I have a short question - why are we going with OpenSSL 3.1.4 here specifically, instead of 3.1.5, or the newer 3.2 or 3.3 families? https://www.openssl.org/source/ lists 3.3.0 as the newest version, released ~2 weeks ago, with 3.1.5 released this january

eregon:
Because https://github.com/rbenv/ruby-build/blob/346c0dccaee436a69ddd3bcf0498566982260e19/script/update-cruby#L28 and e.g. https://github.com/rbenv/ruby-build/blob/346c0dccaee436a69ddd3bcf0498566982260e19/share/ruby-build/3.3.0#L1

In general OpenSSL version are updated in their own PR (and there is https://github.com/rbenv/ruby-build/blob/master/script/update-openssl)

eregon:
IOW, please make a PR if you'd like to update openssl ;)

colszowka:
@eregon Got it, thanks! I'm not sure though which version to go with, is it safe to go all the way to 3.3.0 from 3.1.4, or just 3.1.5? I'm not familiar with the versioning policies and compatibility of OpenSSL

mislav:
@colszowka Thanks for pinging about this. I'm making a PR to upgrade to openssl 3.1.5.

> why are we going with OpenSSL 3.1.4 here specifically, instead of 3.1.5, or the newer 3.2 or 3.3 families?

We could also update our Ruby build formulae to depend on OpenSSL 3.2 or 3.3, but since we've been depending on OpenSSL 3.1.x for a while now and since that branch is still maintained by the OpenSSL team, my vote is to stick with the 3.1.x branch for as long as it's maintained or until there is a concrete need to upgrade to OpenSSL 3.2/3.3, e.g. to gain access to a newer feature.

PR #2386, opened 2024-05-21
Mark Ruby 3.0.7 as unsupported

CamJN:
This one was missed in https://github.com/rbenv/ruby-build/pull/2307 or was added after, hopefully this removes ruby 3.0.7 from `rbenv install -l` unless you list-all


hsbt:
👍

PR #2388, opened 2024-05-22
Mark Ruby 3.0 as EOL

mislav:
Ruby 3.0 reached End of Life on 23 Apr 2024.

Changeset generated with `script/update-eol`

Followup to https://github.com/rbenv/ruby-build/pull/2307


hsbt (approved):
👍

PR #2389, opened 2024-05-22
Truncate teeny version from dev definitions

hsbt:
I remember https://github.com/rbenv/ruby-build/pull/2311#issuecomment-1820206672. We don't need to add `.0` version to `x.y.0-dev` definitions.

I'm not sure how people use this definition usually. But I believe this change is low impact for users.


eregon (approved):

eregon:
Thanks, looks good

PR #2392, opened 2024-05-24
Clarify RUBY_BUILD_ROOT env var

### LINKED ISSUE #2391: RUBY_BUILD_ROOT docs misleading
### Steps to reproduce the behavior

1. have directory containing build definitions at `/tmp/rby/<defn_files>`
2. run: `RUBY_BUILD_ROOT=/tmp/rby ruby-build --list`

### Expected vs. actual behavior

Expected: listing of files in `/tmp/rby`
Actual: nothing listed

The docs state that `RUBY_BUILD_ROOT` may point to a custom directory that contains build definition files. Specifically, that the default is "share/ruby-build".

> Custom build definition directory. (Default: share/ruby-build)

In actuality, `RUBY_BUILD_ROOT` must point to a directory that _contains_ a `share/ruby-build/` directory that may contain definition files. Note how it is used by concatenating with "share/ruby-build": https://github.com/rbenv/ruby-build/blob/210578da4c2f7252fd5b60227d124e28f20c590b/bin/ruby-build#L1367

The default is in fact found by resolving where the ruby-build executable is, which is assumed to have a `share/ruby-build` dir.

This has been the behavior for as long as RUBY_BUILD_ROOT as been configurable: https://github.com/rbenv/ruby-build/commit/ff75ca72045ce54b4c47d1b5c51e3d921cc63c96

Given that the behavior has not changed, and that changing it now would be a breaking change, I propose that only the readme be updated to clarify the true behavior. (I'll open a PR soon.)

jasonkarns:
The RUBY_BUILD_ROOT itself does not default to share/ruby-build/ but rather must point to a directory that has definitions under `share/ruby-build/`.

The misleading verbiage was introduced: https://github.com/rbenv/ruby-build/commit/af2df4e74a305e4018331531e43e5d9276bb6875

Original phrasing introduced: https://github.com/rbenv/ruby-build/commit/ff75ca72045ce54b4c47d1b5c51e3d921cc63c96

Could still use some rephrasing I think. It's awkward to explain the env var must point to a directory that itself contains a particular path; without implying that the env var itself contains the "share/ruby-build/" path.

fixes #2391

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ The build process may be configured through the following environment variables:
 | `RUBY_BUILD_MIRROR_URL`         | Custom mirror URL root.                                                                          |
 | `RUBY_BUILD_MIRROR_PACKAGE_URL` | Custom complete mirror URL (e.g. http://mirror.example.com/package-1.0.0.tar.gz).                |
 | `RUBY_BUILD_SKIP_MIRROR`        | Bypass the download mirror and fetch all package files from their original URLs.                 |
-| `RUBY_BUILD_ROOT`               | Custom build definition directory. (Default: `share/ruby-build`)                                 |
+| `RUBY_BUILD_ROOT`               | Custom build definition directory containing `share/ruby-build/<defs>`. (Default: ruby-build's installation prefix ) |
 | `RUBY_BUILD_TARBALL_OVERRIDE`   | Override the URL to fetch the ruby tarball from, optionally followed by `#checksum`.             |
 | `RUBY_BUILD_DEFINITIONS`        | Additional paths to search for build definitions. (Colon-separated list)                         |
 | `CC`                            | Path to the C compiler.                                                                          |
--- a/README.md
+++ b/README.md
@@ -83,9 +83,9 @@ The build process may be configured through the following environment variables:
 | `RUBY_BUILD_MIRROR_URL`         | Custom mirror URL root.                                                                          |
 | `RUBY_BUILD_MIRROR_PACKAGE_URL` | Custom complete mirror URL (e.g. http://mirror.example.com/package-1.0.0.tar.gz).                |
 | `RUBY_BUILD_SKIP_MIRROR`        | Bypass the download mirror and fetch all package files from their original URLs.                 |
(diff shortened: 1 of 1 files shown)


jasonkarns:
*edit* as the original PR author, I can't approve. But I do 👍 your changes @mislav .

mislav:
> One thing that `RUBY_BUILD_ROOT` makes useful, is the ability to force a homebrew-installed ruby-build to act as if it's running from, say, `~/code/rbenv/ruby-build`

Could `export RUBY_BUILD_DEFINITIONS=~/code/rbenv/ruby-build/share/ruby-build` accomplish the same?

mislav (approved):

jasonkarns:
I would say not, because the intention when testing things is to operate as if the "working clone" of ruby-build were the only source of definitions. 

Granted, there are arguably better setups for testing; I'm not holding this approach as a model. But I don't think that appending to the definitions path fits the same use case (depending on what exactly is being tested).

eregon:
You can see in https://github.com/rbenv/ruby-build/blob/db6c17529bf14de3985c098123345c15aa3d6a8b/bin/ruby-build#L1367 the only usage of RUBY_BUILD_ROOT in code. So it does nothing more than RUBY_BUILD_DEFINITIONS (AFAIK).
So I think we should deprecate it and remove it, aka remove it from the README as a way to deprecate.

> My approach for contributing to rbenv repos has changed over time, but presently I have rbenv and ruby-build installed via homebrew. And then separately (in my ~/code dir), I have them cloned for development/contributing.

What I do for ruby-build is I have `~/code/ruby-build/bin/ruby-build` in PATH, that's much simpler IMO.

eregon:
Ah, you're right both set of definitions would be available when setting RUBY_BUILD_DEFINITIONS.
I think it's much cleaner to use a single clone of ruby-build than trying to mix them though.

mislav:
> Granted, there are arguably better setups for testing; I'm not holding this approach as a model. But I don't think that appending to the definitions path fits the same use case

I'd say: let's keep the current RUBY_BUILD_ROOT feature for use-cases like yours, but instead of better documenting it, lets deprecate it in documentation so that other people don't use it.

I was tempted to remove it from documentation completely, but I think it's better to leave it visible for posterity.

PR #2397, opened 2024-05-30
Fix 3.1.6 version definition

rafaelfranca:
There is an extra quote and it was causing a failure:

```
/home/vscode/.rbenv/plugins/ruby-build/share/ruby-build/3.1.6: line 2: unexpected EOF while looking for matching `"'
```


hsbt:
🙏 Thanks

hsbt:
I just released with your patch. https://github.com/rbenv/ruby-build/releases/tag/v20240530.1

PR #2408, opened 2024-06-11
Fix package name for openssl-3.0.14

mislav:
Followup to https://github.com/rbenv/ruby-build/pull/2407

@hsbt There was a small renaming issue. I've used `script/update-openssl 3.0.14 <SHA>` to fix it.


hsbt:
>I've used script/update-openssl 3.0.14 <SHA> to fix it.

I didn't know that. Thanks 🙏

PR #2428, opened 2024-07-26
Add Ruby 3.2.5

sferik:
* Ruby 3.2.5 Released
https://www.ruby-lang.org/en/news/2024/07/26/ruby-3-2-5-released/


mislav:
Thank you!!

Issue #2279, opened 2023-10-21
tr: Illegal byte sequence on macOS when installing JRuby


eregon:
### Steps to reproduce the behavior

`ruby-build --verbose jruby-9.4.4.0 ~/.rubies/jruby-9.4.4.0` on macOS

### Expected vs. actual behavior

No such error

### Logs

https://github.com/ruby/ruby-builder/actions/runs/6596732799/job/17923080695#step:19:49
```
ruby-build --verbose $RUBY_BUILD_RUBY_NAME $PREFIX
/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/ruby-build.20231021101301.2735.PEZ7jH ~/work/ruby-builder/ruby-builder
Downloading jruby-dist-9.4.4.0-bin.tar.gz...

10/21 10:13:02 [NOTICE] Downloading 1 item(s)

10/21 10:13:02 [NOTICE] Download complete: /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/ruby-build.20231021101301.2735.PEZ7jH/6ab12670afd8e5c8ac9305fabe42055795c5ddf9f8e8f1a1e60e260f2d724cc0

Download Results:
gid   |stat|avg speed  |path/URI
======+====+===========+=======================================================
911eb4|OK  |       0B/s|/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/ruby-build.20231021101301.2735.PEZ7jH/6ab12670afd8e5c8ac9305fabe42055795c5ddf9f8e8f1a1e60e260f2d724cc0

Status Legend:
(OK):download completed.
-> https://dqw8nmjcqpjn7.cloudfront.net/6ab12670afd8e5c8ac9305fabe42055795c5ddf9f8e8f1a1e60e260f2d724cc0

10/21 10:13:02 [NOTICE] Downloading 1 item(s)

10/21 10:13:02 [NOTICE] Allocating disk space. Use --file-allocation=none to disable it. See --file-allocation option in man page for more details.

10/21 10:13:03 [NOTICE] Download complete: /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/ruby-build.20231021101301.2735.PEZ7jH/jruby-9.4.4.0.tar.gz

Download Results:
gid   |stat|avg speed  |path/URI
======+====+===========+=======================================================
[...]

PR #2429, opened 2024-07-30
Clear LC_ALL for tr call

headius:
This resolves additional issues when tr is called in a shell that has set LC_ALL to a specific encoding. In that case, the tr may fail to ignore binary files, expecting that they are in some encoding.

See https://github.com/rbenv/ruby-build/issues/2279


headius:
@eregon @mislav I took a stab at this based on the discussion. It should clear `LC_ALL` in the `tr` call, which I believe resolves the encoding error on environments with an existing `LC_ALL` env.

Please check my work.

mislav (approved):
Thank you for following up

PR #2430, opened 2024-08-01
Auto update Ruby definition with Ruby releases

hsbt:
Background:

I and release manager of Ruby improve release workflow recently. I make this manual workflow to automated workflow. 

After Ruby release, https://github.com/ruby/ruby/pull/11286 notify to related repositories like docker image, snap package and etc.

I would like to create new definition of ruby-build with above notification. This PR makes it with repository_dispatch feature. After that, we can update definition with simple REST call like:

```
curl -L -X POST \
          -H "Authorization: Bearer xxx" \
          -H "Accept: application/vnd.github+json" \
          -H "X-GitHub-Api-Version: 2022-11-28" \
          https://api.github.com/repos/rbenv/ruby-build/dispatches \
          -d '{"event_type": "update-ruby", "client_payload":{"ruby_version":"3.3.4", "openssl_version":"3.0.14"}}'
```

### DIFF
--- a/.github/workflows/update-ruby.yml
+++ b/.github/workflows/update-ruby.yml
@@ -1,0 +1,48 @@
+name: Update Ruby definitions
+
+on:
+  workflow_dispatch:
+    inputs:
+      ruby_version:
+          description: 'Ruby version'
+          required: true
+          default: '3.3.4'
+      openssl_version:
+            description: 'OpenSSL version'
+            required: true
+            default: '3.0.14'
+
+  repository_dispatch:
+    types: [update-ruby]
+
+jobs:
+  update-ruby:
+    name: Update Ruby definitions
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Store Ruby version
+        run: echo "RUBY_VERSION=${{ github.event.client_payload.ruby_version || github.event.inputs.ruby_version }}" >> $GITHUB_ENV
+
+      - name: Store ABI version
+        run: echo "ABI_VERSION=$(echo ${{ env.RUBY_VERSION }} | cut -d '.' -f 1-2)" >> $GITHUB_ENV
+
+      - name: Store OpenSSL version
+        run: echo "OPENSSL_VERSION=${{ github.event.client_payload.openssl_version || github.event.inputs.openssl_version }}" >> $GITHUB_ENV
+
+      - name: Run script/update-cruby
+        run: |
+          curl -sL https://cache.ruby-lang.org/pub/ruby/${{ env.ABI_VERSION }}/ruby-${{ env.RUBY_VERSION }}.tar.gz
+          curl -sL https://www.openssl.org/source/openssl-${{ env.OPENSSL_VERSION }}.tar.gz
+          script/update-cruby ${{ env.RUBY_VERSION }} ${{ env.OPENSSL_VERSION }} .
+
+      - name: commit and push
+        run: |
+          git config user.name "GitHub Actions Bot"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git add .
+          git commit -m "Added ${{ env.RUBY_VERSION }} with OpenSSL ${{ env.OPENSSL_VERSION }}"
+          git push
+        env:
+          GITHUB_TOKEN: ${{ secrets.COMMITTER_TOKEN }}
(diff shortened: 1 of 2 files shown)


script/update-cruby
```diff
@@ -24,7 +26,17 @@ else
   exit 1
 fi
 
+openssl_url="https://www.openssl.org/source/openssl-${openssl_version}.tar.gz"
+if command -v sha256sum >/dev/null; then
+  openssl_sha256=$(sha256sum "$release_directory/$openssl_basename" | cut -d ' ' -f 1)
```
eregon, line 31: I think you need to download the openssl archive before this line, otherwise it won't be there

eregon: Ah it's done in the workflow, I see. That doesn't seem practical if this script is ever run manually though.

.github/workflows/update-ruby.yml
```diff
@@ -0,0 +1,49 @@
...
+          curl -sL https://cache.ruby-lang.org/pub/ruby/${{ env.ABI_VERSION }}/ruby-${{ env.RUBY_VERSION }}.tar.gz
+          curl -sL https://www.openssl.org/source/openssl-${{ env.OPENSSL_VERSION }}.tar.gz
+          script/update-cruby ${{ env.RUBY_VERSION }} ${{ env.OPENSSL_VERSION }} .
+
+      - name: commit and push
+        run: |
+          git config user.name "GitHub Actions Bot"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git add .
+          git commit -m "rake sync"
```
eregon, line 46: `"Add Ruby ${{ env.RUBY_VERSION }}"` is a much better message

.github/workflows/update-ruby.yml
```diff
@@ -0,0 +1,49 @@
...
+          git push
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
eregon, line 49: Because this is using the default GITHUB_TOKEN [it will not trigger other workflows](https://github.com/orgs/community/discussions/25702#discussioncomment-3248819) like https://github.com/rbenv/ruby-build/blob/master/.github/workflows/mirror.yml
That seems a problem, as that one should run.
It will also not trigger the ci workflow, which doesn't seem good.
That could result in the archive not being mirrored or having an incorrect checksum, etc.

A solution is to use a PAT for this.

Or instead of pushing to create a PR, I think that's safer and clearer what's going on.
That's what I'm using for setup-ruby:
https://github.com/ruby/ruby-builder/blob/master/.github/actions/create-pr-to-setup-ruby/action.yml

hsbt: Thanks, I use https://github.com/rbenv/ruby-build/blob/master/.github/workflows/release.yml#L19 for this.

hsbt:
This and hotfix are working fine: https://github.com/rbenv/ruby-build/commit/d22fa95a6e4c77945304c16ebe0d9513fec98cfb 

Thanks @eregon

.github/workflows/update-ruby.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Store ABI version
+        run: echo "ABI_VERSION=$(echo ${{ env.RUBY_VERSION }} | cut -d '.' -f 1-2)" >> $GITHUB_ENV
+
+      - name: Store OpenSSL version
+        run: echo "OPENSSL_VERSION=${{ github.event.client_payload.openssl_version || github.event.inputs.openssl_version }}" >> $GITHUB_ENV
+
+      - name: Run script/update-cruby
+        run: |
+          curl -sL https://cache.ruby-lang.org/pub/ruby/${{ env.ABI_VERSION }}/ruby-${{ env.RUBY_VERSION }}.tar.gz
+          curl -sL https://www.openssl.org/source/openssl-${{ env.OPENSSL_VERSION }}.tar.gz
```
mislav, line 37: Are these lines supposed to do or verify something? If we wanted to download these URLs to local files, I do not think curl will do it unless `-O` was used?

hsbt: Yes, I added `-O` at https://github.com/rbenv/ruby-build/commit/e12c32f71245d4238c8e085ad7484bbcb972f602

opti:
@hsbt can this auto update be triggered for the already released ruby v3.3.5?

PR #2432, opened 2024-08-14
Allow "nonexistent version" test to work without brew

piperswe:
Currently, this test fails if Homebrew isn't present as it expects a
Homebrew-specific output. This commit varies the expected output based
on whether Homebrew is installed.

This patch was originally created for the Debian package for ruby-build:
https://salsa.debian.org/ruby-team/ruby-build/-/blob/607093e5198cc75ffa8e9a3b551843be5f2f6d93/debian/patches/0001-Fix-nonexistent-version-test-to-work-regardless-of-e.patch

### DIFF
--- a/test/rbenv.bats
+++ b/test/rbenv.bats
@@ -88,11 +88,19 @@ OUT
 }
 
 @test "nonexistent version" {
+  stub_git_dir=
+  if [ ! -d "${BATS_TEST_DIRNAME}"/../.git ]; then
+    stub_git_dir="${BATS_TEST_DIRNAME}"/../.git
+    mkdir "$stub_git_dir"
+  fi
+
   stub_repeated brew false
   stub_ruby_build 'echo ERROR >&2 && exit 2' \
     "--definitions : echo 1.8.7 1.9.3-p0 1.9.3-p194 2.1.2 | tr ' ' $'\\n'"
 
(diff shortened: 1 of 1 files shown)


mislav (commented):
Hi, thank you for offering to make the test more stable. I'm unclear about the original failure on systems without `brew`, though. The test scenario in question has this line:

```
stub_repeated brew false
```

This creates a test-specific stub for the `brew` binary for the duration of the test. This stub is dynamically injected into PATH in test mode. So, for the purpose of the test it should not matter whether `brew` exists on the user's system or not. Our `brew` stub should take precedence.

So if this test fails on systems without `brew` for anymore, we shouldn't aim to fix a single test scenario and its output, but to fix the `stub_repeated brew` system that's used in tests. Thoughts?

piperswe:
Turns out I had forgotten the context - the patch is a few years old at this point. It's not actually anything to do with Homebrew, and I can remove that part. The Debian build environment doesn't have the `.git`directory. I'll update this PR to only check for the existence of `.git` - I think it's reasonable to want this test to pass when you download the tarball from GitHub instead of cloning.

mislav:
Thanks for this and sorry for the late reaction. I've amended the approach in a way that I find easier to read/maintain: temporarily create the `.git` directory if missing.

PR #2443, opened 2024-09-09
Fix `ruby-build` upgrade instructions

iheanyi:
I get bit by this every time I need to upgrade Ruby. `brew upgrade` actually upgrades all installed packages, whereas `brew install ruby-build` will update a singular dependency. This PR updates those instructions so we don't get bit by this.

See https://github.com/orgs/Homebrew/discussions/3126 for more information.

### DIFF
--- a/test/rbenv.bats
+++ b/test/rbenv.bats
@@ -126,7 +126,7 @@ See all available versions with \`rbenv install --list-all'.
 
 If the version you need is missing, try upgrading ruby-build:
 
-  brew upgrade ruby-build
+  brew install ruby-build
 OUT
 
   unstub brew


monfresh:
That is not entirely correct. The comments in that Homebrew discussion are misunderstanding how Homebrew works. If you run `brew upgrade ruby-build`, it will not upgrade every single package you have installed. Instead, the default behavior is to also upgrade any outdated dependencies, and outdated dependencies of those dependencies. So, in some cases, it will feel like it's updating everything.

This is similar behavior to updating gems in a Ruby project. If you run `bundle update rails`, it will also update any outdated dependencies of Rails that satisfy the versioning rules. This is considered a good thing generally.

If you don't want this behavior, you can set the env var `HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK`, as explained in the documentation when you run `brew upgrade --help`:

```
Unless HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK is set, brew upgrade or brew
reinstall will be run for outdated dependents and dependents with broken
linkage, respectively.
```

iheanyi:
@monfresh Ohhhhh, TIL something new! Thanks for pointing me to that setting, I appreciate it. 🙏🏾

mislav:
Thanks @monfresh and @jasonkarns. I would just like to point out that `ruby-build` Homebrew formula has quite some dependent formulae:
```rb
  depends_on "autoconf"
  depends_on "libyaml"
  depends_on "openssl@3"
  depends_on "pkg-config"
  depends_on "readline"
```
Unfortunately, since many other formulae also depend on these (especially on `openssl`), an innocent attempt at upgrading ruby-build might set of a whole cascade of upgrades that can take more than half an hour.

In our upgrade instructions, we _could_ potentially advise setting the environment variable:
```sh
HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=yes brew upgrade ruby-build
```
but that's a mouthful and I fear that it obscures the actual upgrade command, which is by itself simple and memorable. So for now, I would also vote to not change anything, unless this becomes a big pain point for lots of ruby-build users.

PR #2447, opened 2024-09-23
Add -d, --dir option to install to <prefix>/<definition> instead of <prefix>

eregon:
* Similar to ruby-install's -r, --rubies-dir flag.

So this enables for example:
```
$ ruby-build -d truffleruby+graalvm-24.1.0 ~/.rubies
...
==> Installed truffleruby+graalvm-24.1.0 to /home/eregon/.rubies/truffleruby+graalvm-24.1.0
```

Which I find quite convenient compared to having to copy-paste the definition names after the rubies directory.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #
-# Usage: ruby-build [-kpv] <definition> <prefix> [-- <configure-args...>]
+# Usage: ruby-build [-dvpk] <definition> <prefix> [-- <configure-args...>]
 #        ruby-build {--list|--definitions}
 #        ruby-build --version
 #
@@ -8,6 +8,7 @@
 #   --definitions   List all local definitions, including outdated ones
 #   --version       Show version of ruby-build
 #
+#   -d, --dir       Install the Ruby in <prefix>/<definition> instead of <prefix>
 #   -v, --verbose   Verbose mode: forward all build output to stdout/stderr
 #   -p, --patch     Apply a patch from stdin before building
 #   -k, --keep      Do not remove source tree after installation
@@ -1358,6 +1359,7 @@ unset HAS_PATCH
 unset IPV4
 unset IPV6
 unset EARLY_EXIT
+unset APPEND_DEFINITION_TO_PREFIX
 
 RUBY_BUILD_INSTALL_PREFIX="$(abs_dirname "$0")/.."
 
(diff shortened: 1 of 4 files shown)


bin/ruby-build
```diff
@@ -8,6 +8,7 @@
 #   --definitions   List all local definitions, including outdated ones
 #   --version       Show version of ruby-build
 #
+#   -d, --dir       Install the Ruby in <prefix>/<definition> instead of <prefix>
```
mislav, line 11: My nitpick is that maybe this flag should receive a value instead of being a boolean flag.

When I see an invocation like this:
```
ruby-build 3.2.1 --dir ~/.rubies
```
it would seem to me like the `--dir` flag is passed a value of `~/.rubies`, but that's not how this is implemented right now. This would likely work: `ruby-build --dir 3.2.1 ~/.rubies`, and so would this: `ruby-build 3.2.1 ~/.rubies --dir`, but I do not think these invocations should work because they look "wrong" and I believe that they should error out. WDYT?

mislav (commented):
Thanks for this!!

mislav:
BTW, what do you think about this potential feature being lumped in with all the others, while we're going in this direction:
```sh
ruby-build truffleruby+graalvm -d ~/.rubies
#=> (installs the latest "truffleruby+graalvm" version)
```

bin/ruby-build
eregon: I tried to fit the existing usage `ruby-build [-kpv] <definition> <prefix>`.
If the prefix becomes optional it feels like opening Pandora's box, so that's why I preferred an option without a value. Also it seems currently ruby-build doesn't have any option taking an argument, and I would like to avoid having to add that.

I think all 3 variants you showed are fine to use.
I used `--dir` in the example because it works and looks cool, but `ruby-build --dir 3.2.1 ~/.rubies` is just as fine.

I'm happy to update the PR's description if that helps.

eregon:
> BTW, what do you think about this potential feature being lumped in with all the others, while we're going in this direction:

It seems nice to have but that feels out of scope and I don't really have time to look into that. I don't really need it either.
There is also the question for CRuby if e.g. `3.3` should work, etc.

bin/ruby-build
eregon: Actually I should update the usage line to `ruby-build [-kpvd]` and then the documented way to use this is `ruby-build -d definition prefix`.

eregon: I updated the usage line and the PR descriptions to use `ruby-build -d`.
The fact `-d` in the middle works is then like an easter egg.

mislav (approved):
Thank you!

I've added a man page entry and a test.

eregon:
Thank you!

PR #2451, opened 2024-10-07
Add Ruby 3.4.0-preview2

a-lavis:


eregon:
The checksum doesn't match it seems: https://github.com/rbenv/ruby-build/actions/runs/11220084120/job/31190198169?pr=2451

eregon:
I fixed it and it's now matching what's at https://www.ruby-lang.org/en/news/2024/10/07/ruby-3-4-0-preview2-released/

PR #2462, opened 2024-11-05
Allow named pipes as build definition argument

mislav:
Allow ruby-build to read a build definition from a named pipe in addition to just from regular files.

Example:
```
ruby-build <(echo "...") /path/to/ruby
```


jasonkarns (approved):
nice!

PR #2465, opened 2024-11-06
Suggest to install rubies in a user-writable directory in the README

eregon:
Otherwise `gem install` and `bundle install` (without `--path`) won't work without `sudo`. In fact the previous example install command wouldn't work at all without `sudo`.


mislav (approved):
Good thinking!

PR #2466, opened 2024-11-09
Add an environment variable to always build openssl/libssl

eregon:
* This is useful when one wants to avoid depending e.g. on which openssl version is installed in Homebrew, notably for https://github.com/ruby/setup-ruby/issues/668

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1122,6 +1122,8 @@ normalize_semver() {
 # If a compatible Homebrew-installed OpenSSL version is found during
 # checking, Ruby will be linked to it and the check will return false.
 needs_openssl() {
+  [ -z "$RUBY_BUILD_VENDOR_OPENSSL" ] || return 0
+
   [[ "$RUBY_CONFIGURE_OPTS ${RUBY_CONFIGURE_OPTS_ARRAY[*]}" != *--with-openssl-dir=* ]] || return 1
 
   local system_version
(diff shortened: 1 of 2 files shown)


bin/ruby-build
```diff
@@ -1122,6 +1122,8 @@ normalize_semver() {
 # If a compatible Homebrew-installed OpenSSL version is found during
 # checking, Ruby will be linked to it and the check will return false.
 needs_openssl() {
+  [ "$RUBY_BUILD_BUILD_OPENSSL" = "true" ] && return 0
```
mislav, line 1125: 1. Would a better name be "RUBY_BUILD_VENDOR_OPENSSL" so that it's clear that openssl will be vendored and to avoid saying "build" twice?
2. We generally allow any non-blank string to be a truthy value for an environment variable, so I would rewrite this check as:

```suggestion
  [ -z "$RUBY_BUILD_VENDOR_OPENSSL" ] || return 0
```

mislav (commented):
Thank you for suggesting! Would you add a note about the new option to the README too?

mislav:
For posterity: this adds an opt-out mechanism first requested in https://github.com/rbenv/ruby-build/pull/2275#issuecomment-1836367229

bin/ruby-build
eregon: Done, thanks

eregon:
I docuemnted it in the README, it should be good to go

mislav (approved):

PR #2493, opened 2025-01-10
Fix library lookup path in the vendored OpenSSL

### LINKED ISSUE #2488: The openssl built by ruby-build has somewhat broken library dependencies
### Steps to reproduce the behavior

On Linux:
```
$ RUBY_BUILD_VENDOR_OPENSSL=true ruby-build --verbose ruby-3.4.1 ~/.rubies/ruby-3.4.1-openssl-3.4.0
$ cd ~/.rubies/ruby-3.4.1-openssl-3.4.0/openssl
$ bin/openssl version          
bin/openssl: /lib64/libssl.so.3: version `OPENSSL_3.4.0' not found (required by bin/openssl)
bin/openssl: /lib64/libssl.so.3: version `OPENSSL_3.2.0' not found (required by bin/openssl)
bin/openssl: /lib64/libcrypto.so.3: version `OPENSSL_3.3.0' not found (required by bin/openssl)
bin/openssl: /lib64/libcrypto.so.3: version `OPENSSL_3.4.0' not found (required by bin/openssl)
bin/openssl: /lib64/libcrypto.so.3: version `OPENSSL_3.2.0' not found (required by bin/openssl)
$ ldd bin/openssl     
bin/openssl: /lib64/libssl.so.3: version `OPENSSL_3.4.0' not found (required by bin/openssl)
bin/openssl: /lib64/libssl.so.3: version `OPENSSL_3.2.0' not found (required by bin/openssl)
bin/openssl: /lib64/libcrypto.so.3: version `OPENSSL_3.3.0' not found (required by bin/openssl)
bin/openssl: /lib64/libcrypto.so.3: version `OPENSSL_3.4.0' not found (required by bin/openssl)
bin/openssl: /lib64/libcrypto.so.3: version `OPENSSL_3.2.0' not found (required by bin/openssl)
	linux-vdso.so.1 (0x00007ffdc43e5000)
	libssl.so.3 => /lib64/libssl.so.3 (0x00007fa625c4b000) <= BUG
	libcrypto.so.3 => /lib64/libcrypto.so.3 (0x00007fa625600000) <= BUG
	libc.so.6 => /lib64/libc.so.6 (0x00007fa62541e000)
	libz.so.1 => /lib64/libz.so.1 (0x00007fa625c31000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fa625d0a000)
$ ldd lib/libssl.so   
lib/libssl.so: /lib64/libcrypto.so.3: version `OPENSSL_3.3.0' not found (required by lib/libssl.so)
lib/libssl.so: /lib64/libcrypto.so.3: version `OPENSSL_3.4.0' not found (required by lib/libssl.so)
lib/libssl.so: /lib64/libcrypto.so.3: version `OPENSSL_3.2.0' not found (required by lib/libssl.so)
	linux-vdso.so.1 (0x00007fffc97f4000)
	libcrypto.so.3 => /lib64/libcrypto.so.3 (0x00007feaaca00000) <= BUG
	libc.so.6 => /lib64/libc.so.6 (0x00007feaac81e000)
	libz.so.1 => /lib64/libz.so.1 (0x00007feaad07b000)
	/lib64/ld-linux-x86-64.so.2 (0x00007feaad1b5000)
```

I added `<= BUG` to the output to highlight the problematic lines.
So we see `bin/openssl` links to system libssl and system libcrypto, when it should link those inside the openssl prefix.
And also that the built libssl links to system libcrypto and not in-the-same-dir libcrypto.
Since the system openssl in this case is 3.1.4 it results in relatively clear errors, but if they were closer it could maybe crash or so.

Luckily this doesn't seem to affect the Ruby openssl extension, which links correctly to the correct libssl and libcrypto:
```
$ chruby ruby-3.4.1-openssl-3.4.0                          
$ ldd $(ruby -ropenssl -e 'puts $".grep(/openssl\.(so|bundle)/)')
	linux-vdso.so.1 (0x00007ffc87fd8000)
	libruby.so.3.4 => /home/eregon/.rubies/ruby-3.4.1-openssl-3.4.0/lib/libruby.so.3.4 (0x00007fa5b4400000)
	libssl.so.3 => /home/eregon/.rubies/ruby-3.4.1-openssl-3.4.0/openssl/lib/libssl.so.3 (0x00007fa5b42fc000)
	libcrypto.so.3 => /home/eregon/.rubies/ruby-3.4.1-openssl-3.4.0/openssl/lib/libcrypto.so.3 (0x00007fa5b3c00000)
	libm.so.6 => /lib64/libm.so.6 (0x00007fa5b421b000)
	libc.so.6 => /lib64/libc.so.6 (0x00007fa5b3a1e000)
	libz.so.1 => /lib64/libz.so.1 (0x00007fa5b4b13000)
	libcrypt.so.2 => /lib64/libcrypt.so.2 (0x00007fa5b4ada000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007fa5b4ab5000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fa5b4ba8000)
```
But that still seems brittle because I'm not sure it's guaranteed the libcrypto dependency there always wins over the dependency declared in libssl. Having them with different values certainly seems bad.
I also checked with `ruby -ropen-uri -e 'puts URI.send(:open, %{https://rubygems.org/}) { |f| f.read(1024) }; sleep'` and `lsof` and that looks fine, only the correct ibssl/libcrypto are loaded. But maybe by "chance".

I think it would be nice to fix this so the built openssl has correct library dependencies and we don't run the risk of these conflicting dependencies to stop working at some point.

I noticed this with OpenSSL 3.4 but I think it happens with older versions as well like 3.0.
FWIW this is the patch I used because I wanted to check against OpenSSL 3.4 specifically (for https://github.com/oracle/truffleruby/issues/3724):
```patch
diff --git a/share/ruby-build/3.4.1 b/share/ruby-build/3.4.1
index adb10b78..d692532c 100644
--- a/share/ruby-build/3.4.1
+++ b/share/ruby-build/3.4.1
@@ -1,2 +1,2 @@
-install_package "openssl-3.0.15" "https://github.com/openssl/openssl/releases/download/openssl-3.0.15/openssl-3.0.15.tar.gz#23c666d0edf20f14249b3d8f0368acaee9ab585b09e1de82107c66e1f3ec9533" openssl --if needs_openssl:1.0.2-3.x.x
+install_package "openssl-3.4.0" "https://github.com/openssl/openssl/releases/download/openssl-3.4.0/openssl-3.4.0.tar.gz" openssl --if needs_openssl:1.0.2-3.x.x
 install_package "ruby-3.4.1" "https://cache.ruby-lang.org/pub/ruby/3.4/ruby-3.4.1.tar.gz#3d385e5d22d368b064c817a13ed8e3cc3f71a7705d7ed1bae78013c33aa7c87f" enable_shared standard
```

mislav:
When a vendored OpenSSL is needed for compiling Ruby, that OpenSSL installation ends up with its `bin/openssl` executable broken due to not finding "libssl.so" and "libcrypto.so" in the global load path for libraries. This doesn't seem to negatively affect the Ruby "openssl" extension at runtime, but does constitute a broken OpenSSL installation nevertheless.

This change causes the `bin/openssl` executable and related shared libraries to be built with an "RPATH" pointing to the "lib" directory of the vendored OpenSSL. See https://wiki.openssl.org/index.php/Compilation_and_Installation#Using_RPATHs

Fixes #2488

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1215,7 +1215,7 @@ build_package_openssl() {
   [[ "$1" != openssl-1.0.* ]] || nokerberos=1
 
   # Compile a shared lib with zlib dynamically linked.
-  package_option openssl configure --openssldir="$OPENSSLDIR" --libdir="lib" zlib-dynamic no-ssl3 shared ${nokerberos:+no-ssl2 no-krb5}
+  package_option openssl configure --openssldir="$OPENSSLDIR" --libdir="lib" -Wl,-rpath="$OPENSSL_PREFIX_PATH/lib" zlib-dynamic no-ssl3 shared ${nokerberos:+no-ssl2 no-krb5}
 
   # Skip building OpenSSL docs, which is slow.
   local make_target="install_sw install_ssldirs"
--- a/test/build.bats
+++ b/test/build.bats
@@ -397,7 +397,7 @@ DEF
   unstub make
 
   assert_build_log <<OUT
-openssl-1.1.1w: [--prefix=${INSTALL_ROOT}/openssl,--openssldir=${INSTALL_ROOT}/openssl/ssl,--libdir=lib,zlib-dynamic,no-ssl3,shared]
+openssl-1.1.1w: [--prefix=${INSTALL_ROOT}/openssl,--openssldir=${INSTALL_ROOT}/openssl/ssl,--libdir=lib,-Wl,-rpath=${INSTALL_ROOT}/openssl/lib,zlib-dynamic,no-ssl3,shared]
 make -j 2
 make install_sw install_ssldirs
 ruby-3.2.0: [--prefix=$INSTALL_ROOT,--with-openssl-dir=$INSTALL_ROOT/openssl,--with-ext=openssl,psych,+] PKG_CONFIG_PATH=${TMP}/install/openssl/lib/pkgconfig


eregon:
I checked locally and it works, thank you for the quick fix!

bin/ruby-build
```diff
@@ -1215,7 +1215,7 @@ build_package_openssl() {
   [[ "$1" != openssl-1.0.* ]] || nokerberos=1
 
   # Compile a shared lib with zlib dynamically linked.
-  package_option openssl configure --openssldir="$OPENSSLDIR" --libdir="lib" zlib-dynamic no-ssl3 shared ${nokerberos:+no-ssl2 no-krb5}
+  package_option openssl configure --openssldir="$OPENSSLDIR" --libdir="lib" -Wl,-rpath="$OPENSSL_PREFIX_PATH/lib" zlib-dynamic no-ssl3 shared ${nokerberos:+no-ssl2 no-krb5}
```
jeremy, line 1218:
```bash
clang -fPIC -arch arm64 -O3 -Wall -L. -dynamiclib -current_version 1.1 -compatibility_version 1.1 -Wl,-search_paths_first -L/opt/homebrew/opt/ruby/lib -Wl,-rpath=/Users/dorin/.local/share/mise/installs/ruby/2.7.8/openssl/lib -install_name /Users/dorin/.local/share/mise/installs/ruby/2.7.8/openssl/lib/libcrypto.1.1.dylib -o …
                  
ld: unknown options: -rpath=/Users/dorin/.local/share/mise/installs/ruby/2.7.8/openssl/lib 
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Looks like this needs to use `-Wl,-rpath,"$OPENSSL_PREFIX_PATH/lib"` for Clang (comma rather than equals).

jeremy: RPATH is automatically set up on BSDs, so this is only needed on Linux. Also, comma syntax appears to be preferred.

```suggestion
  # Add RPATH for Linux/GCC
  case "$(uname -s)" in
    Linux)
      if cc -v 2>&1 | grep -q "gcc"; then
        EXTRA_RPATH_FLAGS="-Wl,-rpath,${OPENSSL_PREFIX_PATH}/lib -Wl,--enable-new-dtags"
      fi
      ;;
  esac

  package_option openssl configure --openssldir="$OPENSSLDIR" --libdir="lib" zlib-dynamic no-ssl3 shared ${nokerberos:+no-ssl2 no-krb5}
```

eregon: Could you make a PR?
BTW `clang` on Linux would also need the RPATH, so we shouldn't do this only with gcc.
Re `--enable-new-dtags` I don't know if necessary, I got a RUNPATH anyway without it.

jeremy: Can't shepherd through a PR or test on all these platforms at the moment. Recommend reverting this PR for now so macOS users can install Ruby.

Note that `--enable-new-dtags` is set by default on some distros but not others. If you have a RUNPATH, it's because it's on by default for you (Gentoo, Debian, … others?). Behavior is subtly different with RUNPATH taking precedence and respecting LD_LIBRARY_PATH thereafter.

mislav: Thanks @jeremy. I've undone `rpath` argument passing on non-Linux platforms, and added clang detection.

I've also expedited a release thinking that the previous change already made it into a release, but it turns out I've just released both changes together, so fingers crossed that this is the right incantation. https://github.com/rbenv/ruby-build/releases/tag/v20250114

If more regressions are discovered, I think I would also revert the whole rpath thing since it's more trouble than it's worth. I must admit I don't know much about RPATH vs RUNPATH, but openssl should not need this many extra compiler flags just to produce a working build in a custom location.

jeremy: Ah! Looks like mise installs ruby-build from git which gave us an early preview of the forthcoming release. Nice to catch this early, then.

eregon:
I filed  two openssl issues upstreams for the gotchas mentioned here: https://github.com/openssl/openssl/issues/26422 https://github.com/openssl/openssl/issues/26423

PR #2496, opened 2025-01-14
Fix installing OpenSSL on non-Linux or with clang

mislav:
The recent rpath argument addition broke compilation of OpenSSL with clang since it seems expect this syntax (comma instead of "="):

    -Wl,-rpath,<path>

Also, it doesn't seem that the rpath argument is necessary for any platform other than "Linux"; that is, we should skip passing it on BSD and macOS.

Thanks @jeremy https://github.com/rbenv/ruby-build/pull/2493#pullrequestreview-2550800730

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1211,11 +1211,21 @@ build_package_openssl() {
   # Tell Ruby to use this openssl for its extension.
   package_option ruby configure --with-openssl-dir="$OPENSSL_PREFIX_PATH"
 
-  local nokerberos
-  [[ "$1" != openssl-1.0.* ]] || nokerberos=1
-
   # Compile a shared lib with zlib dynamically linked.
-  package_option openssl configure --openssldir="$OPENSSLDIR" --libdir="lib" -Wl,-rpath="$OPENSSL_PREFIX_PATH/lib" zlib-dynamic no-ssl3 shared ${nokerberos:+no-ssl2 no-krb5}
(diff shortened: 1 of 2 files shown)


bin/ruby-build
```diff
@@ -1211,11 +1211,21 @@ build_package_openssl() {
...
-  package_option openssl configure --openssldir="$OPENSSLDIR" --libdir="lib" -Wl,-rpath="$OPENSSL_PREFIX_PATH/lib" zlib-dynamic no-ssl3 shared ${nokerberos:+no-ssl2 no-krb5}
+  package_option openssl configure --openssldir="$OPENSSLDIR" --libdir="lib" zlib-dynamic no-ssl3 shared
+
+  # Help OpenSSL find its own shared libraries on Linux.
+  if [ "$(uname -s)" = "Linux" ]; then
+    # shellcheck disable=SC2031
+    if [ "$CC" = "clang" ] || "${CC:-cc}" --version 2>&1 | grep -wq clang; then
+      package_option openssl configure -Wl,-rpath,"$OPENSSL_PREFIX_PATH/lib"
+    else
+      package_option openssl configure -Wl,-rpath="$OPENSSL_PREFIX_PATH/lib"
```
jeremy, line 1223: Can use the comma syntax `-Wl,-rpath,"$OPENSSL_PREFIX_PATH/lib"` for GCC as well (the commas are split into separate args to pass to the linker, giving `ld -rpath …` instead of `ld -rpath=…`) so we can drop the clang check.

eregon: https://github.com/rbenv/ruby-build/pull/2497

PR #2498, opened 2025-01-15
Improve test output for failed stubs

mislav:
The `stub`/`unstub` test helpers are unfortunately opaque as both are undocumented, and when stubbing fails for some reason, determining why is left as a guessing/debugging exercise. This was a time-sink for myself on many an occasion.

This improves the `unstub <PROGRAM>` helper output with the following features:

- If the number of PROGRAM invocations was fewer than the number of stubbed (i.e. expected) arguments, it now prints `expected X invocations, got Y` to stderr;

- If PROGRAM got invoked but `unstub PROGRAM` still failed, it was likely that some of the invocations didn't match the stubbed ones. However, the human reading the test output has no insight into which invocations actually happened. The unstub helper now prints:

      re-run test with <PROGRAM>_STUB_DEBUG=3 to log `<PROGRAM>' invocations

  This "debug" mode already existed, but is off by default, and could previously only be discovered by someone who read the `stub` implementation code. After following the new suggestion, actual invocations of PROGRAM and all its arguments will be printed to stderr during the test run, hopefully making it clearer for the human debugging the test.

### DIFF
--- a/test/stubs/stub
+++ b/test/stubs/stub
@@ -14,8 +14,8 @@ _STUB_RESULT="${PROGRAM}_STUB_RESULT"
 _STUB_END="${PROGRAM}_STUB_END"
 _STUB_DEBUG="${PROGRAM}_STUB_DEBUG"
 
-if [ -n "${!_STUB_DEBUG}" ]; then
-  echo "$program" "$@" >&${!_STUB_DEBUG}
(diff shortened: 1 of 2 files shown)


eregon (approved):
Nice!

PR #2517, opened 2025-04-02
Don't delete jruby.sh

headius:
I'm not sure why this file was being deleted but this code dates way back! Unfortunately deleting this interferes with a new version of the native JRuby launcher we want to release which is dependent on the shell script being present. This PR removes jruby.sh from the `remove_windows_files` function so it can be used by the new launcher.

See https://github.com/jruby/jruby-launcher/pull/48#issuecomment-2772272290

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -935,7 +935,7 @@ build_package_artichoke() {
 remove_windows_files() {
   # shellcheck disable=SC2164
   cd "$PREFIX_PATH"
-  rm -f bin/*.exe bin/*.dll bin/*.bat bin/jruby.sh
+  rm -f bin/*.exe bin/*.dll bin/*.bat
 }
 
 clean_prefix_path_truffleruby() {


Earlopain:
I gave this a try with `jruby-launcher 2.0.0.pr1` (see linked PR) and jruby 9.4.12.0, it seems to work ok. `gem` / `ruby` / `bundle` etc. commands all do what they're supposed to.

headius:
This should also not interfere with the installation of the "classic" native launcher, since that overwrites the `jruby` command and is not affected by `jruby.sh`

mislav:
Thank you!

headius:
@mislav Can you explain how this fix gets out to users? What process do they need to follow in e.g. rbenv?

We are trying to decide if we have to hack the jruby-launcher gem to bring its own jruby.sh to work around the deletion on older ruby-build.

headius:
Thank you for the clarification and for the release!

> I think it's worthwhile to add the hack to jruby-launcher gem

We ended up at the same conclusion, and the just-released jruby-launcher 2.0 will copy the latest jruby.sh into bin if it is not already present. I believe it should install and work properly for all rbenv installs prior to the change in this PR.

PR #2520, opened 2025-04-14
Add JRuby 10.0.0.0

headius:


share/ruby-build/jruby-10.0.0.0
```diff
@@ -0,0 +1,2 @@
+require_java 8
```
expeehaa, line 1: Since the release notes for JRuby 10 say that the minimum Java version was bumped to 21, shouldn’t this line also say `require_java 21` then?

hsbt:
@headius I bump up required java version to 21 at https://github.com/rbenv/ruby-build/commit/4cdc5a720f5410b3128bf64458d42c65ad5568c0

eregon:
@headius FYI https://github.com/ruby/ruby-builder/actions/runs/14461082292/job/40553708470 failed. I'm on holidays so won't be able to fix that. You or someone else can make a PR to fix it and I should be able to merge it

headius:
@hsbt Thank you for the fix!

eregon:
There is https://github.com/ruby/ruby-builder/pull/22

PR #2522, opened 2025-04-18
Add 3.5.0-preview1

hsbt:


rafaelp:
Somehow the release after this merge removed ruby 3.4.3 (that as added in the previous release)
Maybe because it was merged into master instead of rbenv:master?
(I don't know this project flow, just guessing)

mislav:
@rafaelp Hi, `3.4.3` was added in ruby-build [v20250415](https://github.com/rbenv/ruby-build/releases/tag/v20250415) and was not removed since.

You could check `rbenv install --version` to see which ruby-build you have locally.

PR #2523, opened 2025-04-18
Honor `$JAVA_HOME` in `require_java`

ntkme:
JRuby does not require `$JAVA_HOME/bin` to be part of `$PATH`, therefore we should test java version with `$JAVA_HOME/bin/java -version` when `$JAVA_HOME` is defined.

See: https://github.com/ruby/ruby-builder/pull/22#issuecomment-2814646152

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -970,7 +970,11 @@ after_install_package() {
 require_java() {
   local required="$1"
   local java_version version_string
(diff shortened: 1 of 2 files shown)


ntkme:
Pushed a commit to unset `$JAVA_HOME` before running test so that the existing tests are not affected by `$JAVA_HOME` which is set by default in GitHub Actions.

PR #2530, opened 2025-05-07
Add JRuby 10.0.0.1 and 9.4.12.1

headius:
This adds JRuby 10.0.0.1 and 9.4.12.1, which include the jruby-openssl fix for https://github.com/jruby/jruby-openssl/security/advisories/GHSA-72qj-48g4-5xgx.

### DIFF
--- a/share/ruby-build/jruby-10.0.0.1
+++ b/share/ruby-build/jruby-10.0.0.1
@@ -1,0 +1,2 @@
+require_java 21
+install_package "jruby-10.0.0.1" "https://repo1.maven.org/maven2/org/jruby/jruby-dist/10.0.0.1/jruby-dist-10.0.0.1-bin.tar.gz#0ba34ac5dfec7c22659b14db668a06284db7fc1c820c49c04b92271a6636bafb" jruby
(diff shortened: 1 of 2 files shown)


share/ruby-build/jruby-10.0.0.1
```diff
@@ -0,0 +1,2 @@
+require_java 8
```
mislav, line 1: Shouldn't this require a newer Java version? https://github.com/rbenv/ruby-build/commit/4cdc5a720f5410b3128bf64458d42c65ad5568c0

mislav (changes requested):

mislav:
> I've based this on the 10.0.0.0 PR, which I saw today has not yet been merged.

Not sure what you mean, since the 10.0.0.0 PR was merged and released three weeks ago? https://github.com/rbenv/ruby-build/releases/tag/v20250415

headius:
Sorry, I got my repositories crossed.

Yes, should be 21 minimum, I'll fix that. Thanks!

share/ruby-build/jruby-10.0.0.1
headius: I have fixed that and re-pushed, but where does this come from? I just used the update script. Does it look at the current Java in PATH and somehow use that?

mislav (approved):

PR #2531, opened 2025-05-07
Modify `update-jruby` script to require Java 21+ for JRuby 10+

mislav:
Any jruby version starting with `10.*` will have `require_java 21` instead of `require_java 8`.

/cc @headius https://github.com/rbenv/ruby-build/pull/2530#discussion_r2078079546


hsbt (approved):
👍

Issue #2535, opened 2025-05-15
When will 3.4.4 will be available on brew/ruby-build?


kapso:

rud:
Nudge: would somebody be so kind as to cut a fresh release of this project? Thank you in advance :)

jasonkarns:
Released https://github.com/rbenv/ruby-build/releases/tag/v20250516

Not sure why 3.4.4 wasn't included in the release changelog though 😕 . It was added by 87f57bb7a8176afa7f9256a1bd6fa18d251f210e

mislav:
@jasonkarns Thanks for catching that. It looks like the definition for Ruby 3.4.4 was added using the [update-ruby workflow](https://github.com/rbenv/ruby-build/blob/master/.github/workflows/update-ruby.yml), which creates a commit directly on the main branch. When we tag ruby-build releases, the release notes are automatically generated using GitHub's [generated release notes feature](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes), which only lists merged PRs but does not list individual commits. /cc @hsbt

rud:
Beautiful, thank you

hsbt:
@jasonkarns Thanks! I got the short trip at last week.

I always update the changelog for Ruby releases manually.

jasonkarns:
@hsbt no worries, happy to help. As that was my first release flow, I have a couple followup PRs to make to the docs and whatnot.



@mislav 
> which creates a commit directly on the main branch

I see. I might PR a change such that the update-ruby workflow creates merge commits? Not sure if GH release notes specifically require PRs or if merge commits will do. Or perhaps will toy with the release.yml config file for managing those notes?

PR #2537, opened 2025-05-22
Document examples of using custom defs

jasonkarns:
Add examples of passing a custom build definition file in place of a 'known' ruby version. This is meant to more clearly document how users might use a custom build definition for a recently-released ruby that is not yet bundled in ruby-build.

This is primarily intended as a piece of documentation we can point to in Issues, wiki or to downstream tooling (like asdf-ruby) for how to use ruby-build to install additional builds of ruby. My _hope_ is that with a few examples, more folks self-service with _just-released_ rubies.

Additionally, document in the readme how to use `RUBY_BUILD_DEFINITIONS` path environment variable to provide a _collection_ of custom build definitions. This is meant to give guidance on how an organization might share multiple custom build definitions, or downstream users of ruby-build (like asdf-ruby) might publish unreleased rubies for installation.

nit/quibbling over wording: custom build defs _are_ recognized by ruby-build, just not bundled or available out of the box.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,30 @@ Depending on the context, ruby-build does a little bit more than the above: for 
 
 #### Custom Build Definitions
 
-To install a version of Ruby that is not recognized by ruby-build, you can specify the path to a custom build definition file in place of a Ruby version number.
+To install a version of Ruby that is not available in ruby-build, you can specify the path to a custom build definition file in place of a Ruby version number.
+
+```sh
+# As a standalone program
+$ ruby-build -d /path/to/3.4-custom /opt/rubies  # installs to /opt/rubies/3.4-custom
+
+# As an rbenv plugin
+$ rbenv install /path/to/3.4-custom              # installs to $(rbenv root)/versions/3.4-custom
+```
+
+You can also provide a _directory_ of custom build definition files.
+The path(s) will be searched along with ruby-build's bundled `share/ruby-build/` directory.
+(Perhaps a collection of 3rd-party build definitions published as a git repo,
+or an organization's custom build definitions distributed in-house.)
+
+```sh
+# As a standalone program
+$ RUBY_BUILD_DEFINITIONS=/path/to/custom/defs ruby-build --definitions              # lists all available versions of Ruby, including custom defs
+$ RUBY_BUILD_DEFINITIONS=/path/to/custom/defs ruby-build -d 3.5-custom /opt/rubies  # installs to /opt/rubies/3.5-custom
+
+# As an rbenv plugin
+$ RUBY_BUILD_DEFINITIONS=/path/to/custom/defs rbenv install --list                  # lists all available versions of Ruby, including custom defs
+$ RUBY_BUILD_DEFINITIONS=/path/to/custom/defs rbenv install 3.5-custom              # installs to $(rbenv root)/versions/3.5-custom
+```
 
 Check out [default build definitions][definitions] as examples on how to write definition files.
 


README.md
```diff
@@ -65,7 +65,30 @@ Depending on the context, ruby-build does a little bit more than the above: for
 
 #### Custom Build Definitions
 
-To install a version of Ruby that is not recognized by ruby-build, you can specify the path to a custom build definition file in place of a Ruby version number.
+To install a version of Ruby that is not available in ruby-build, you can specify the path to a custom build definition file in place of a Ruby version number.
+
+```sh
+# As a standalone program
+$ ruby-build -d path/to/3.4-custom /opt/rubies  # installs to /opt/rubies/3.4-custom
```
eregon, line 72:
```suggestion
$ ruby-build -d /path/to/3.4-custom /opt/rubies  # installs to /opt/rubies/3.4-custom
```

README.md
```diff
@@ -65,7 +65,30 @@ Depending on the context, ruby-build does a little bit more than the above: for
...
+
+# As an rbenv plugin
+$ rbenv install path/to/3.4-custom              # installs to $(rbenv root)/versions/3.4-custom
```
eregon, line 75:
```suggestion
$ rbenv install /path/to/3.4-custom              # installs to $(rbenv root)/versions/3.4-custom
```

README.md
```diff
@@ -65,7 +65,30 @@ Depending on the context, ruby-build does a little bit more than the above: for
...
+```
+
+You can also provide a _directory_ of custom build definition files.
+The path(s) will be searched along with ruby-build's bundled `share/ruby-build/` directory.
+(Perhaps a collection of 3rd-party build definitions published as a git repo,
+or an organization's custom build definitions distributed in-house.)
+
+```sh
+# As a standalone program
+$ RUBY_BUILD_DEFINITIONS=path/to/custom/defs ruby-build --definitions              # lists all available versions of Ruby, including custom defs
```
eregon, line 85:
```suggestion
$ RUBY_BUILD_DEFINITIONS=/path/to/custom/defs ruby-build --definitions              # lists all available versions of Ruby, including custom defs
```
And same below

eregon:
The docs look good to me.

The reasons given in the description confuse me though:
> My hope is that with a few examples, more folks self-service with just-released rubies.

I think this is unlikely to happen and not that useful, first because CRuby releases are added very quickly here and second because if people want to do it themselves it's better they fork this repo, add the definition and make a PR, than just creating a file on their disk and not share it with others (while being not much more effort).
In fact IIRC for CRuby releases it's even automated.

> downstream users of ruby-build (like asdf-ruby) might publish unreleased rubies for installation.

ruby-build already provides many `*-dev` definitions, so I don't really see a real-world use-case for this.
It would be better contributed to ruby-build directly.

Actually it sounds to me like whatever you are using is lagging a few ruby-build releases behind, is that the case?

eregon (approved):

README.md
eregon: `/` because the command should work regardless of the current working directory, and `/path/to/..` is I think also clearer it needs to be changed by the user.

jasonkarns:
The push for this PR was a result of https://github.com/rbenv/ruby-build/issues/2535. Ruby had been added right away, but we hadn't released a version of ruby-build.

When they referenced the asdf-ruby plugin https://github.com/asdf-vm/asdf-ruby/issues/454, I wanted to link them to documentation that showed how users don't have to wait until ruby-build is released to take advantage of build definition files.

eregon:
Thanks for context.
I think for asdf and other Ruby installers using ruby-build the best workaround in that case is to just use master, as shown in https://github.com/asdf-vm/asdf-ruby/issues/454#issuecomment-2885320540.
Because the definition was already there, so the easiest is to just pick it up using master vs a release.

But yeah it would be good to solve this CRuby auto update but no release issue.
Whenever I add a TruffleRuby version I do a release, so there is no such issue.
I think we could do a release on every commit of ruby-build if that's an easier way to automate things.

Alternatively requiring only a workflow dispatch for a release would make it more convenient, that's what I do in setup-ruby:
https://github.com/ruby/setup-ruby/blob/master/.github/workflows/release.yml

PR #2543, opened 2025-06-24
Allow installation of Ruby 2.7, 3.0, 3.1 on Fedora >= 42

### LINKED ISSUE #2542: Treat Fedora spins as Fedora
There are official spins of [Universal Blue](https://universal-blue.org/), Fedora Atomic Desktop based images, such as Aurora, Bazzite, BlueFin, and uCore.
Many people, myself [included](https://github.com/pboling/galtzo), have created  further spins on top of those.

All of these are _fedora_, which is reported by `/etc/os-release`'s `ID_LIKE`, while the `NAME` reports the specific release, which should rarely need to be considered by a project like ruby-build.

Some relevant lines from my `/etc/os-release`
```
NAME="Aurora"
ID=aurora
ID_LIKE="fedora"
VERSION_ID=42
PLATFORM_ID="platform:f42"
```

I think we can use this to better target fixes for Fedora-based systems.

<details>
<summary>cat /etc/os-release</summary>

```
NAME="Aurora"
VERSION="latest-42.20250623.1 (Kinoite)"
RELEASE_TYPE=stable
ID=aurora
ID_LIKE="fedora"
VERSION_ID=42
VERSION_CODENAME="Stargazer"
PLATFORM_ID="platform:f42"
PRETTY_NAME="Aurora (Version: latest-42.20250623.1 / FROM Fedora Kinoite 42)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:universal-blue:aurora:42"
DEFAULT_HOSTNAME="aurora"
HOME_URL="https://getaurora.dev/"
DOCUMENTATION_URL="https://docs.getaurora.dev"
SUPPORT_URL="https://github.com/ublue-os/aurora/issues/"
BUG_REPORT_URL="https://github.com/ublue-os/aurora/issues/"
SUPPORT_END=2026-05-13
VARIANT="Kinoite"
VARIANT_ID=aurora-dx-hwe
OSTREE_VERSION='latest-42.20250623.1'
BUILD_ID="97252ff"
IMAGE_ID="aurora-dx-hwe"
```

</details>

pboling:
Fixes https://github.com/rbenv/ruby-build/issues/2542

~~Sadly this only allows installation of Ruby 3.1, as it fails in a different way on Ruby 3.0.~~ UPDATE: This works as-is for Ruby 3.0 once a C++ compiler is installed (aside: why does Ruby 3.0 need C++?).

```
sudo rpm-ostree install gcc-c++
systemctl reboot
```

This is an alternative resolution for https://github.com/rbenv/ruby-build/discussions/2529, and partially obviates the need for the [entry in the wiki](https://github.com/rbenv/ruby-build/wiki#fedora-42-and-ruby-31) (at least it bumps back the need for an external solution to Ruby < 2.7).

Also worth noting that the workaround in the wiki is not possible on Fedora Atomic Desktop systems, e.g. Universal Blue-based Aurora, Bazzite, BlueFin, or uCore, because they disable the `dnf` tool for installing software (due to being "Atomic").

```
> ruby-build 3.1.7 ~/.rubies/ruby-3.1.7

==> Downloading ruby-3.1.7.tar.gz...by-3.1.7
-> curl -q -fL -o ruby-3.1.7.tar.gz https://cache.ruby-lang.org/pub/ruby/3.1/ruby-3.1.7.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 19.8M  100 19.8M    0     0  10.1M      0  0:00:01  0:00:01 --:--:-- 10.1M
==> Installing ruby-3.1.7...

WARNING: ruby-3.1.7 is past its end of life and is now unsupported.
It no longer receives bug fixes or critical security updates.

-> ./configure "--prefix=$HOME/.rubies/ruby-3.1.7" --enable-shared --with-readline-dir=/home/linuxbrew/.linuxbrew/opt/readline --with-ext=openssl,psych,+
-> make -j 22
-> make install
==> Installed ruby-3.1.7 to /home/pboling/.rubies/ruby-3.1.7
```

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -207,7 +207,7 @@ is_freebsd() {
 }
 
 is_fedora() {
-  if [ -r /etc/os-release ]; then
+  [ -r /etc/os-release ] || return 1
       # shellcheck disable=SC1091
       source /etc/os-release
       # This will treat Aurora, Bazzite, BlueFin, uCore, and other Fedora variants as `fedora`.
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -211,13 +211,9 @@ is_fedora() {
       # shellcheck disable=SC1091
       source /etc/os-release
       # This will treat Aurora, Bazzite, BlueFin, uCore, and other Fedora variants as `fedora`.
-      # shellcheck disable=SC2153
-      local os_id
-      od_id="${ID_LIKE:-$ID}"
-      local os_version
-      os_version=$VERSION_ID
-      [ "${od_id}" = "fedora" ] || return 1
-      [ $# -eq 0 ] || [ "${os_version}" -ge "$1" ]
+      if [[ "${ID_LIKE:-$ID}" != "fedora" || ( $# -gt 0 && "$VERSION_ID" -lt "$1") ]]; then
+        return 1
+      fi
   fi
 }
 
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -728,10 +728,12 @@ build_package_standard() {
     if [ -z "$CC" ] && is_mac 1010; then
       export CC=clang
     elif [ "$ruby_semver" -lt 300200 ]; then
-      # Ruby older than 3.2 can't compile on GCC v15 unless options are provided to
-      #   use older GNU standard that is effectively "like GCC v14" => -std=gnu17
+      # GCC v15 changed the default mode from gnu17 to gnu23.
+      # See: https://gcc.gnu.org/gcc-15/changes.html#c
+      # Ruby older than 3.2 can't compile on GCC v15 in the default mode of gnu23.
+      # See: https://github.com/rbenv/ruby-build/discussions/2529
       if is_fedora && is_fedora 42; then
-        # Fedora 42+ has GCC v15 as default
+        # Fedora 42+ uses GCC v15 with mode gnu23 as default, so we configure for mode gnu17
         export CFLAGS="$CFLAGS -std=gnu17"
       fi
     fi
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -728,7 +728,7 @@ build_package_standard() {
       # See: https://gcc.gnu.org/gcc-15/changes.html#c
       # Ruby older than 3.2 can't compile on GCC v15 in the default mode of gnu23.
       # See: https://github.com/rbenv/ruby-build/discussions/2529
-      if is_fedora && is_fedora 42; then
+      if is_fedora 42; then
         # Fedora 42+ uses GCC v15 with mode gnu23 as default, so we configure for mode gnu17
         export CFLAGS="$CFLAGS -std=gnu17"
       fi
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -730,7 +730,7 @@ build_package_standard() {
       # See: https://github.com/rbenv/ruby-build/discussions/2529
       if is_fedora 42; then
         # Fedora 42+ uses GCC v15 with mode gnu23 as default, so we configure for mode gnu17
-        export CFLAGS="$CFLAGS -std=gnu17"
+        export CFLAGS="-std=gnu17 $CFLAGS "
       fi
     fi
     # ./configure --prefix=/path/to/ruby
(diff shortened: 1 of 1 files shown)


bin/ruby-build
```diff
@@ -206,6 +206,21 @@ is_freebsd() {
...
+      # shellcheck disable=SC1091
+      source /etc/os-release
+      # This will treat Aurora, Bazzite, BlueFin, uCore, and other Fedora variants as `fedora`.
+      # shellcheck disable=SC2153
+      local os_id
+      od_id="${ID_LIKE:-$ID}"
+      local os_version
+      os_version=$VERSION_ID
+      [ "${od_id}" = "fedora" ] || return 1
+      [ $# -eq 0 ] || [ "${os_version}" -ge "$1" ]
```
mislav, line 220: Nitpick: this could be simplified, something like
```suggestion
      if [[ "${ID_LIKE:-$ID}" != "fedora" || ( $# -gt 0 && "$VERSION_ID" -lt "$1") ]]; then
        return 1
      fi
```

bin/ruby-build
```diff
@@ -206,6 +206,21 @@ is_freebsd() {
   [ "$(uname -s)" = "FreeBSD" ]
 }
 
+is_fedora() {
+  if [ -r /etc/os-release ]; then
```
mislav, line 210: If `/etc/os-release` doesn't exist or is not readable, `is_fedora` would exit with 0 (success status). Let's change to an early return if the file is not readable:
```suggestion
  [ -r /etc/os-release ] || return 1
```

bin/ruby-build
```diff
@@ -712,6 +727,15 @@ build_package_standard() {
     fi
     if [ -z "$CC" ] && is_mac 1010; then
       export CC=clang
+    elif [ "$ruby_semver" -lt 300200 ]; then
+      # GCC v15 changed the default mode from gnu17 to gnu23.
+      # See: https://gcc.gnu.org/gcc-15/changes.html#c
+      # Ruby older than 3.2 can't compile on GCC v15 in the default mode of gnu23.
+      # See: https://github.com/rbenv/ruby-build/discussions/2529
+      if is_fedora && is_fedora 42; then
```
mislav, line 735: Would this condition work/be simpler?
```suggestion
      if is_fedora 42; then
```

bin/ruby-build
```diff
@@ -712,6 +727,15 @@ build_package_standard() {
...
+        # Fedora 42+ uses GCC v15 with mode gnu23 as default, so we configure for mode gnu17
+        export CFLAGS="$CFLAGS -std=gnu17"
```
mislav, line 737: Would we consider reversing the order of arguments so that user-supplied RUBY_CFLAGS could potentially override the `-std` flag, if necessary?
```suggestion
        export CFLAGS="-std=gnu17 $CFLAGS "
```

mislav (changes requested):
This is great; thank you! Some minor suggestions, only one important

mislav:
After this is merged, does it fully alleviate the need to install gcc14 as currently outlined in workarounds in the wiki?

pboling:
Yes it does, for Ruby 3.1.  And for the older ones you just need to make sure you have a c++ compiler installed, which was probably already a pre-requisite.

pboling:
> @mislav My only real qualm here is that the fix is actually unrelated to fedora.  Its a fix for GCC v15.  The issue is we don't have a parser for GCC version as far as I can tell.
> 
> 
> 
> In an ideal world we would adjust the CFLAGS based on GCC version so the flavor of linux isn't relevant.  Perhaps this is good enough for now; it can be iterated on, and solves a decent chunk of current and future installed base.

pboling:
This site has a table of the default modes for each version op to v13:

https://sc.ddbj.nig.ac.jp/en/guides/software/DevelopmentEnvironment/gcc/

Fortunately, based on that table, we don't need to do much switching at all!

eregon:
For awareness, this causes `ruby-dev` and Ruby < 3.2 to be compiled with -O0 (unoptimized) on Fedora 42+: https://github.com/rbenv/ruby-build/issues/2637
Fix in https://github.com/rbenv/ruby-build/pull/2638

Issue #2544, opened 2025-06-25
ruby-build ignores pkg-config when deciding whether to install OpenSSL


baarde:
If no compatible OpenSSL version is found, ruby-build downloads and installs OpenSSL.

When looking for a compatible OpenSSL version, ruby-build will try the system OpenSSL, then Homebrew-installed OpenSSL versions.

ruby-build should to be able to find an OpenSSL version installed by other package managers, using `pkg-config` (like Autoconf does when building Ruby). 

### Steps to reproduce the behavior

* Install Ruby dependencies, as well as pkg-config, using your package manager of choice (say MacPorts):
  ```sh
  sudo port install pkgconfig openssl libyaml
  ```
* Check that pkg-config is present and finds OpenSSL:
  ```sh
  pkg-config --cflags --libs openssl
  ```
* Check that OpenSSL version is 3.x.x:
  ```sh
  pkg-config --modversion openssl
  ```
* Check that Ruby builds correctly on your machine:
  ```sh
  cd ruby-3.4.4 \
    && ./autogen.sh \
    && ./configure --prefix=/var/tmp/ruby-3.4.4 --enable-shared --with-ext=openssl,psych,+ \
    && make install
   ```
* Install Ruby using ruby-build:
   ```sh
   ruby-build 3.4.4 /var/tmp/ruby-3.4.4
   ```

### Expected behavior

ruby-build should use the OpenSSL package installed by the package manager.

### Actual behavior

ruby-build downloads and builds OpenSSL.

### Logs

```
==> Downloading openssl-3.0.16.tar.gz...
-> curl -q -fL -o openssl-3.0.16.tar.gz https://dqw8nmjcqpjn7.cloudfront.net/57e03c50feab5d31b152af2b764f10379aecd8ee92f16c985983ce4a99f7ef86
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 14.6M  100 14.6M    0     0  44.3M      0 --:--:-- --:--:-- --:--:-- 44.4M
==> Installing openssl-3.0.16...
-> ./config --prefix=/var/tmp/ruby-3.4.4/openssl --openssldir=/var/tmp/ruby-3.4.4/openssl/ssl --libdir=lib zlib-dynamic no-ssl3 shared
-> make -j 12
-> make install_sw install_ssldirs
==> Installed openssl-3.0.16 to /var/tmp/ruby-3.4.4
[...]

PR #2547, opened 2025-06-27
Add support for pkg-config

baarde:
When pkg-config is installed and finds the libraries, use those rather than relying on Homebrew or bundling OpenSSL.

Fixes https://github.com/rbenv/ruby-build/issues/2544

<hr />

### A few comments about package managers
* I've tried hard not to break things for Homebrew. When the libraries and pkg-config have been installed using Homebrew, pkg-config finds the libraries in `/opt/homebrew/Cellar/`. However, those have their version number in the path, so I decided to ignore any library in `*/Cellar/*` to avoid breaking Ruby when running `brew upgrade`.
* When the libraries and pkg-config have been installed using different package managers, whoever comes first in the search path wins.
* A package manager that distributes ruby-build should have pkg-config as a dependency as well as all required and optional libraries (openssl, readline, libffi, gmp) to make sure all the linked libraries come from the same source.

#### Skip  `--with-xyz-dir=`
pkg-config is used internally by Ruby build scripts. So my initial idea was to skip the `--with-xyz-dir=` option when a library is discovered by pkg-config and let the build scripts do their job. However:
* That doesn't work for gmp.
* That may not work for other libraries with older Rubies.
* Having the `--with-xyz-dir=` options appear in the output is helpful to make sure that the correct libraries are used.
* "Explicit is better than implicit." 😉

#### Do nothing
My goal was to write a simple fix to have ruby-build work well with MacPorts (and other package managers). But doing so without breaking things turned out a little more complex than anticipated. Maybe it's just not worth it.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1034,6 +1034,8 @@ needs_yaml() {
 use_pkgconfig_yaml() {
   local libdir
   libdir="$(pkg-config --variable=prefix yaml-0.1 2>/dev/null || true)"
+  # We don't want to link to Homebrew Cellar because links break when packages are upgraded.
+  [[ "$libdir" != */Cellar/* ]] || return 1
   if [ -d "$libdir" ]; then
     package_option ruby configure --with-libyaml-dir="$libdir"
   else
@@ -1065,6 +1067,8 @@ use_freebsd_yaml() {
 use_pkgconfig_gmp() {
   local libdir
   libdir="$(pkg-config --variable=prefix gmp 2>/dev/null || true)"
+  # We don't want to link to Homebrew Cellar because links break when packages are upgraded.
+  [[ "$libdir" != */Cellar/* ]] || return 1
   if [ -d "$libdir" ]; then
     package_option ruby configure --with-gmp-dir="$libdir"
   else
@@ -1100,6 +1104,8 @@ use_freebsd_readline() {
 use_pkgconfig_readline() {
   local libdir
   libdir="$(pkg-config --variable=prefix readline 2>/dev/null || true)"
+  # We don't want to link to Homebrew Cellar because links break when packages are upgraded.
+  [[ "$libdir" != */Cellar/* ]] || return 1
   if [ -d "$libdir" ]; then
     package_option ruby configure --with-readline-dir="$libdir"
   else
@@ -1121,6 +1127,8 @@ use_homebrew_readline() {
 use_pkgconfig_libffi() {
   local libdir
   libdir="$(pkg-config --variable=prefix libffi 2>/dev/null || true)"
+  # We don't want to link to Homebrew Cellar because links break when packages are upgraded.
+  [[ "$libdir" != */Cellar/* ]] || return 1
   if [ -d "$libdir" ]; then
     package_option ruby configure --with-libffi-dir="$libdir"
   else
@@ -1160,7 +1168,11 @@ pkgconfig_openssl_version() {
 }
 
 pkgconfig_openssl_prefix() {
-  pkg-config --variable=prefix openssl 2>/dev/null || true
+  local libdir
+  libdir="$(pkg-config --variable=prefix openssl 2>/dev/null || true)"
+  # We don't want to link to Homebrew Cellar because links break when packages are upgraded.
+  [[ "$libdir" != */Cellar/* ]] || return 0
+  printf "%s" "$libdir"
 }
 
 # List all Homebrew-installed OpenSSL versions and their filesystem prefixes.
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -675,16 +675,16 @@ build_package_standard() {
     local ruby_semver="$(normalize_semver "${package_name#ruby-}")"
     if [[ "$RUBY_CONFIGURE_OPTS ${RUBY_CONFIGURE_OPTS_ARRAY[*]}" != *--with-readline-dir=* && "$ruby_semver" -lt 300300 ]]; then
       # Ruby 3.3+ does not need external readline: https://github.com/rbenv/ruby-build/issues/2330
-      use_homebrew_readline || use_freebsd_readline || true
+      use_pkgconfig_readline || use_homebrew_readline || use_freebsd_readline || true
     fi
     if [[ "$RUBY_CONFIGURE_OPTS ${RUBY_CONFIGURE_OPTS_ARRAY[*]}" != *--with-libffi-dir=* ]]; then
-      use_freebsd_libffi || true
+      use_pkgconfig_libffi || use_freebsd_libffi || true
     fi
     if [[ "$RUBY_CONFIGURE_OPTS ${RUBY_CONFIGURE_OPTS_ARRAY[*]}" != *--with-libyaml-dir=* ]]; then
-      use_homebrew_yaml || use_freebsd_yaml || true
+      use_pkgconfig_yaml || use_homebrew_yaml || use_freebsd_yaml || true
     fi
     if [[ "$RUBY_CONFIGURE_OPTS ${RUBY_CONFIGURE_OPTS_ARRAY[*]}" != *--with-gmp-dir=* ]]; then
-      use_homebrew_gmp || true
+      use_pkgconfig_gmp || use_homebrew_gmp || true
     fi
     if [[ "$RUBY_CONFIGURE_OPTS ${RUBY_CONFIGURE_OPTS_ARRAY[*]}" != *--with-openssl-dir=* ]]; then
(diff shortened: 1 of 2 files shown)


halostatue:
@headius @eregon It would be really nice to have ruby-build work better with MacPorts by using pkg-config as this PR enables. I stopped trusting Homebrew for dev build dependencies a long time ago.

I think that there may be similar configuration changes required for TruffleRuby because it requires that I override the libyaml prefix for all runs.

mislav (changes requested):
Hi, thank you so much for the contribution, and sorry for the late review.

I generally agree that we should support discovering openssl from pkg-config if available instead of blindly downloading and compiling a new OpenSSL instance, but for the sake of a smaller diff, I would strongly prefer if the initial code change was scoped to just that.

Also, since Ruby's build system already supports discovering libraries from pkg-config, I would avoid trying to replicate that functionality too much in ruby-build. I know you suggested that explicitly passing `--with-*-dir=...` flags is good for visibility, but we also want let Ruby's build system have a chance to do the right thing. We only add support for discovering Homebrew-installed libraries because Ruby doesn't have any capabilities to do that.

So here's an idea for a minimal fix: instead of all the changes in this branch, how about scoping the fix to just `needs_openssl` so that it returns false if a compatible openssl version was found with pkg-config? Something like this:

```sh
  local system_version
  if has_broken_mac_openssl; then
    system_version="$(pkg-config --modversion openssl 2>/dev/null || true)"
  else
    system_version="$(system_openssl_version)"
  fi
```

The logic behind this would be: if on a Mac with a broken system OpenSSL version (the default for all macOS systems), treat the pkgconfig-discovered version as the "system" version and exit if that satisfies the requirement. Would this be enough to support MacPorts?

mislav:
> if you look at the linked MacPorts PR, there's a few libraries which are available via MacPorts that aren't found by ruby-build without explicit configuration.

I've taken a look at the linked PR, but I don't understand which are these libraries you are referring to. Are you saying that there can be MacPorts-installed libraries such as libyaml that are discoverable using `pkg-config`, but Ruby itself won't discover them unless an explicit `--with-*-dir=` was passed to its configure step?

> I've seen a problem where ruby-build appears to favour Homebrew-discovered libraries

Ah, I think I understand this part. If Homebrew-discovered libraries were found by ruby-build, then indeed it will automatically configure Ruby by having `--with-*-dir=` point to Homebrew libraries, overriding Ruby build system otherwise discovering those libraries from elsewhere on the system. Out project assumes that if you have `brew` in your PATH and libraries installed via Homebrew, you likely want to use those libraries. Is that a wrong assumption?

We could potentially avoid doing the Homebrew discovery step if the same libraries were found using pkg-config. However, in that case I still wouldn't pass the paths to those libraries using `--with-*-dir=`, and instead just pass nothing and let Ruby figure it out.

Or, we could add an environment variable that would prevent `brew` from ever being consulted while looking up libraries, which could help users like you who have Homebrew installed but don't necessarily want to link any programs to it.

eregon:
> 1. If brew libraries are installed, they are used. This is incorrect when MacPorts is installed and the library can be found by pkg-config.

I think this is subjective. In that case I think using either the library from Homebrew or MacPorts (or in general from one of the package managers where that package is installed) is correct. I'm pretty sure many people prefer Homebrew to MacPorts, and some might have both installed (though that's clearly a recipe for troubles).

> 2. If the brew libraries are not installed, there is no search in MacPorts and ruby-build falls back to bundled OpenSSL, requires an explicit libyaml declaration, and skips GMP, both of which are installed by MacPorts.

That I understand would be better, to consult `pkg-config`.
(BTW how does that work for MacPorts, where does it install the `pkg-config` executable and how does that end up in `PATH`?)

However, I have found that using `pkg-config` is unreliable and problematic in several cases:
* There is [a pure-Ruby gem named `pkg-config`](https://rubygems.org/gems/pkg-config) with an executable of the same name which has slightly different semantics. In my experience, it just breaks stuff and so calling to `pkg-config` if that is installed breaks things.
* Not considering the gem, there are multiple implementations of `pkg-config` (e.g. `pkgconf` on Fedora & related) and the semantics do not always match, see for example https://github.com/flavorjones/mini_portile/issues/118
* Only the first `pkg-config` in PATH will be tried, but it might not be the right one. There could be like 4 of them or so (gem, system, brew, macports).
* There is well-known issue with older versions of the openssl gem where one needed to set both `--with-openssl-dir` and `PKG_CONFIG_PATH` otherwise you'd end up with mixing headers from some openssl version and library from some other openssl version, this got fixed in https://github.com/ruby/openssl/pull/486. So at least in that case it's not fine to just rely on `pkg-config` and auto-detection.
* As you said, pkg-config might give too precise paths like you said it does for Homebrew, that's counter-productive.

For these reasons I'm wary to use `pkg-config` for anything, at least I would use it as the last thing to try, after trying well-known package managers like Homebrew & MacPorts.
It's also possible that pkg-config config files in installed packages are wrong or missing and so `pkg-config` just returns incorrect data then.

> I think that there may be similar configuration changes required for TruffleRuby because it requires that I override the libyaml prefix for all runs.

TruffleRuby explicitly looks for Homebrew and MacPorts, documented here
https://github.com/oracle/truffleruby/blob/master/doc/user/installing-libyaml.md
https://github.com/oracle/truffleruby/blob/master/doc/user/installing-libssl.md
and the search is done in
https://github.com/oracle/truffleruby/blob/master/lib/truffle/truffle/libyaml-prefix.rb
https://github.com/oracle/truffleruby/blob/master/lib/truffle/truffle/openssl-prefix.rb

It should work, as it tries MacPorts if it doesn't find it in Homebrew.
TruffleRuby tries Homebrew first, because that's what we can test because it's what we have in CI and BTW also I think what most Rubyists use.

---

IMO these complications mostly stem from the fact macOS doesn't have a proper package manager where libraries are just found without extra config. On Linux it tends to just work because there is typically one system package manager and stuff just ends up in `/usr`.
Homebrew for darwin-amd64 installs to `/usr/local` and that just works, but Homebrew for darwin-aarch64 installs to `/opt/homebrew` and that's not found by default and needs to be explicit, and worse than that macOS doesn't allow adding to DYLD_LIBRARY_PATH etc :( (see https://github.com/Homebrew/brew/issues/13481)
I suppose in theory `pkg-config` could maybe be a solution for this, but the reasons above I don't think it's reliable enough and also plenty of packages don't ship a config for `pkg-config`, or simply the system doesn't have `pkg-config` installed, so I guess it would be of little help e.g. for the `ffi` gem to find libraries.

---

I like @mislav's suggestion in https://github.com/rbenv/ruby-build/pull/2547#pullrequestreview-3066424505.
For cases of having both the package installed in Homebrew and MacPorts  I think it's reasonable for those users preferring MacPorts to explicitly specify it with `--with-openssl-dir`, etc. We can't satisfy both cases anyway, some people would prefer Homebrew in that scenario.

halostatue:
I don't agree with your assessment that developers would prefer homebrew over MacPorts of both are installed—as I said earlier the problem is that Homebrew tends to move the files around when they decide that they're no longer where they think they should be.

MacPorts (default install) is always in `/opt/local` and nowhere else and doesn't move files around at a whim the way that Homebrew does. If libraries are found in both places, the stable one should be preferred.

I'm not on my home computer to be able to assess the rest of what you said at this point, so this comment might be updated.

mislav:
Hi, thank you for the contribution and for simplifying the diff like I asked! I merged this after making some tweaks of my own, namely:

1. Homebrew's Cellar directory is now matched precisely against paths reported by `$(brew --prefix)/Cellar` and `$(brew --repository)/Cellar`. Any other path with "Cellar" being one of its segment is not considered related to Homebrew.
2. I've fixed some tests that were failing due to test stubs that were outdated or incomplete. (The stubs approach in our test suite is a little messy and brittle, so apologies if that was hard to pin down when you were working on your contribution.)

bin/ruby-build
```diff
@@ -1106,13 +1106,18 @@ has_broken_mac_openssl() {
...
-
-  pkg-config --modversion openssl 2>/dev/null || true
+# Print the prefix of a library found by pkg-config, but only if it
+# doesn't come from Homebrew's cellar.
+pkgconfig_prefix() {
+  local prefix brew_prefix
+  prefix="$(pkg-config --variable=prefix "$1" 2>/dev/null || true)"
+  [ -n "$prefix" ] || return 1
+  brew_prefix="$(brew --prefix 2>/dev/null || true)"
+  if [[ -n $prefix_prefix && ( $prefix == "$brew_prefix"/Cellar/* || \
```
eregon, line 1116: `$brew_prefix` -> `$brew_prefix` probably?

Issue #2546, opened 2025-06-26
ruby-build throws invalid arithmetic operator for OS versions with floating-point


rkumarpa:
### Steps to reproduce the behavior

```
# Install rbenv
curl -fsSL https://github.com/rbenv/rbenv-installer/raw/HEAD/bin/rbenv-installer | bash

# rbenv init
eval "$(~/.rbenv/bin/rbenv init - bash)"

# Check rbenv version
rbenv --version

# Install ruby
rbenv install 2.7.3
```

### Expected vs. actual behavior
The expected behavior is successful installation. But we are seeing an error which says invalid arithmetic operator.

The issue is here - https://github.com/rbenv/ruby-build/blame/master/bin/ruby-build#L214 specifically this 

> "$VERSION_ID" -lt "$1"

The bash doesn’t compare floating point with _**lt**_ (in our case 7.9)

### Logs

```shell
root/.rbenv/plugins/ruby-build/bin/ruby-build: line 214: [[: 7.9: syntax error: invalid arithmetic operator (error token is ".9")
BUILD FAILED (Oracle Linux Server 7.9 on x86_64 using ruby-build 20250610-1-g5693b62d)

```

tomwrobel:
Ditto for this - breaks for RHEL builds

PR #2549, opened 2025-06-30
Fix Fedora version check when VERSION_ID is decimal

mislav:
This fixes RHEL builds where VERSION_ID=7.9

Fixes https://github.com/rbenv/ruby-build/issues/2546


pboling (commented):
LGTM

PR #2576, opened 2025-10-12
Support RHEL 9+/Fedora 34+ CA bundle path

mikelolasagasti:
Detect `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem` used in modern RHEL and Fedora systems.

Also improve the error message when no system CA bundle is found.


mislav:
Thank you and sorry for the late review & merge!

PR #2578, opened 2025-10-20
Replace non-POSIX `type` checks with portable `command -v`

mikelolasagasti:


mislav (commented):
Thank you for the suggestion, but ruby-build is a bash codebase, not a POSIX-compatible project. Notice that all our shebangs are `#!/usr/bin/env bash`, not `#!/usr/bin/env sh`. Since bash is a good-enough lowest common denominator for all the platforms we aim to support, it has worked well for us and we haven't felt an incentive to move our scripts to be posix-compatible.

Closing this because it's a lot of changes that don't seem to be like they would tangibly make ruby-build run smoother across more platforms.

PR #2595, opened 2026-01-02
Build libyaml from source when Homebrew/FreeBSD packages unavailable

tobi:
## Summary

When `brew` is not in `PATH`, `use_homebrew_yaml` fails silently and Ruby compiles without libyaml/psych support. 

This PR adds a fallback that downloads and builds libyaml from source when brew isn't in the path, similar to how OpenSSL is already handled.

## Changes

- Add `build_package_yaml` to install libyaml to `$PREFIX_PATH/libyaml`
- Add `fetch_libyaml_latest_version` to get latest version from GitHub API (falls back to 0.2.5)
- Add `build_yaml_from_source` as fallback in the yaml detection chain
- Fix `use_freebsd_yaml` to return failure when not on FreeBSD (was silently succeeding)

## After

```
$ PATH=/usr/bin:/bin ruby-build 3.4.1 /tmp/ruby
ruby-build: building libyaml 0.2.5 from source
==> Downloading yaml-0.2.5.tar.gz...
==> Installing yaml-0.2.5...
# ... continues with Ruby build ...
$ /tmp/ruby/bin/ruby -e "require 'yaml'"
# works!
```

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -679,7 +679,7 @@ build_package_standard() {
       use_freebsd_libffi || true
     fi
     if [[ "$RUBY_CONFIGURE_OPTS ${RUBY_CONFIGURE_OPTS_ARRAY[*]}" != *--with-libyaml-dir=* ]]; then
-      use_homebrew_yaml || use_freebsd_yaml || true
+      use_homebrew_yaml || use_freebsd_yaml || build_yaml_from_source || true
     fi
     if [[ "$RUBY_CONFIGURE_OPTS ${RUBY_CONFIGURE_OPTS_ARRAY[*]}" != *--with-gmp-dir=* ]]; then
       use_homebrew_gmp || true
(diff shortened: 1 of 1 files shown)


eregon:
> Starting from Ruby 2.2, we stopped using the `needs_yaml` declaration in ruby formulae, although my memory is fuzzy and I'm not sure why. It might have something to do with the psych gem bundling its own libyaml at some point? /cc @hsbt [d646595](https://github.com/rbenv/ruby-build/commit/d646595b878772e23ac85c6bb414a1f72337d272)

Yes, psych used to vendor libyaml, but then more recently no longer does that and instead has the `--with-libyaml-source-dir` option.

eregon:
> When `brew` is not in `PATH`, `use_homebrew_yaml` fails silently and Ruby compiles without libyaml/psych support.

This should not happen, because of https://github.com/rbenv/ruby-build/blob/079b5af587b9471e37c81472548377ca88828058/bin/ruby-build#L711-L714

Did this break in CRuby or does this work as expected and ruby-build fails in that case?

eregon:
For context, ruby-build in general does not install dependencies itself, instead they are [documented here](https://github.com/rbenv/ruby-build/wiki#suggested-build-environment). The list is long so I'm unsure if just handling libyaml is a big help. ruby-build or CRuby should definitely fail early if libyaml is missing though.

eregon:
> Did this break in CRuby or does this work as expected and ruby-build fails in that case?

I tried on my machine by removing `libyaml-devel` and it fails as expected:
```
$ ruby-build ruby-4.0.0 ~/.rubies/ruby-4.0.0-test   
==> Downloading ruby-4.0.0.tar.gz...
...
==> Installing ruby-4.0.0...
-> ./configure "--prefix=$HOME/.rubies/ruby-4.0.0-test" --enable-shared --with-ext=openssl,psych,+
-> make -j 16
*** Following extensions are not compiled:
psych:
	Could not be configured. It will not be installed.
	Check /tmp/ruby-build.20260104173455.588508.nksbP5/ruby-4.0.0/ext/psych/mkmf.log for more details.

BUILD FAILED (Fedora Linux 41 on x86_64 using ruby-build 20251225-2-g079b5af5)
```

From that, I think keeping things as they are is best, it's already a clear error and AFAIK it's not hard to install libyaml (unlike OpenSSL which is messy due to Ruby needing very specific versions).
Also just solving libyaml means still missing e.g. `readline gmp rust` on macOS.

PR #2604, opened 2026-02-17
Support building from arbitrary git revision

byroot:
While bisecting, or building a nightly Ruby CI and some other situations, it is useful to be able to build a specific Ruby revision rather than just the tip of Ruby's master branch.

This PR allows to do this easily with:

```bash
RUBY_DEV_REPO=https://github.com/byroot/ruby.git RUBY_DEV_REF=my-branch@deadbeef ruby-build ruby-ruby path/to/install
```

(of course no strong opinion on the naming of these env vars, perhaps they should be prefixed by `RUBY_BUILD_`).

This is way more convenient than generating a definition on the fly, or than to maintain a long list of definitions like I used to do at https://github.com/Shopify/ruby-definitions

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,8 @@ The build process may be configured through the following environment variables:
 | `RUBY_MAKE_INSTALL_OPTS`        | Additional `make install` options (applies only to Ruby source).                                 |
 | `NO_COLOR`                      | Disable ANSI colors in output. The default is to use colors for output connected to a terminal.  |
 | `CLICOLOR_FORCE`                | Use ANSI colors in output even when not connected to a terminal.                                 |
+| `RUBY_REPO`                     | The URL of the git repository to use when building `ruby-dev`                                    |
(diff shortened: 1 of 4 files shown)


byroot:
👋 @hsbt or @eregon any opinions on this?

hsbt:
Hi, I prefer this feature because I'm also build custom `ruby-dev` version every day.

I think `RUBY_DEV_REPO` and `RUBY_DEV_REF` can also be `RUBY_REPO` and `RUBY_REF`, but is this because they are used by ruby-dev?

and Is it a mistake to say `ruby-ruby` in your example instead of `ruby-dev`?

byroot:
> or just the commit would be enough?

I may be missing something from newer git versions, but since we only clone a single branch, the commit we are interested in might be missing.

There might be a way I don't know to directly clone the commit we want, if so let me know and I'm happy to explore this further.

PR #2610, opened 2026-03-21
Support building with version prefix

### LINKED ISSUE #2165: Version aliases
Please make this work:

`rbenv install 2`

this and 2.7 should currently install 2.7.7.

byroot:
Commonly in CI pipelines and other scripts, users may want to install the latest from a specific series, e.g. `3.4` or `jruby-10.0`.

Example: https://github.com/trilogy-libraries/trilogy/pull/274

```bash
$ bin/ruby-build 3.4 /tmp/3.4
...
Downloading ruby-3.4.9.tar.gz...

$ bin/ruby-build 4.0 /tmp/4.0
...
Downloading ruby-4.0.2.tar.gz...
```

Fixes https://github.com/rbenv/ruby-build/issues/2165


eregon (dismissed):
Looks great.
Could you also update the README to mention this new possibility? (`ruby-build version-prefix`, the `--resolve` thing probably doesn't need a mention as it's more internal).

It'd be good if @mislav could also take a look.
If he's not available I think it's OK to merge this in a few days, the change looks safe to me & ready to go.

mislav (changes requested):
Hi, thanks for working on this feature!

A couple of things don't sit well with me at the moment:

1. Is adding a new `ruby-build --resolve` flag really necessary? Couldn't we get away without adding one, for the sake of simplicity?
 
2. `filter_version_prefix()` and `resolve()` implementations feel a little over-complex to me. As I understand it, this feature is about finding the last Ruby version from `ruby-build --definitions` that starts with `{version_prefix}.*` - couldn't that be done with a one-liner? `bin/ruby-build --definitions | grep "^${v//./\.}\." | tail -n1`

3. `ruby-build -d ruby-3.2 ~/.rubies/` results in `~/.rubies/3.2.10` being created, but I specifically requested the "ruby-" prefix to be present. Some Ruby version managers expect every MRI Ruby to be prefixed with "ruby-". If I did `ruby-build -d ruby-3.2.10 ~/.rubies/`, that results in the expected `~/.rubies/ruby-3.2.10` installation location.

4. If I do `rbenv install 3.3` with the intention to have Ruby 3.3.10 installed, what should happen if `~/.rbenv/versions/3.3.10` already exists?

Ref. https://github.com/rbenv/ruby-build/issues/2165

eregon (approved):
I have tested with `3.3`, `ruby-3.3`, `truffleruby` and `truffleruby-33` and all works as expected (i.e. installed to a resolved version directory, preserving a ruby- prefix), both with `ruby-build` and `rbenv install`.

byroot:
@eregon I believe I addressed your concerns.

byroot:
> please allow me two weeks for re-review.

Gentle nudge 😄

mislav (changes requested):
I appreciate all the updates! Looks like "ruby-" prefix handling still fails in these cases:

1. `ruby-build -d ruby-3.2 ~/.rubies` - I've expected this to create `~/.rubies/ruby-3.2.11`, but it couldn't resolve "ruby-3.2"
2. `ruby-build ruby-3.2 /tmp/ruby-3.2` - likely same error cause as above, only now we aren't in `--dir` mode so it's maybe a different code path

Minor: `ruby-build --resolve nonexist` prints nothing but also exits with zero status. Perhaps it should exit with nonzero to indicate failure mode?

byroot:
Ack. Thanks for the review, I'll address it in a couple days once Kaigi is over

jasonkarns:
Apologies if this is clear to everyone else but I just want to make sure we're all on the same page.

> I am more open to fuzzy-matching version inputs to ruby-build then I am about fuzzy-resolving the version identifier from .ruby-version files. The latter is a separate feature and should be discussed separately.

@mislav , Am I understanding you correctly that you would _not_ want (or expect) `rbenv install` to fuzzy match from `.ruby-version`? (That's the way I interpreted the above statement.)

If so, that sounds contrary to @eregon and the sample indicating that `rbenv install` _does_ fuzzy match from `.ruby-version`:

> Yes, this PR has always been only about prefix-matching when installing, and nothing else.


To be clear, I do think it's natural to allow that fuzzy version matching, because it's just a pass-through to ruby-build itself. And I would find it confusing (I think) for ruby-build to have different semantics from rbenv-install. (other than simply "prefilling" ruby-build's arguments with rbenv aware values)

But I would _also_ expect `rbenv install 3.3` to fuzzy match to 3.3.11 because I think rbenv-install should behave the same regardless if it gets the version as a cli option or from the .ruby-version file.

I certainly don't think any of the other rbenv functionality needs fuzzy matching (nothing for any other rbenv subcommands other than `install`); not least because they aren't involved with installation nor are they managed by the ruby-build tool and repo.

But should this PR be merged as it stands now, this will be a (potentially significant) behavior change from how `rbenv install` works. (both with and without an explicit version option)

byroot:
@mislav I fixed the two issues and added tests for them.

mislav:
> But should this PR be merged as it stands now, this will be a (potentially significant) behavior change from how `rbenv install` works. (both with and without an explicit version option)

@jasonkarns I think I understand your concern that if `rbenv install` (without arguments) picks up the version from `.ruby-version` and ends up fuzzy-matching that to resolve e.g. "3.3" to "3.3.11", it would be a significant departure in rbenv's design philosophy that included never fuzzy-matching the contents of a `.ruby-version` file. Even though it's a minor detail, I see now that the workflow @eregon demonstrated feels like a bug, and I would propose only allowing `rbenv install` (without arguments) to respect the contents of a `.ruby-version` file if the version name gets matched exactly.

Would adding such a constraint to this PR alleviate your concerns, or was there something else?

eregon:
I think it's counter-productive and non-intuitive to prevent that, because then you break that `rbenv install` is just `rbenv install $(cat .ruby-version)`, which seems a bad trade-off to me.
Then maybe prefix matching should not be done at all for `rbenv-install`, but that is also non-intuitive since `rbenv install $V` should be `ruby-build $V DEST`.

I doubt anyone would ever complain because `rbenv install` works in more cases and there isn't really any ambiguity, this is **not** fuzzy matching, it's prefix matching. I think everyone understands that `3.3` means latest `3.3.x` release.

@jasonkarns your [first comment](https://github.com/rbenv/ruby-build/pull/2610#issuecomment-4104132395) actually expressed you wanted those semantics, maybe we all agree they are kind of intuitive?
> Given `.ruby-version` containing `3.4`
> When I run `rbenv install`
> Then it should install version 3.4.9

mislav:
> I doubt anyone would ever complain because `rbenv install` works in more cases and there isn't really any ambiguity, this is **not** fuzzy matching, it's prefix matching. I think everyone understands that `3.3` means latest `3.3.x` release.

I could swing both ways. I don't know if it as all common to have `.ruby-version` files containing only a version prefix rather than a full version name, but seeing how that wouldn't work with rbenv proper anyway (with or without this PR), I'm not terribly concerned about a bit of undefined or surprising behavior about what happens with `rbenv install` with no arguments. However, if someone from our org had qualms about this, I'd definitely hold off merging this until that's been addressed.

mislav (changes requested):
@byroot Should `--resolve` print the same version name if the argument passed was already a fully qualified version name?

```
$ ruby-build --resolve 3.4.9 || echo UNRESOLVED
UNRESOLVED
```

byroot:
@mislav good catch. I pushed a fix.

As for the debate about `.ruby-version` this is why I didn't want to get into the ruby-build part.
Personally, it doesn't bother me that `3.4` in `.ruby-version` would select the newest `3.4.x`, I think it's pretty intuitive.

Whether you want such behavior is another question though, because it might resolve to different things on different machine assuming not everyone has the same version of ruby-build, but I can imagine it being useful in larger companies with controlled dev environments.

mislav (approved):
Thank you for all the work so far!

I've pushed my changes fixing a few small things, simplifying code, and touching up docs:

- The special argument `ruby-build "ruby" /path/to/destination` is now supported to install the latest CRuby. I figured that it might be nice to add this since version-less arguments like "jruby" and "truffleruby" already worked to install their latest versions, and since arguments like `ruby-4` also worked.
- Fixed version sorting so that `ruby-build --resolve jruby` correctly resolves to JRuby 10.x instead of JRuby 9.x
- Added a case-normalization feature so that it's possible to pass arguments like "JRuby", "PicoRuby", "TruffleRuby" and have them resolve to their ruby-build definition names
- Added man page documentation for all this
- And more (see my commit message)

jasonkarns:
@mislav 
> it would be a significant departure in rbenv's design philosophy that included never fuzzy-matching the contents of a .ruby-version file.

Yep. That's the main reason I wanted to be clear here.

@eregon 
> I think it's counter-productive and non-intuitive to prevent that, because then you break that rbenv install is just rbenv install $(cat .ruby-version), which seems a bad trade-off to me.

I think it's natural to assume that `rbenv install` behaves the same as `rbenv install $(cat .ruby-version)`.

And I think it's also natural to assume that `rbenv install 3.3` behaves the same as `ruby-build 3.3 -d $(rbenv root)/versions/...`

Because both of these presumptions have been the behavior for a decade.

@mislav 
> I could swing both ways. I don't know if it as all common to have `.ruby-version` files containing only a version prefix rather than a full version name, but seeing how that wouldn't work with rbenv proper anyway (with or without this PR),

True, but in practice, [`rbenv-aliases`](https://github.com/tpope/rbenv-aliases) has provided that functionality for a long time and anecdotally, quite a few of my past clients and projects have all used that fuzzy matching via the rbenv-aliases plugin. (And I'm pretty sure the shortened version in .ruby-version file also still works with Bundler's `file:` option, and CircleCI's tooling.)

> if someone from our org had qualms about this, I'd definitely hold off merging this until that's been addressed.

I definitely don't have qualms, I just wanted to make sure that the behavior change is known and acknowledged, since it departs from current behavior. I think the rbenv user base is likely to receive the change well, anyhow, judging by the long list of issues that have been opened (and declined) requesting this exact or similar feature.

@byroot 
> As for the debate about `.ruby-version` this is why I didn't want to get into the ruby-build part.

It's not really possible to avoid. :D `rbenv install` is just a wrapper script around ruby-build, so changing ruby-build's behavior changes `rbenv install`. And if we were  to make ruby-build fuzzy matching require a flag, (thereby preventing fuzzy matching with `rbenv install`) I would _also_ consider that a deviation of behavior. Since that would break the implied understanding that rbenv-install is a simple pass-through to ruby-build.

---

All this to say, I'm 👍 on the feature and in favor of keeping rbenv-install's behavior consistent as a thin wrapper to ruby-build. I think I honestly preferred the version without the --resolve flag, as I suspect the majority of folks want that to be the default. (I wonder about an inverse `--strict` flag, so that the "resolve" behavior is the default?) 🤷 But I don't feel strongly about that.


Only last comment is on the doc phrasing:
`Print the most recent definition name matching a version prefix`

Just a reminder that users of the tool frequently confuse words like "latest", "available", "most recent" to think that ruby-build is checking _online_ for truly the latest released versions; when in actuality, it's the highest version of available definitions _on their machine_.  But I don't have any suggestions to improve it. (The other flag descriptions use similar words like "latest" and "available".)

mislav:
> Only last comment is on the doc phrasing: `Print the most recent definition name matching a version prefix`
> 
> Just a reminder that users of the tool frequently confuse words like "latest", "available", "most recent" to think that ruby-build is checking _online_ for truly the latest released versions; when in actuality, it's the highest version of available definitions _on their machine_.

@jasonkarns That's a great point. Since this feature is all about selecting the "latest" Ruby version matching a version prefix, I've added blurbs of documentation that clarify what ruby-build means by "latest" Ruby version. The docs now point out that ruby-build only ever knows about Ruby versions that it bundles definition files for. It also gives a tip for people to check out the [ruby-install](https://github.com/postmodern/ruby-install#readme) project if they want their installer tool to always do remote requests to download the actual list of latest Ruby versions.

mislav:
Thanks @byroot and everyone who participated! Shipped in https://github.com/rbenv/ruby-build/releases/tag/v20260501 🚀

PR #2614, opened 2026-03-29
Add PicoRuby 3.4.2 with new r2p2 executable

hasumikin:
This commit adds PicoRuby 3.4.2 and introduces a ~~new build function to accommodate changes in the build artifacts.~~ new artifacts: r2p2

Version numbering context:
PicoRuby's version numbers are aligned with mruby's VM code specification compatibility. Version 3.4 indicates compatibility with mruby 3.4 VM code format, or RITE0300. Version 3.4.2 indicates minor bug fix in PicoRuby.

Build artifact changes:
Starting from version 3.4.2, PicoRuby has replaced the irb executable with r2p2 (POSIX version of PicoRuby shell system including IRB). The new build produces three executables:
- picoruby: main Ruby interpreter
- picorbc: Ruby bytecode compiler
- r2p2: interactive REPL (replaces picoirb)

Implementation:
- ~~Added build_package_picoruby_r2p2() function for version 3.4.2+~~
- ~~Existing build_package_picoruby() remains for version 3.0.0~~
- ~~The new function creates only the 'ruby' symlink, omitting 'irb'~~
- Added conditional symlink logic in `build_package_picoruby()`: symlinks picoirb as irb if present
  - If there's no picoirb in artifacts, symlink for irb is not created
- Added picoruby-3.4.2 definition file

The split in build functions allows proper support for both the legacy 3.0.0 release with irb and newer releases with r2p2.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -885,7 +885,9 @@ build_package_picoruby() {
   mkdir -p "$PREFIX_PATH"
   cp -fR build/host/* include "$PREFIX_PATH"
   ln -fs picoruby "$PREFIX_PATH/bin/ruby"
-  ln -fs picoirb "$PREFIX_PATH/bin/irb"
+  if [ -f build/host/bin/picoirb ]; then
+    ln -fs picoirb "$PREFIX_PATH/bin/irb"
+  fi
 }
 
 build_package_jruby() {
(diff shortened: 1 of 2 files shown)


bin/ruby-build
```diff
@@ -888,6 +888,13 @@ build_package_picoruby() {
   ln -fs picoirb "$PREFIX_PATH/bin/irb"
 }
 
+build_package_picoruby_r2p2() {
```
mislav: I appreciate the effort, but I am not sold on the separate build function for picoruby releases going forward. If the only change is that picoruby-3.4 does not ship with `picoirb` going forward, I would edit the existing `build_package_picoruby` function to make the following step be conditional based on the existence of `picoirb`:

```sh
ln -fs picoirb "$PREFIX_PATH/bin/irb"
```

Does that make sense or am I missing some potential drawbacks?

mislav (changes requested):
Hi, thanks for the contribution and sorry for the late review!

bin/ruby-build
hasumikin: @mislav 
You are right. I fixed my implementation. Check it again when you have time

PR #2621, opened 2026-05-03
Fix definition lookup for ruby-prefixed names like `ruby-dev`

amatsuda:
Currently, `rbenv install ruby-dev` fails as follows since 92d57fdbb64cb45e0b7e1ca2369e8bd8d004986e introduced via #2610.

```
% ruby-build --version
ruby-build 20260501

% rbenv install -f -v ruby-dev
ruby-build: definition not found: ruby-dev

The following versions contain `ruby-dev' in the name:
  jruby-dev
  mruby-dev
  ruby-dev
  truffleruby-dev

See all available versions with `rbenv install --list-all'.

If the version you need is missing, try upgrading ruby-build:

  brew upgrade ruby-build
```

This patch fixes this regression by trying the exact match first, then falling back to the partial match intruduced via #2610.

### DIFF
--- a/bin/ruby-build
+++ b/bin/ruby-build
@@ -1420,6 +1420,10 @@ resolve_version() {
 
   # Look for an exact match first.
   for DEFINITION_DIR in "${RUBY_BUILD_DEFINITIONS[@]}"; do
+    if [ -f "${DEFINITION_DIR}/$1" ]; then
+      echo "$1"
+      return 0
+    fi
     if [ -f "${DEFINITION_DIR}/${version}" ]; then
       echo "$version"
       return 0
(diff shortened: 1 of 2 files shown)


eregon (approved):
Thank you for the fix!

eregon:
@mislav I'll merge this now since this is a low-risk bug fix and it fixes a significant released breaking change.
Could you let me know if it's OK if I merge such bug fixes? If not, I'm sorry about that.

eregon:
Released as https://github.com/rbenv/ruby-build/releases/tag/v20260503

mislav:
@amatsuda Thank you!

@eregon Yes it’s perfectly fine to rush fixes for regressions like these; much appreciated.

PR #2632, opened 2026-07-14
Stop release automation for homebrew-core

mislav:
Since about three weeks ago, the homebrew-core repository started requiring its PR template to be submitted with every PR body. The bump-homebrew-formula-action automation does not send the PR template, thus the PRs it creates will be auto-closed by homebrew-core.

This quits using bump-homebrew-formula-action in favor of Homebrew's autobump mechanism. https://docs.brew.sh/Autobump


jasonkarns:
Keeping the thread here for the upstream (homebrew) conversation. https://github.com/Homebrew/homebrew-core/pull/293038#issuecomment-4970341308

I think if homebrew is open to adding an exception for bump-homebrew-formula-action, it could be worthwhile to keep the automation. My vote is dependent upon that decision.

mislav:
Agreed; if they add an exception specifically for us that simplifies things greatly. However, I wouldn't hold my breath, since that would set a precedent for 3rd-party tools to be exempt from parts of Homebrew's own triage process, and I don't actually expect Homebrew to grant us that. (I probably wouldn't grant it in their place either.)

MikeMcQuaid:
Sorry for pain here folks, taking a look.

MikeMcQuaid:
https://github.com/Homebrew/.github/pull/454 should help here (it'll get synced to other repos after). The motivation for these checks is just the deluge of AI spam PRs we receive.

MikeMcQuaid:
Should be fixed now. If you still continue to see this: feel free to @ me directly and I'll fix it ASAP.

Thanks for keeping Ruby tooling alive folks, love you all ❤️

mislav:
Much appreciated!

PR #2639, opened 2026-09-15
Set -x flag for release script to show commands being run

eregon:
* This makes it much easier to rerun a command if it failed and to know where the script stopped.


mislav (approved):
Good idea! I never needed this since I made the script and I would know by heart which part of it failed based on the output, but I can imagine this would be useful to everyone else.

script/release
```diff
@@ -8,7 +8,7 @@
 # - creates a new Release on GitHub
 # - [automated] a GitHub Action will create a Homebrew PR for the new release
 
-set -e
+set -ex
```
jasonkarns, line 11: Thoughts on using something like:

```
[ -n "${DEBUG-}" ] && set -x
```

So that the extra debug/tracing output is opt-in rather than always on?

eregon: Is there a use case for wanting to not see the extra output?
I think the extra output doesn't matter if it works, but it's important to be there if it fails.
Having it only after setting `DEBUG` is too late.

eregon:
What failed in my case is the `gh release create ...` line (because I only give read-only access to `gh` by default).
I had to rerun and manually copy/paste the values for the arguments, workable but would be so much easier if it just printed the full command, which it does with this change.

It actually happened multiple times to me already. There used to be an early `gh something` check to ensure everything works, but anyway I think this simple change is good enough.

